PR #21, opened 2023-01-02
Realm page update

bokdeuk-jeong:
1. Change the scope of s2tlb sync to inner sharable
2. [Temporary] Share realm data pages with nw
3. Add RSI for granule change. Initial submission.
    Will use event handler as rmi does.

### DIFF
--- a/rmm/armv9a/src/rsi/mod.rs
+++ b/rmm/armv9a/src/rsi/mod.rs
@@ -1,0 +1,1 @@
+pub const RSI_REMAP_PAGE: u64 = 0xfeed_0001;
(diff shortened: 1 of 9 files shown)


rmm/armv9a/src/rsi/mod.rs
```diff
@@ -0,0 +1 @@
+pub const RSI_REMAP_PAGE: u64 = 0xfeed_0001;
```
isaac2-lee, line 1: const value 사이에 "_" 는 의도된 값인가요?

bokdeuk-jeong: 네 가독성 높이라고 16bits마다 _ 넣어 줬어요.

isaac2-lee (approved):
approved this pull request

PR #32, opened 2023-02-08
assets: qemu update

bokdeuk-jeong:
Symptom:
    virtio reset in realms will cause vcpu reset with unwanted
    register values.

    Root cause:
    The vcpu state that qemu gets from kvm is not synchronized with
    the vcpu state that rmm is keeping.

    Solution[temporary]:
    Don't call cpu_synchronize_state() in arm_cpu_virtio_is_big_endian()
    which is a subroutine of virtio_set_state() in qemu.

../qemu-system-aarch64 \
        -kernel Image_realmvm \
        -initrd initramfs-busybox-aarch64.cpio.gz \
        -append "earlycon=pl011,mmio,0x1c0a0000 console=ttyAMA0 ssbd=force-off printk.devkmsg=on" \
        --enable-kvm \
        -cpu host \
        -smp 1 \
        -M virt,gic-version=3 \
        -device virtio-blk-pci,drive=blk0,scsi=off \
        -drive if=none,id=blk0,file={your-image-file},format=raw,file.locking=off \
        -m 256M \
        -nographic


isaac2-lee (approved):
approved this pull request.

PR #46, opened 2023-03-22
lib: Detach uart as a library 

bitboom:


bitboom:
Replace with https://github.com/Samsung/islet/pull/57

PR #58, opened 2023-04-25
change relationship between monitor and mainloop

bokdeuk-jeong:
1. Make the event handle loop(mainloop) extensible. previous:
   Handler(context, rmi, smc, ...);
   - All member objects of Monitor are delivered as arg.
   - Adding new features to monitor requires changes to Handler param. i.e., All existing listeners have to be modified to add the new param. now:
   Handler(context, monitor);

2. Place the event registration to mainloop and remove overlay functions. mainloop does the actual work in run(). previous: monitor simply calls run to make mainloop start its work. now: directly call run() in mainloop at main().


bitboom (approved):
Looks good to extensible event handler.

PR #60, opened 2023-04-27
rsi: process RSI through an event handle routine

bokdeuk-jeong:
Required data struct to process RSI
1. Context: prepared at the REC_ENTER handle routine to place an RSI request.
2. monitor to access smc (e.g., for measurement), RmmPagetable for IPA_STATE_SET and so on.
3. struct 'Run' to place RSI specific output.


isaac2-lee (approved):
approve this pull request

PR #65, opened 2023-05-09
sdk: Add C-style Attestation API

bitboom:
This PR
- Adds parse API to get claims' value
- Add C-style API by cbindgen
  - `cbindgen` creates C/C++11 headers for Rust libraries which expose a public C API
  - _Do not modify header directly_ (header is generated by `cargo build`)
- Add CI test
  - build & test (x86_64)
  - example (x86_64)
  - build (arm64)

```sh
+--------+                                         
| sdk    |   => cbindgen => header  =>  || app (c++) ||
| (rust) |   => cdylib  => library => 
+-------+                                          
```

## Rust
```rust
let user_data = b"User data";
let report = attester::attest(user_data).unwrap();
let claims = verifier::verify(&report).unwrap();
println!("Debug: {:?}", claims);

if let claim::Value::Bytes(value) = &claims.value(config::STR_USER_DATA).unwrap() {
    assert_eq!(user_data, &value[..user_data.len()]);
} else {
    assert!(false, "Wrong user data");
}

if let claim::Value::String(value) = &claims.value(config::STR_PLAT_PROFILE).unwrap() 
    assert_eq!(value.as_str(), "http://arm.com/CCA-SSD/1.0.0");
} else {
    assert!(false, "Wrong platform profile");
}
```

## Cpp
```cpp
using byte = unsigned char;
...
std::string user_data("User custom data");
if (islet_attest((const unsigned char*)user_data.c_str(),
                 user_data.size(),
                 report_out.data(),
                 &report_out_len) != 0) {
    return -1;
}
...
if (islet_verify(report_out.data(), report_out_len, claims_out.data(), &claims_out_len) != 0) {
    return -1;
}
...
islet_print_claims(claims_out.data(), claims_out_len);
...
if (islet_parse("User data", claims_out.data(), claims_out_len, value_out.data(), &value_out_len)) {
    return -1;
}
```


jinbpark (approved):

sdk/src/claim/platform.rs
```diff
@@ -1,29 +1,10 @@
...
-    HashAlgo = 2402,
-    SWComponents = 2399,
-    VerificationService = 2400,
-}
+use serde::{Deserialize, Serialize};
 
-#[derive(Debug)]
-pub struct SWComponent0 {
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SWComponent {
```
bokdeuk-jeong, line 4: nice to have a common struct for both realm platform. :)

PR #70, opened 2023-05-15
expand granule region for fvp dram

isaac2-lee:
1. Increase granule covered space for fvp dram1
2. remove find_granule unwrap() 
3. add RMI_DATA_DESTROY

### DIFF
--- a/rmm/monitor/src/rmi/realm/mod.rs
+++ b/rmm/monitor/src/rmi/realm/mod.rs
@@ -22,7 +22,10 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let mm = rmm.mm;
         let params_ptr = ctx.arg[1];
         mm.map(params_ptr, false);
-        granule::set_granule(ctx.arg[0], GranuleState::RD, mm);
+        if granule::set_granule(ctx.arg[0], GranuleState::RD, mm) != granule::RET_SUCCESS {
+            ctx.ret[0] = rmi::ERROR_INPUT;
+            return;
+        }
 
         let param = unsafe { Params::parse(params_ptr) };
         trace!("{:?}", param);
@@ -54,7 +57,11 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rmi = rmm.rmi;
         let _rd = unsafe { Rd::into(ctx.arg[0]) };
         let ret = rmi.remove(0); // temporarily
-        granule::set_granule(ctx.arg[0], GranuleState::Delegated, rmm.mm);
+        if granule::set_granule(ctx.arg[0], GranuleState::Delegated, rmm.mm) != granule::RET_SUCCESS
+        {
+            ctx.ret[0] = rmi::ERROR_INPUT;
+            return;
+        }
 
         match ret {
             Ok(_) => ctx.ret[0] = rmi::SUCCESS,
--- a/rmm/monitor/src/rmi/realm/mod.rs
+++ b/rmm/monitor/src/rmi/realm/mod.rs
@@ -7,7 +7,7 @@ use crate::event::Mainloop;
 use crate::listen;
 use crate::rmi;
 use crate::rmm::granule;
-use crate::rmm::granule::{GranuleState, RmmGranule};
+use crate::rmm::granule::GranuleState;
 
 extern crate alloc;
 
@@ -22,8 +22,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let mm = rmm.mm;
         let params_ptr = ctx.arg[1];
         mm.map(params_ptr, false);
-        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::Delegated);
-        g_rd.set_state(GranuleState::RD, mm);
+        granule::set_granule(ctx.arg[0], GranuleState::RD, mm);
 
         let param = unsafe { Params::parse(params_ptr) };
         trace!("{:?}", param);
@@ -55,9 +54,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rmi = rmm.rmi;
         let _rd = unsafe { Rd::into(ctx.arg[0]) };
         let ret = rmi.remove(0); // temporarily
-
-        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::RD);
-        g_rd.set_state(GranuleState::Delegated, rmm.mm);
+        granule::set_granule(ctx.arg[0], GranuleState::Delegated, rmm.mm);
 
         match ret {
             Ok(_) => ctx.ret[0] = rmi::SUCCESS,
--- a/rmm/monitor/src/rmi/realm/mod.rs
+++ b/rmm/monitor/src/rmi/realm/mod.rs
@@ -21,16 +21,12 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rmi = rmm.rmi;
         let mm = rmm.mm;
         let params_ptr = ctx.arg[1];
-
-        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::Delegated).unwrap();
-        let g_param = granule::find_granule(params_ptr, GranuleState::Undelegated).unwrap();
+        mm.map(params_ptr, false);
+        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::Delegated);
         g_rd.set_state(GranuleState::RD, mm);
-        g_param.set_state(GranuleState::Param, mm);
 
         let param = unsafe { Params::parse(params_ptr) };
         trace!("{:?}", param);
-
-        g_param.set_state(GranuleState::Undelegated, mm);
 
         // TODO:
         //   Validate params
@@ -47,6 +43,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             }
             Err(_) => ctx.ret[0] = rmi::RET_FAIL,
         }
+        mm.unmap(params_ptr);
     });
 
     listen!(mainloop, rmi::REC_AUX_COUNT, |ctx, _| {
@@ -59,7 +56,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let _rd = unsafe { Rd::into(ctx.arg[0]) };
         let ret = rmi.remove(0); // temporarily
 
-        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::RD).unwrap();
+        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::RD);
         g_rd.set_state(GranuleState::Delegated, rmm.mm);
 
         match ret {
--- a/rmm/monitor/src/rmm/granule.rs
+++ b/rmm/monitor/src/rmm/granule.rs
@@ -2,17 +2,18 @@ use crate::rmm::PageMap;
 extern crate alloc;
 use alloc::collections::btree_map::BTreeMap;
 
-pub const FVP_DRAM0_BASE: usize = 0x8000_0000;
-pub const FVP_DRAM0_SIZE: usize = 0x7C00_0000;
-pub const FVP_DRAM0_END: usize = FVP_DRAM0_BASE + FVP_DRAM0_SIZE - 1;
+use spinning_top::{Spinlock, SpinlockGuard};
 
-pub const FVP_DRAM1_BASE: usize = 0x8_8000_0000;
-pub const FVP_DRAM1_SIZE: usize = 0x8000_0000;
-pub const FVP_DRAM1_END: usize = FVP_DRAM1_BASE + FVP_DRAM1_SIZE - 1;
+const FVP_DRAM0_REGION: core::ops::Range<usize> = core::ops::Range {
+    start: 0x8000_0000,
+    end: 0x8000_0000 + 0x7C00_0000 - 1,
+};
+const FVP_DRAM1_REGION: core::ops::Range<usize> = core::ops::Range {
+    start: 0x8_8000_0000,
+    end: 0x8_8000_0000 + 0x8000_0000 - 1,
+};
 
 const GRANULE_SIZE: usize = 4096;
-const GRANULE_SHIFT: usize = 12;
-const GRANULE_BASE_ADDRESS: usize = FVP_DRAM0_BASE;
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub enum GranuleState {
@@ -23,17 +24,15 @@ pub enum GranuleState {
     RecAux,
     Data,
     RTT,
-    Param,
 }
 
 #[derive(Default, Clone, Copy, PartialEq)]
 pub struct Granule {
     state: GranuleState,
-    // refcount: usize,
-    idx: usize,
+    addr: usize,
 }
 
-static mut GRANULE: BTreeMap<usize, Granule> = BTreeMap::new();
+static mut GRANULE: Spinlock<BTreeMap<usize, Granule>> = Spinlock::new(BTreeMap::new());
 
 impl Default for GranuleState {
     fn default() -> Self {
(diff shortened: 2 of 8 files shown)


rmm/monitor/src/rmm/granule.rs
```diff
@@ -118,13 +121,8 @@ impl RmmGranule for Granule {
 }
 
 // Define a function find_granule which returns a Granule instance from the GRANULE array using the given address and expected state.
-pub fn find_granule(addr: usize, expected_state: GranuleState) -> Option<&'static mut Granule> {
+pub fn find_granule(addr: usize, expected_state: GranuleState) -> &'static mut Granule {
```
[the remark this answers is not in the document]
isaac2-lee: exoected_state, it's for validate check state.
in case of original idea, if current state is not same as expected, return the None or do panic.
now, it just warn with log.

rmm/monitor/src/rmm/granule.rs
```diff
@@ -145,25 +143,26 @@ pub fn find_granule(addr: usize, expected_state: GranuleState) -> Option<&'stati
                     idx: idx,
                 };
                 GRANULE.insert(idx, new);
-                Some(GRANULE.get_mut(&idx).unwrap())
+                GRANULE.get_mut(&idx).unwrap()
```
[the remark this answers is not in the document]
isaac2-lee: matched Granule and this are not same, It's new one.

isaac2-lee:
4. deleted Param state on GranuleState
5. add zeroize when set Delegated state

isaac2-lee:
6. move the set_state caller inside granule

---
for specification
Granule has a pas variable.
when I will implement codes about ipa state, it will be included.

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -58,9 +54,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rmi = rmm.rmi;
         let _rd = unsafe { Rd::into(ctx.arg[0]) };
         let ret = rmi.remove(0); // temporarily
-
-        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::RD).unwrap();
-        g_rd.set_state(GranuleState::Delegated, rmm.mm);
+        granule::set_granule(ctx.arg[0], GranuleState::Delegated, rmm.mm);
```
bokdeuk-jeong, line 57: ctx.arg[0] refers to the physical address of RD which already delegated prior to calling REALM_CREATE and is set to RD.
Why are you changing its state from RD to Delegated?

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -21,17 +21,12 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rmi = rmm.rmi;
         let mm = rmm.mm;
         let params_ptr = ctx.arg[1];
-
-        let g_rd = granule::find_granule(ctx.arg[0], GranuleState::Delegated).unwrap();
-        let g_param = granule::find_granule(params_ptr, GranuleState::Undelegated).unwrap();
-        g_rd.set_state(GranuleState::RD, mm);
-        g_param.set_state(GranuleState::Param, mm);
+        mm.map(params_ptr, false);
+        granule::set_granule(ctx.arg[0], GranuleState::RD, mm);
```
bokdeuk-jeong, line 25: Which one do you think is better to set the granule status of RD,  here or in the Rd::new( )?
I think that the rmi handler doesn't have to know what status of each memory granule should be and struct RD should be responsible for this.

rmm/monitor/src/rmm/granule.rs
```diff
@@ -56,194 +57,139 @@ impl RmmGranule for Granule {
...
     }
 }
 
-// Define a function find_granule which returns a Granule instance from the GRANULE array using the given address and expected state.
-pub fn find_granule(addr: usize, expected_state: GranuleState) -> Option<&'static mut Granule> {
-    let idx = granule_addr_to_idx(addr);
-    if idx >= RMM_MAX_GRANULES {
-        warn!("check the granule index {}", idx);
-        return None;
+pub fn set_granule(addr: usize, state: GranuleState, mm: PageMap) {
```
bokdeuk-jeong, line 84: set_granule() doesn't return success or failure. Is it successful always?

rmm/monitor/src/rmm/granule.rs
```diff
@@ -56,194 +57,139 @@ impl RmmGranule for Granule {
...
+    let granule = find_granule(addr);
+    let prev_state = granule.get_state();
+    granule.set_state(state);
+    match state {
+        GranuleState::Delegated => {
+            if prev_state != GranuleState::Undelegated
+                && prev_state != GranuleState::Delegated
+                && prev_state != GranuleState::RTT
+            {
+                granule.zeroize();
```
bokdeuk-jeong, line 94: 👍

rmm/monitor/src/rmm/granule.rs
```diff
@@ -56,194 +57,139 @@ impl RmmGranule for Granule {
...
-// Define a function find_granule which returns a Granule instance from the GRANULE array using the given address and expected state.
-pub fn find_granule(addr: usize, expected_state: GranuleState) -> Option<&'static mut Granule> {
-    let idx = granule_addr_to_idx(addr);
-    if idx >= RMM_MAX_GRANULES {
-        warn!("check the granule index {}", idx);
-        return None;
+pub fn set_granule(addr: usize, state: GranuleState, mm: PageMap) {
+    let granule = find_granule(addr);
+    let prev_state = granule.get_state();
+    granule.set_state(state);
```
bokdeuk-jeong, line 87: I think we need to check current state first and check whether the state transition to the new state is valid. If it is okay, set the state to the new.

rmm/monitor/src/rmm/granule.rs
```diff
@@ -56,194 +57,139 @@ impl RmmGranule for Granule {
...
+            }
+            GRANULE.remove(&granule.addr());
+        },
+        _ => {
+            // TODO: after RIPAS implement, it will be changed to panic
+            if prev_state != GranuleState::Delegated {
+                warn!(
+                    "granule[{:X}] state is strange {:?}<-{:?}",
+                    addr, state, prev_state
+                );
```
bokdeuk-jeong, line 115: In this case, returning results indicating failure seems better.

rmm/monitor/src/rmm/granule.rs
```diff
@@ -56,194 +57,139 @@ impl RmmGranule for Granule {
...
+                GRANULE.get_mut(&addr).expect("granule insert failed!")
             }
         }
     }
 }
 
 // Define a function granule_addr_to_idx which returns the index of the granule using the given address.
-fn granule_addr_to_idx(addr: usize) -> usize {
-    if addr < GRANULE_BASE_ADDRESS || addr > FVP_DRAM0_END {
+fn check_validate_addr(addr: usize) {
```
bokdeuk-jeong, line 142: Currently you expect RMM/realm to use DRAM only. However, if we want to allocate dedicated IO resources (MMIO) to the realm world, which is already assigned to the normal world only, this validation check for the delegation doesn't work for such resources. Any idea on this?

rmm/monitor/src/rmm/granule.rs
```diff
@@ -136,34 +123,42 @@ pub fn find_granule(addr: usize, expected_state: GranuleState) -> Option<&'stati
...
+                }
                 let new = Granule {
-                    state: expected_state,
+                    state: GranuleState::Undelegated,
                     // refcount: 0,
                     idx: idx,
                 };
                 GRANULE.insert(idx, new);
-                Some(GRANULE.get_mut(&idx).unwrap())
+                GRANULE.get_mut(&idx).expect("granule insert failed!")
```
jinbpark: A simple suggestion here-:

I think `GRANULE` must be protected by a lock (not only GRANULE, but everything related to memory management in RMM in order to support multi cores). Of course, we don't have to deal with multi core environments yet so it's fair to say using `GRANULE` without lock is safe as of now.

But, in a long term, it would be better to use any kind of lock for variables that can be accessed in parallel.
(In idomatic rust, [Guard pattern](https://rust-unofficial.github.io/patterns/patterns/behavioural/RAII.html) is commonly used for lock or security-related protection. e.g., if we want to enforce some "validation check" on some struct, we can go with guard pattern to realize this)
Using locks for such variables (in accordance with RMM specs) would work on both single-core and multi-core so it has no side-effects.

To sum up, it would be nice to see that this concurrency safety is explicitly stated as one of todo items-:)

rmm/monitor/src/rmm/granule.rs
```diff
@@ -109,25 +102,19 @@ impl RmmGranule for Granule {
...
-    if idx >= RMM_MAX_GRANULES {
-        warn!("check the granule index {}", idx);
-        return None;
-    }
-
     unsafe {
         match GRANULE.get_mut(&idx) {
             Some(g) => {
+                // TODO: after RIPAS implement, it will be changed to panic
                 if expected_state != g.get_state() {
```
jinbpark: Suggestion here -:)

I think the way to validate some conditions (i.e.,  `if expected_state != g.get_state()`) can be implemented in a more rusty way, using Guard pattern. This is also good for error handling. 

For example,
```
1. struct Granule
2. struct GranuleGuard
    --> newly introduced as part of guard pattern. 
    --> it validates some conditions
    --> only if validations pass, find_granule returns "GranuleGuard" instead of "Granule"
3. callers (who call find_granule())
    --> access and modify "Granule" through "GranuleGuard"
```

What this pattern says is only who owns "GranuleGuard" can access and modify "Granule" behind "GranuleGuard" and owning "GranuleGuard" means that it has passed all kinds of validation checks.

This is how to enforce validation checks to access a real data (i.e., Granule) using a guard struct (i.e., GranuleGuard) in general in rust.

I think this change doesn't have to be done in this PR.  But I hope this change comes in another PR.
(though, it won't be a high priority job-:) as it just kind of refactoring toward idiomatic rust.)

rmm/monitor/src/rmi/realm/mod.rs
isaac2-lee: according to flow of realm creation,
after REALM_DESTORY, UNDELEGATE command will be reached for RD.(RTT, too - but we can't conider rtt memory yet)
for state machine flow, we have to change state.

rmm/monitor/src/rmi/realm/mod.rs
isaac2-lee: for readability, inside RD::new will be better. but, for managing granule state clearly, current position will be better.
according to realm destruction flow, we have to change state by flow

rmm/monitor/src/rmm/granule.rs
isaac2-lee: yes. success or panic..

rmm/monitor/src/rmm/granule.rs
isaac2-lee: now, we always zeroize the memory on delegated state. 
IMO, if DATA_CREATE_UNKNOWN command will be implement, we have to check it.

rmm/monitor/src/rmm/granule.rs
isaac2-lee: ok. I will

rmm/monitor/src/rmm/granule.rs
isaac2-lee: yes. we can apply the lock later. nice suggestion!

rmm/monitor/src/rmi/realm/mod.rs
bokdeuk-jeong: Granule status of RD comes together with RD lifecycle. I am not sure that update the granule status here aprart from RD make it more manageable. We have to change two place to make the status between RD and granule synchronized. I don't think we can say it is manageable.

rmm/monitor/src/rmm/granule.rs
isaac2-lee: well.. if address will be invalid, it will be panic, warn is just failed case to check state.

rmm/monitor/src/rmm/granule.rs
```diff
@@ -41,209 +41,146 @@ impl Default for GranuleState {
...
-            }
+        match GRANULE.get_mut(&addr) {
+            Some(g) => g,
             None => {
                 let new = Granule {
-                    state: expected_state,
+                    state: GranuleState::Undelegated,
                     // refcount: 0,
-                    idx: idx,
+                    addr: addr,
```
bitboom, line 123: We can init members [shorthand](https://rust-lang.github.io/rfcs/1682-field-init-shorthand.html) way. And it is preferred style :)

```suggestion
                    addr,
```

rmm/monitor/src/rmm/granule.rs
```diff
@@ -41,209 +41,146 @@ impl Default for GranuleState {
...
             }
         }
     }
 }
 
-// Define a function granule_addr_to_idx which returns the index of the granule using the given address.
-fn granule_addr_to_idx(addr: usize) -> usize {
-    if addr < GRANULE_BASE_ADDRESS || addr > FVP_DRAM0_END {
+// Define a function check_validate_addr which check the address is valid.
+fn check_validate_addr(addr: usize) {
```
bitboom, line 133: I think `validate_addr` is enough.

```suggestion
fn validate_addr(addr: usize) {
```

rmm/monitor/src/rmi/realm/mod.rs
bitboom: I think this function should not make panic.. We should return proper fail value to the monitor.

rmm/monitor/src/rmm/granule.rs
bitboom: As you already know, [logging by uart](https://github.com/Samsung/islet/blob/main/lib/uart/src/pl011.rs#L174) is already protected by locking. We should protect global variables like this :)

rmm/monitor/src/rmm/granule.rs
```diff
@@ -41,209 +41,146 @@ impl Default for GranuleState {
...
+    if addr < GRANULE_BASE_ADDRESS
+        || (addr > FVP_DRAM0_END && addr < FVP_DRAM1_BASE)
+        || addr > FVP_DRAM1_END
```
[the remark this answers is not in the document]
isaac2-lee: applied it

rmm/monitor/src/rmm/granule.rs
isaac2-lee: applied it

rmm/monitor/src/rmm/granule.rs
isaac2-lee: applied it

rmm/monitor/src/rmm/granule.rs
```diff
@@ -2,14 +2,17 @@ use crate::rmm::PageMap;
...
 use alloc::collections::btree_map::BTreeMap;
 
-pub const FVP_DRAM0_BASE: usize = 0x8000_0000;
-pub const FVP_DRAM0_SIZE: usize = 0x7C00_0000;
-pub const FVP_DRAM0_END: usize = FVP_DRAM0_BASE + FVP_DRAM0_SIZE - 1;
+use spinning_top::{Spinlock, SpinlockGuard};
+
+const FVP_DRAM0_BASE: usize = 0x8000_0000;
+const FVP_DRAM0_SIZE: usize = 0x7C00_0000;
+const FVP_DRAM0_END: usize = FVP_DRAM0_BASE + FVP_DRAM0_SIZE - 1;
```
bitboom, line 9: Since `Range` is half open, we don't need to define *_END.
How about to fix like this? I think we could fix `validate_addr()` more concise too.

```suggestion
const FVP_DRAM0_REGION: core::ops::Range = core::ops::Range { start: 0x8000_0000, end: 0x7C00_0000 });
```

rmm/monitor/src/rmm/granule.rs
jinbpark: @bitboom, as a side note, according to the RMM spec and the codes of tf-rm, `GRANULE` doesn't seem to need to be protected under a lock (tf-rmm doesn't protect `GRANULE` as well), instead each entry (`struct Granule`) behind that global variable needs to be protected.

What's interesting in memory management of RMM is the RMM spec says how to implement lock mechanism for several key structures.  (e.g., some field is accessed through atomic operations while others are accessed through locks. + some structures require refcount, seems quite complicated..)

I think we would benefit from Rust's idiomatic secure coding practices when implementing locking not just for `Granule` but also other structures. Locking seems a huge deal in RMM from security standpoint. (It would be an interesting engineering job-!)
We may need a separate PR that only focuses on implementing Locking securely in accordance with RMM spec.

rmm/monitor/src/rmm/granule.rs
isaac2-lee: applied it

rmm/monitor/src/rmm/granule.rs
isaac2-lee: applied it

bitboom (approved):

PR #79, opened 2023-05-26
Add code generation model

jinbpark:
## What is included in this PR
- A script to generate a simple code generation model (more precisely, it's not a code generation model like ChatGPT, rather it's just a simple classification model at this moment)
- All components in confidential-ml demo (Runtime/Device/ModelProvider) are capable of taking either "word prediction model" and "code generation model" in the same way.
- A client-side GUI (browser-based) support (mirroring-like GUI support)

## What is not included in this PR (will be done in the next PR)
- I'll update the docker image accordingly that includes node-js server codes to support a GUI demonstration.
- I'll update README.md to describe instructions for testing with GUI.


bitboom:
[minor] Indents are mixed (2 or 4). Could you please apply formatting like below? 

```sh
astyle --style=google --indent=spaces=2 *.cc *.h
```

jinbpark:
> astyle --style=google --indent=spaces=2 *.cc *.h

I forgot it, reflected:

bitboom (approved):

PR #82, opened 2023-05-31
Update network configuration

jinbpark:
## What is included in this PR

- Simply speaking, with this PR, applications running on ARM CCA can act as a server. 
- Technically speaking, it involves network configuration changes as follows.
  - before:  A realm doesn't have a static ip address as kvmtool launches with a user mode networking.
  - after: A realm has a static ip address as kvmtool launches with a tap networking.

- NOTE that this network configuration is heavily **based on the suggestion made by @p-sawicki2** few weeks ago (appreciate it!). Back then, we didn't have need to run an application as a server on ARM CCA, so I didn't reflect his suggestion into PR. But now, we need it as the device part of confidential-ml demo should be running as a server to receive input from web server.

- `./scripts/fvp-cca --normal-world=linux-net --realm=linux --rmm=tf-rmm` would go with a default network configuration. And in most cases, the default configuration would be sufficient. (To test it, we don't need spare IP addresses because it takes a set of dummy IP addresses and uses them only locally) See `docs/network.md` for details.

## TODO

- docker image update, integrating chatbot GUI update (e.g., shielding images) and GUI development for Ceritifier service.


bitboom:
[minor] I think it's good to apply default option enabled network _without linux-net_.

```sh
./scripts/fvp-cca -nw=linux -rm=linux
```

p-sawicki2 (approved):
LGTM

jinbpark:
> ./scripts/fvp-cca -nw=linux -rm=linux

I'll do this in the next PR, which is going to be done today to update README.md and a docker image to share it with vmware.

PR #83, opened 2023-06-02
lower exception: handle system register access

bokdeuk-jeong:


rmm/monitor/src/rmi/rec/params.rs
```diff
@@ -17,6 +17,10 @@ impl Params {
     pub fn pc(&self) -> usize {
         unsafe { self.pc.val as usize }
     }
+
+    pub fn gprs(&self) -> impl Iterator<Item = &u64> {
```
hihi-wang, line 21: Is there any specific reason to return `iterator` here? If not, I think it's more idiomatic to enforce function caller to explicitly use `iter()` like :
~~~
some_obj.gprs().iter().some_method()
~~~

bokdeuk-jeong: If we return the iterator, isn't the caller able to use it with gprs().iter() as you said? Currently, as the index also matters, the caller is using gprs with gprs().enumerate().
gprs is an array of u64. As the length of the array can be changed in the future, I didn't want to expose the length information to its callers or the callers to directly manupulate the gprs as an array. Having this in mind, I defined gprs() to return an iterator. Considering this, what would you suggest?

hihi-wang: It's same on caller side, as you said. But I think it's better to not return iterator here, it's bit awkward as function name does not meant iterator. Plus, above `pc()` does not return value of it, so I think it's better to match with other methods.

Returning slice `&[u64]` would be better in this case, as we don't need to return ownership of gprs.

bokdeuk-jeong: That's a good point that the function name gprs() doesn't bear any implication of iterator. However, I am still not happy with returning an array reference. The caller should not care about the array length. It can be of any size. I don't want to the caller make any mistake from the array manipulation. Any better idea? :)

hihi-wang: Caller does not need to care of length as it's slice, not array. Caller can access elements on it regardless to return slice or iterator. One difference is that caller can indexing the slice. I think it's better to return slice, unless if you really need to prevent indexing.

bokdeuk-jeong: I don't want to directly access an element with a specific index. However the slice has the length information at least and prevent error. Thanks a lot!

jinbpark (approved):

PR #85, opened 2023-06-09
Add islet CLI & realm signature verification

bitboom:
This PR
  - expands `rsictl` to `islet CLI` which can run on x86_64 as simulated version (thanks to @Havner who developed `rsictl` 😄)
  - detaches `cose signature verification logic` to the library and apply it to SDK

_Now only `attest` and `verify` are work on x86_64_

## Command Line Interface
ISLET provides Command Line Interface(CLI) which can explore CCA operations.
CLI supports both x86_64(simulated) and aarch64.

You can explore Attestation like below
```sh
$ cd cli
$ make x86_64
$ ./islet attest --output=./report
$ ./islet verify --input=./report
== Signature Verification:
Sign Algo        = [ES384]
Public Key       = ["0476f988091be585ed41801aecfab858...]
Data             = ["846a5369676e61747572653144a10138...]
Signature        = ["ec4f3b28a00feabd1f58f94acb27fdc7...]
== End of Signature Verification
== Realm Token cose:
Protected header               = Header { alg: Some(Assigned(ES
Unprotected header             = Header { alg: None, crit: [],
Signature                      = [ec4f3b28a00feabd1f58f94acb27f
== End of Realm Token cose
== Realm Token:
Realm challenge                (#10) = [abababababababababababa
Realm personalization value    (#44235) = [00000000000000000000
Realm hash algo id             (#44236) = "sha-256"
Realm public key hash algo id  (#44240) = "sha-256"
Realm signing public key       (#44237) = [0476f988091be585ed41
Realm initial measurement      (#44238) = [6190eb90b293886c172e
```

---

## Signature verification with SDK sample
```
$ cd sdk
$ make run-simulated

ISLET SDK examples: A simulated app running on x86_64
Simulated attestation operation on x86_64.
Verify Realm Signature.
== Signature Verification:
Sign Algo	 = [ES384]
Public Key	 = ["0476f988091be585ed41801aecfab858...]
Data		 = ["846a5369676e61747572653144a10138...]
Signature	 = ["ec4f3b28a00feabd1f58f94acb27fdc7...]
== End of Signature Verification
```


jinbpark:
a Attest/Verify library capability + a CLI tool that uses that library --> this combination is great-:

With this PR, from Certifier's perspective, they will be able to directly verify an attestation report in the case of "CCA Shim" (what they call for x86_64_SDK), without having to feed certificates into the Ceritifer's Init() API.
I think this suffices to enable the verification capability of CCA Shim to the level they want.

Though, we need to study how Veraison deals with CCA report verification (in particular, about CPAK), to get to know what are additionally needed to support a more realistic verification on CCA FVP.

jinbpark (approved):

jinbpark:
One question: 
how is this PR going to be merged into Certifier? this question is about deployment, that is to say, how to allow 3rd parties (e.g., Certifier) to use our SDK. 
Do you have a particular plan for this? A straightforward way (using Git) is to split "/sdk" into a separate repo in order for Certifier to add "SDK" as a submodule. But I don't think this is an ideal way to do it..
Maybe, if Certifier has their own rust-repo (for rust binding or whatever), they can simply refer to our SDK through Cargo.toml, which would be better.

bokdeuk-jeong (approved):
It's nice to have a sperate library for the verfication.

bitboom:
Thanks for pointing out! Deploying SDK is an important issue.

> One question: how is this PR going to be merged into Certifier? this question is about deployment, that is to say, how to allow 3rd parties (e.g., Certifier) to use our SDK.

For now, `Certifier` downloads `Islet` as archive [at here](https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/main/third_party/islet/setup.sh#LL36C14-L36C20).

> Maybe, if Certifier has their own rust-repo (for rust binding or whatever), they can simply refer to our SDK through their Cargo.toml, which would be better.

I think it is good to upload our SDK to public crate . Then, client who uses our sdk is able to link directly in Cargo.toml.
But.. `Certifier` does not use rust, we need to add glue code to make the library for c++.

PR #86, opened 2023-06-12
Refactor smc call

jinbpark:
#### Functional

```
pub struct Context {
    cmd: Command,
    arg: Vec<usize>,    // before:  [usize; 4]
    ret: Vec<usize>,     // before:  [usize; 8]
}
```
Previously, we used a fixed-length array for "arg" and "ret". (can take 4 arguments at most), but the number of "arg" can exceed the limit according to the RMM spec. To bring flexibility here, this PR uses `Vec<usize>` for both arg and ret.

#### Security

According to the RMM spec, each RMI command is forced to have a fixed number of "arg" and "ret". (e.g., REALM_CREATE command has to have 4 args and 1 ret). This PR takes into account this per-RMI command constraint,
so that we can detect any mistakes, not complying with RMM spec, in advance at the phase of runtime testing (i.e., tf-a-tests or ACS test)

we can also extend per-command constraint by considering validating  type or value inside argument registers. (such extension can be added into constraint.rs)

Here is how it works to get a better security.
1. declare constraints for each RMI command
```
// rmi/constraint.rs
// declare the constraint that "REC_AUX_COUNT" has to have 2 args and 2 ret.
m.insert(rmi::REC_AUX_COUNT, Constraint::new(rmi::REC_AUX_COUNT, 2, 2));
```

2. enforce the constraint at runtime
```
// mainloop.rs - dispatch()
rmi::constraint::validate |arg_num, ret_num| (
    // arg_num: the number of arguments this rmi command is allowed to have
    // ret_num:  the number of ret this rmi command is allowed to have
    // we allocate spaces for arg_num and ret_num;
    ctx.init_arg(&ret[1..arg_num]);
    ctx.resize_ret(ret_num);
)

// mainloop.rs - run()
ctx.do_rmi(|arg, ret| {   // do_rmi() copies "ret" to "arg" at the end, which simplifies codes to manage "arg" and "ret".
     handler(arg, ret, monitor);   // call a proper handler
}),
```

3. runtime panic if something violates the constrain
```
listen!(mainloop, rmi::REC_AUX_COUNT, |_, ret, _| {
        ret[0] = rmi::SUCCESS;
        ret[1] = rmi::MAX_REC_AUX_GRANULES;
        // ret[2] = 0;  --> if we try to access ret[2], which does not match with spec, it causes a runtime panic while tf-a-tests.
    });
```

## Notes

- While running tf-a-tests with this PR, I found two out-of-bound accesses that violate the RMM spec.
  But it turned out they are intentional violation for our own purpose. The two points were,
    - REALM_CREATE:  `ret[1] = id;`
    - REC_CREATE:  `ret[1] = vcpuid;`
  I added "NOTE comments" for these two cases in rmi/constraint.rs.

- About the below change:  change from spsc to VecDeque.
```
pub struct Mainloop {
    pub queue: Mutex<VecDeque<Context>>,
    pub on_event: BTreeMap<usize, Handler>,
}
```
As Context has become having "Vec" which does not implement Copy trait, it cannot be used in "spsc",
which requires type "T" to have "Copy" trait.
As "VecDeque" suffices to replace with "spsc" in the current implementation and doesn't require "Copy" trait,
I've decided to go wtih "VecDeque."
But as I left a TODO comment, we might need to think about what structure fits in here the best in the near future.

## An alternative design choice there was

For security, I've decided to go with "security by crash (runtime panic)".
But, using Rust's type system, it would be also possible to check RMI/RSI constraints at compile-time (at least for checking the number of arg/ret), which is more rusty and safer but entails a more amount of code changes.
To avoid such a big change, I've decided to use this security check.


bokdeuk-jeong (approved):
great to have more secure code!

rmm/monitor/src/utils/mod.rs
```diff
@@ -1 +1,2 @@
+#[allow(dead_code)]
 pub mod spsc;
```
bitboom, line 2: Could you please delete spsc code?

jinbpark: deleted:)

rmm/monitor/src/event/mainloop.rs
```diff
@@ -38,42 +35,48 @@ impl Mainloop {
...
+            cmd,
+            |arg_num, ret_num| {
+                let mut ctx = Context::new(cmd);
+                ctx.init_arg(&ret[1..arg_num]);
+                ctx.resize_ret(ret_num);
+                self.queue.lock().push_back(ctx);
+                info!("push success! {:#08x}", cmd);
+            },
+            || {
+                info!("something wrong! {:#08x}", cmd);
```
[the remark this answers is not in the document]
jinbpark: Oh, I forgot to remove it.. it was for debugging.. I'll delete it

rmm/armv9a/src/smc.rs
```diff
@@ -22,20 +22,31 @@ impl monitor::smc::Caller for SMC {
...
-    fn call(&self, command: usize, args: [usize; 4]) -> [usize; 8] {
+    fn call(&self, command: usize, args: &[usize]) -> [usize; 8] {
         let mut ret: [usize; 8] = [0usize; 8];
+        let mut padded_args: [usize; 8] = [0usize; 8];
+        let start = 1;
+        let end = start + args.len();
 
+        let put = |arr: &mut [usize; 8]| {
+            arr[0] = command;
+            arr[start..end].copy_from_slice(args);
```
hihi-wang, line 33: Is `args` guaranteed to not exceed 7? Otherwise, it would invoke slicing panic here.

rmm/monitor/src/event/mainloop.rs
```diff
@@ -38,42 +35,48 @@ impl Mainloop {
...
-        };
-        self.tx.send(ctx);
+        let ret = smc.call(ctx.cmd(), ctx.arg_slice());
+        let cmd = ret[0];
+
+        rmi::constraint::validate(
+            cmd,
+            |arg_num, ret_num| {
+                let mut ctx = Context::new(cmd);
+                ctx.init_arg(&ret[1..arg_num]);
```
bitboom, line 53: I think this should be fixed. As you know, it is half open range.  

```suggestion
                ctx.init_arg(&ret[1.. 1+ arg_num]);
```

Anyway, how does it passed..?

jinbpark: Correct, it works that way.
If arg_num is 3,
ret[0] is used to indicate "command", ret[1] and ret[2] are arguments.
arg[0] = ret[1];
arg[1] = ret[2];
So, ret[1..arg_num] == ret[1..3] --> this is a correct behavior.

rmm/armv9a/src/smc.rs
jinbpark: Guaranteed at least in the current code base. But I think it would be better to have an additional check to ensure that panic never happens.  I'll add it, Thanks-!

jinbpark: Reflected-: But the way to deal with error in that function is not that great, so I left a TODO comment that we need to add a more graceful way to deal with error.

bitboom (approved):
Introducing `Constraints` is nice :) I hope it will be checked at compile-time as you noted.

PR #87, opened 2023-06-13
Rec enter

bokdeuk-jeong:
1. handle exception from a realm for system register access.
2. update handlers for RTT update RMIs.
3. update handlers for RSIs.

### DIFF
--- a/rmm/armv9a/src/exception/lower/synchronous/sys_reg.rs
+++ b/rmm/armv9a/src/exception/lower/synchronous/sys_reg.rs
@@ -32,6 +32,7 @@ define_bits!(AA64PFR0, AMU[47 - 44], SVE[35 - 32]);
 define_iss_id!(ISS_ID_AA64PFR0_EL1, 3, 0, 0, 4, 0);
 
 define_sys_register!(ID_AA64PFR1_EL1);
+define_bits!(AA64PFR1, MTE[11 - 8]);
 define_iss_id!(ISS_ID_AA64PFR1_EL1, 3, 0, 0, 4, 1);
 
 // TODO: current compiler doesn't understand this sysreg
@@ -39,6 +40,20 @@ define_iss_id!(ISS_ID_AA64PFR1_EL1, 3, 0, 0, 4, 1);
 //define_iss_id!(ISS_ID_AA64ZFR0_EL1, 3, 0, 0, 4, 4);
 
 define_sys_register!(ID_AA64DFR0_EL1);
+define_bits!(
+    AA64DFR0,
+    BRBE[55 - 52],
+    MTPMU[51 - 48],
+    TraceBuffer[47 - 44],
+    TraceFilt[43 - 40],
+    PMSVer[35 - 32],
+    CTX_CMPs[31 - 28],
+    WRPs[23 - 20],
+    BRPs[15 - 12],
+    PMUVer[11 - 8],
+    TraceVer[7 - 4],
+    DebugVer[3 - 0]
+);
 define_iss_id!(ISS_ID_AA64DFR0_EL1, 3, 0, 0, 5, 0);
 
 define_sys_register!(ID_AA64DFR1_EL1);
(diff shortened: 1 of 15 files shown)


rmm/armv9a/src/exception/lower/synchronous/sys_reg.rs
```diff
@@ -120,15 +149,21 @@ fn handle_sysreg_id(vcpu: &mut VCPU<Context>, esr: u64) -> u64 {
         ISS_ID_AA64PFR0_EL1 => unsafe { ID_AA64PFR0_EL1.get() & mask },
```
jinbpark, line 149:
```suggestion
ISS_ID_AA64PFR0_EL1 => { ID_AA64PFR0_EL1.get() & mask },
```

I think we can eliminate a number of "unsafe" blocks that access to system registers, by putting "unsafe" inside the function of `define_sys_register! get()`, like below.
```
pub fn get(&self) -> u64 {
        let rtn;
        unsafe {
                    core::arch::asm! {
                        concat!("mrs {}, ", stringify!($regname)),
                        out(reg) rtn
                    }
           }
           rtn
 }
```
It would look more neat but I don't think this change is necessary. (i.e., whether to change or not is up to you-:)

rmm/monitor/src/rsi/constraint.rs
```diff
@@ -72,15 +72,15 @@ lazy_static! {
...
     T: FnMut(usize, usize),
-    G: FnOnce(),
 {
     if let Some(c) = CONSTRAINTS.get(&cmd) {
-        // TODO: command-specific validation routine if needed
         ok_func(c.arg_num, c.ret_num);
     } else {
-        else_func();
+        // rsi.dispatch takes care of unregistered command.
+        // Just limit the array size here.
```
jinbpark, line 83: A nice note here-:

jinbpark (approved):

bitboom (approved):

rmm/armv9a/src/exception/lower/synchronous/sys_reg.rs
bokdeuk-jeong: I read some comments in a paper about the 'unsafe' block: it is perferred and safer that exposing unsafe to the function definition rather than internally digesting it inside a function so as for the caller to handler properly. Based on that, exposing unsafe at get() looks better. However, I don't think the current caller in sys_regs.rs handles it properly. I hope we can disscuss how to handle the unsafe blocks after CC summit.

rmm/monitor/src/rmi/rec/mod.rs
```diff
@@ -118,29 +118,59 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
-                    );
-                }
-                _ => ret[0] = rmi::SUCCESS,
-            },
-            Err(_) => ret[0] = rmi::ERROR_REC,
-        };
+        let mut ret_ns;
+        loop {
+            ret_ns = true;
+            match rmi.run(rec.rd.id(), rec.id(), 0) {
```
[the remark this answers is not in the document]
bokdeuk-jeong: We are not passing rec.rd.id() and rec.id() to EL1. We are passing it to a function in RMM which performs the contexts switch.
> Is '4 (4 arguments and returns at most)' a fixed number specified in RMM spec? or is it just an arbitrary number for our convenience?
Sorry, I don't understand your question.

rmm/armv9a/src/exception/lower/synchronous/sys_reg.rs
jinbpark: > it is perferred and safer that exposing unsafe to the function definition

totally makes sense

rmm/monitor/src/rmi/rec/mod.rs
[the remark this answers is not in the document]
jinbpark: I had to admit my question was a bit confusing..
My question had to do with the following code.
```
// in rmm/armv9a/src/realm/registry.rs
fn run(&self, id: usize, vcpu: usize, incr_pc: usize) -> Result<[usize; 4], &str>
```

So, interpreting my question again here,
`Result<[usize; 4], &str>` --> I was wondering if this '4' is a fixed number specified in RMM spec, or is it for our convenience?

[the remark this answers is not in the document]
bokdeuk-jeong: I haven't looked into it until you asked. :) According to the RMM spec doc, the length of the RMI output varies upto 8, i.e. , using 8 registers. RSI_MEASUREMENT_READ uses 8. Because RMIs are delivered via smc calls, the number of arguments and return values are limited by the SMCCC spec.

PR #88, opened 2023-06-15
Replace worktree to submodule 

bitboom:
This PR changes the way of managing third-party from `worktree` to `submodule`.
The main reason is that ISLET is too heavy since `worktree` manages third-parties as branch.

So the developers who visit ISLET first could not explore easily.

Now application developers, like apps for certifier, do not need to clone whole third-party projects.
But platform developers still have to clone whole third-party projects via submodule.

## Patches to third-party projects.
I will delete third-party branches in this repo. Those were moved at [here](https://github.com/Samsung/islet-asset).
So if you need to add patches for third-parities, please add them to `islet-asset`.

## Caution
I recommend you to setup ISLET newly after this PR merged.

_P.S I disabled AOSP CI temporarily. The cause of failure seems to relate external toolchains._


bokdeuk-jeong (approved):
nice decision with your prompt and adaptive response to the requirements necessary for development!

jinbpark (approved):

PR #89, opened 2023-06-19
Enable networking between Realm and remote hosts

jinbpark:
### What this PR does

- With this PR, FVP Host and Realm are able to communicate through PC_Host's network to remote hosts. (I tested it with a command `wget http://49.247.7.63/` to see if it can download index.html in the remote server)
- But, this has been only confirmed in wired network not wireless network. I'll test the same thing again on wireless, once a new laptop gets available for the demo.

### Future work of network configuration

- To enable remote networking, there must be at least one real IP address required. (previously, we only used dummy IP address for local networking, which would be better for testing purpose)
- I think it would definitely be better to have two different networking options, (1) remote networking and (2) local networking (good enough for testing purpose), and local networking is going to be a default option.

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -16,7 +16,7 @@ First of all, make sure you are in the root directory of ISLET and go throuth th
 In most cases, you would probably not have to customize network-releated arguments and feed them into `fvp-cca`. Using a default configuration would be sufficient.
 ```
 # full command:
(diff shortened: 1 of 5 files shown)


bitboom (approved):
AOSP CI is disabled at https://github.com/Samsung/islet/pull/88. The cause of failure seems to relate external toolchains.

PR #102, opened 2023-07-04
Rsi set ripas

bokdeuk-jeong:
1. update rsi event handler parameter
2. Record requests with RSI_IPA_STATE_SET and its results from RTT_INIT_RIPAS in Rec struct.
3. Move handlers in rec/mod.rs to rec/handlers.rs

### DIFF
--- a/rmm/monitor/src/event/rsihandle.rs
+++ b/rmm/monitor/src/event/rsihandle.rs
@@ -3,6 +3,7 @@ extern crate alloc;
 use super::Context;
 
 use crate::rmi::rec::run::Run;
+use crate::rmi::rec::Rec;
 use crate::rsi;
 use crate::rsi::psci;
 use crate::Monitor;
@@ -10,7 +11,7 @@ use crate::Monitor;
 use alloc::boxed::Box;
 use alloc::collections::btree_map::BTreeMap;
 
-pub type Handler = Box<dyn Fn(&[usize], &mut [usize], &Monitor, &mut Run)>;
+pub type Handler = Box<dyn Fn(&[usize], &mut [usize], &Monitor, &mut Rec, &mut Run)>;
 
 pub struct RsiHandle {
     pub on_event: BTreeMap<usize, Handler>,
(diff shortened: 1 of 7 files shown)


zpzigi754:
Is there any reason to move `Rec` from `rec/mod.rs` to `rec/rec.rs`?
Maintaining the same name for the folder and the file might not be desirable in terms of duplication and hierarchy.

BTW, moving the rec directory to a higher location might be desirable (current: `monitor/src/rmi/rec/` suggestion: `monitor/src/rec/`), as rec is a common data structure accessed by multiple entities (e.g., `rmi`, `rsi`).

bokdeuk-jeong:
> Is there any reason to move `Rec` from `rec/mod.rs` to `rec/rec.rs`? Maintaining the same name for the folder and the file might not be desirable in terms of duplication and hierarchy.
> 
> BTW, moving the rec directory to a higher location might be desirable (current: `monitor/src/rmi/rec/` suggestion: `monitor/src/rec/`), as rec is a common data structure accessed by multiple entities (e.g., `rmi`, `rsi`).

Existing rec/mod.rs contains both struct Rec and event handlers for RMI_REC_OOOO requests.  It is expected that both struct Rec and handlers will be much bigger when we complete the implementation. So I thought that it would be more readable if they are splitted into different source files. 

Moving struct Rec out of rmi sounds reasonable.

zpzigi754:
> Existing rec/mod.rs contains both struct Rec and event handlers for RMI_REC_OOOO requests. It is expected that both struct Rec and handlers will be much bigger when we complete the implementation. So I thought that it would be more readable if they are splitted into different source files.

I agree with the intention of the split. How about moving the handlers into another file like `handlers.rs`? Then, it would be able to avoid creating the same name of file and directory.

bokdeuk-jeong:
> > Existing rec/mod.rs contains both struct Rec and event handlers for RMI_REC_OOOO requests. It is expected that both struct Rec and handlers will be much bigger when we complete the implementation. So I thought that it would be more readable if they are splitted into different source files.
> 
> I agree with the intention of the split. How about moving the handlers into another file like `handlers.rs`? Then, it would be able to avoid creating the same name of file and directory.

I think you opened up a good agenda to discuss. struct Rd and struct Rec are used by both rmi and rsi. 
However, I will move event handlers to handlers.rs and take struct Rec back to mod.rs for now. Thanks!

bokdeuk-jeong:
last commit ammend is not applied to PR. I will close this and open another.

PR #103, opened 2023-07-05
Rsi set ripas

bokdeuk-jeong:
update rsi event handler parameter
Record requests with RSI_IPA_STATE_SET and its results from RTT_INIT_RIPAS in Rec struct.
Move handlers in rec/mod.rs to rec/handlers.rs

### DIFF
--- a/rmm/monitor/src/rmi/rec/mod.rs
+++ b/rmm/monitor/src/rmi/rec/mod.rs
@@ -18,6 +18,14 @@ extern crate alloc;
 pub struct Rec {
     pub rd: &'static Rd,
     vcpuid: usize,
+    ripas: Ripas,
+}
+
+struct Ripas {
+    start: u64,
+    end: u64,
+    addr: u64,
+    state: u8,
 }
 
 impl Rec {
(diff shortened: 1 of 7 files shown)


rmm/monitor/src/rmi/rec/mod.rs
```diff
@@ -131,11 +131,14 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
 
                         rsi::constraint::validate(cmd, |_, ret_num| {
                             let mut rsi_ctx = Context::new(cmd);
-                            rsi_ctx.init_arg(&[rec.rd.id(), rec.id()]);
+                            let rec_ref =
+                                unsafe { ManuallyDrop::<&mut Rec>::into_inner(Rec::into(arg[0])) };
```
zpzigi754: Not about this PR, but asking out of curiosity. Is there any reason that we use `ManuallyDrop` for the Rec structure? (I have noticed that many data structures in islet uses `ManuallyDrop`).

zpzigi754 (approved):

rmm/monitor/src/rmi/rec/mod.rs
bokdeuk-jeong: struct Rd, Rec, Run, and Param are created with the specific memory pages that are shared with or delgated from the normal world. The structs shouldn't be dropped when variables go out of scope as they will be used again even if RMM finishes handling RMI requests and returning to the normal world.  By wrapping it with ManuallyDrop, I wanted prgogrammers get to determine when to drop the structures,( e.g, when a vcpu or a realm is destroyed).  I refered struct Run which Sangwan implemented. :)

zpzigi754: Thank you for the explanation! I guess that if we register the structures in the global registry and deregister them only when it should be dropped, we could avoid using `ManuallyDrop`.

jinbpark: > we could avoid using ManuallyDrop.

@zpzigi754 FYI,
I'll do this job at https://github.com/Samsung/islet/issues/107.
The goal of this would be to figure out an idiomatic way to replace ManuallyDrop.

Issue #107, opened 2023-07-07
Replace ManuallyDrop with a more idiomatic Rust way


jinbpark:
Replace ManuallyDrop with a more idiomatic Rust way

PR #109, opened 2023-07-11
Introduce HostPointer and Accessor

jinbpark:
## Goal

The goal of this PR is to define a common way to access **physical region allocated by the host** (i.e., shared between the host and the rmm) in a secure manner.
All places that use `ManuallyDrop` have to do with that physical region. So, to replace `ManuallyDrop`, it's required as a groundwork to define that interface.
- relevant issue:  #107 

## HostPointer

`struct Pointer` in host/pointer.rs is HostPointer which points to the physical region. To securely map and access that region, this PR uses Rust's guard pattern.
`struct PointerGuard` is a guard for `Pointer`. Only when all security checks are passed, it can get this guard and thus dereference the original pointer. 
See code comments for more detail.

## Accessor trait

To securely access different kinds of `HostPointer`, this PR defines `Accessor` trait.
Any structs that involves shared physical regions have to implement this trait and
specify validation checks required for themselves.
See the example --> `rec/params.rs - impl Accessor for Params`.

## Example:  Params in REALM_CREATE

- before
```
listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| {
    let params_ptr = arg[1];          // (1) this is the physical region that contains "Params" for REALM_CREATE
    mm.map(params_ptr, false);  // (2) map that region in RTT, for RMM to access it.
    let param = unsafe { Params::parse(params_ptr) };  // (3) turn that pointer to the struct of Params.
    ...  // (4) access the struct of Params
    mm.unmap(params_ptr);   // (5) unmap that region in RTT.
}
```

- after:  with this PR
```
listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| {
    let params_ptr = HostPointer::<Params>::new(arg[1] as *const Params, mm);   // (1) create a HostPointer
    let params = params_ptr.acquire();   // (2) try to get a guard to HostPointer. it does several validation checks.
                                                                    // it calls functions in the Accessor trait for validation checks.
    if params.is_none() { ... }   // (3) if something wrong (e.g., map fail), we handle the error.
    ...
    let params = params.unwrap();   // (4) get the guard and access the struct of Params.
    ...
    ...  // (5) don't need to call unmap() explicitly. it will be automatically invoked when dropped.
}
```

## TODO

- apply HostPointer to more number of structs. (e.g., Rd, Rec, ...)
- eliminate ManuallyDrop in Params (is it necessary??)

### DIFF
--- a/rmm/monitor/src/host/mod.rs
+++ b/rmm/monitor/src/host/mod.rs
@@ -4,16 +4,22 @@ use crate::rmm::PageMap;
 
 /// This trait is used to enforce security checks for physical region allocated by the host.
 pub trait Accessor {
-    /// Try to do page-relevant stuff (e.g., RTT map).
+    /// Try to do page-relevant stuff (e.g., RMM map).
     /// returns true only if everything goes well.
-    fn acquire(ptr: usize, page_map: PageMap) -> bool;
+    fn acquire(ptr: usize, page_map: PageMap) -> bool {
+        page_map.map(ptr, false)
+    }
 
     /// Try to clean up page-relevant stuff done by `acquire`.
     /// Structs that implement this trait must synchronize this function with `acquire`.
     /// returns true only if everything goes well.
-    fn release(ptr: usize, page_map: PageMap) -> bool;
+    fn release(ptr: usize, page_map: PageMap) -> bool {
+        page_map.unmap(ptr)
+    }
 
     /// Validate each field in a struct that implements this trait.
     /// returns true only if everything goes well.
-    fn validate(&self) -> bool;
+    fn validate(&self) -> bool {
+        true
+    }
 }
--- a/rmm/monitor/src/host/mod.rs
+++ b/rmm/monitor/src/host/mod.rs
@@ -1,0 +1,19 @@
+pub mod pointer;
+
+use crate::rmm::PageMap;
+
+/// This trait is used to enforce security checks for physical region allocated by the host.
+pub trait Accessor {
+    /// Try to do page-relevant stuff (e.g., RTT map).
+    /// returns true only if everything goes well.
+    fn acquire(ptr: usize, page_map: PageMap) -> bool;
+
+    /// Try to clean up page-relevant stuff done by `acquire`.
+    /// Structs that implement this trait must synchronize this function with `acquire`.
+    /// returns true only if everything goes well.
+    fn release(ptr: usize, page_map: PageMap) -> bool;
+
+    /// Validate each field in a struct that implements this trait.
+    /// returns true only if everything goes well.
+    fn validate(&self) -> bool;
+}
--- a/rmm/monitor/src/host/pointer.rs
+++ b/rmm/monitor/src/host/pointer.rs
@@ -1,0 +1,67 @@
+use crate::host::Accessor as HostAccessor;
+use crate::rmm::PageMap;
+use core::ops::Deref;
+
+/// Type for holding an immutable pointer to physical region allocated by the host
+#[repr(C)]
+pub struct Pointer<T: HostAccessor> {
+    /// pointer to phyiscal region
+    ptr: *const T,
+    /// page_map to map or unmap `ptr` in RTT
+    page_map: PageMap,
+}
+
+impl<T: HostAccessor> Pointer<T> {
+    /// Creates a new pointer pointing to data shared between the host and RMM
+    pub fn new(ptr: *const T, page_map: PageMap) -> Self {
+        Self { ptr, page_map }
+    }
+
+    /// Checks if this pointer is valid. It goes through two validations.
+    ///   (1) T::acquire(): this function is used to validate page-relevant stuff (e.g., RTT map/unmap, GranuleState)
+    ///   (2) T::validate():  this function is used to validate each field in T. (e.g., a constraint on parameter value)
+    /// It returns a guard object only if it passes the two steps.
+    #[inline]
+    pub fn acquire<'a>(&'a self) -> Option<PointerGuard<'a, T>> {
+        if T::acquire(self.ptr as usize, self.page_map) {
+            let guard = PointerGuard { ptr: self };
+            if !guard.validate() {
+                None
+            } else {
+                Some(guard)
+            }
+        } else {
+            None
+        }
+    }
+}
+
+/// Guard for `Pointer`
+pub struct PointerGuard<'a, T: HostAccessor> {
+    ptr: &'a Pointer<T>,
+}
+
+impl<'a, T: HostAccessor> PointerGuard<'a, T> {
+    fn validate(&self) -> bool {
+        let inner = unsafe { &*self.ptr.ptr };
+        inner.validate()
+    }
+}
+
+impl<'a, T: HostAccessor> Deref for PointerGuard<'a, T> {
+    type Target = T;
+
+    /// Safety: this is safe because 
+    /// the only safe way to get this `PointerGuard` is through `Pointer::acquire` method,
+    /// and after the validation, it is safe to dereference the original pointer.
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*self.ptr.ptr }
+    }
+}
+
+impl<'a, T: HostAccessor> Drop for PointerGuard<'a, T> {
+    /// Automatically clean up page-relevant stuff we did in `acquire()`.
+    fn drop(&mut self) {
+        T::release(self.ptr.ptr as usize, self.ptr.page_map);
+    }
+}
--- a/rmm/monitor/src/host/pointer.rs
+++ b/rmm/monitor/src/host/pointer.rs
@@ -7,24 +7,27 @@ use core::ops::Deref;
 pub struct Pointer<T: HostAccessor> {
     /// pointer to phyiscal region
     ptr: *const T,
-    /// page_map to map or unmap `ptr` in RTT
+    /// page_map to map or unmap `ptr` in RMM
     page_map: PageMap,
 }
 
 impl<T: HostAccessor> Pointer<T> {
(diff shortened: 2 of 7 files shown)


rmm/monitor/src/host/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+/// This trait is used to enforce security checks for physical region allocated by the host.
+pub trait Accessor {
+    /// Try to do page-relevant stuff (e.g., RTT map).
+    /// returns true only if everything goes well.
+    fn acquire(ptr: usize, page_map: PageMap) -> bool;
+
+    /// Try to clean up page-relevant stuff done by `acquire`.
+    /// Structs that implement this trait must synchronize this function with `acquire`.
+    /// returns true only if everything goes well.
+    fn release(ptr: usize, page_map: PageMap) -> bool;
```
bitboom, line 14: How about add default implementations? I think most structs are similar.

```suggestion
    fn acquire(ptr: usize, page_map: PageMap) -> bool {
            page_map.map(ptr, false);
    }

    /// Try to clean up page-relevant stuff done by `acquire`.
    /// Structs that implement this trait must synchronize this function with `acquire`.
    /// returns true only if everything goes well.
    fn release(ptr: usize, page_map: PageMap) -> bool {
        page_map.unmap(ptr)
    }
```

rmm/monitor/src/host/mod.rs
```diff
@@ -0,0 +1,19 @@
+pub mod pointer;
+
+use crate::rmm::PageMap;
+
+/// This trait is used to enforce security checks for physical region allocated by the host.
+pub trait Accessor {
+    /// Try to do page-relevant stuff (e.g., RTT map).
```
bokdeuk-jeong, line 7: e.g, RTT <-- 
RTT refers to the realm translation table. However, the pages using ManullyDrop are mapped to the RMM page table, not to the realm s2 page table.

rmm/monitor/src/rmi/rec/params.rs
```diff
@@ -31,6 +30,21 @@ impl Drop for Params {
     }
 }
 
+impl Accessor for Params {
+    fn acquire(ptr: usize, page_map: PageMap) -> bool {
+        page_map.map(ptr, false)
```
bokdeuk-jeong, line 35: We will call acquire() and release() every time we need to access the data structure in it, right? As most RMI handlers access Rd and Rec, if we call map() and unmap() inside acquire() and release(), it means we need to update RMM page table whenever RMI/RSI is processed in RMM. It costs too much. Rd (Ream descriptor), Rec (Realm Execution Context) needs to be mapped only once and stay mapped until Rd or Red is destoyed.

Is there a way we keep it mapped once and use HostPointer?

rmm/monitor/src/host/mod.rs
jinbpark: Nice suggestion, I'll reflect that right away

rmm/monitor/src/host/mod.rs
jinbpark: Right, I'll fix those comments.

rmm/monitor/src/rmi/rec/params.rs
jinbpark: > We will call acquire() and release() every time we need to access the data structure in it

No, they are called only once in one RMI handler function, as they were.
(at least for these two Params struct) 

> Rec (Realm Execution Context) needs to be mapped only once and stay mapped until Rd or Red is destoyed.

I already noticed that. I know Rec/Rd/... are a bit different case than Params as you said.
(FYI: in TF-RMM, map/unmap happen in one RMI function even for Rec/Rd, they don't keep mappings until destroyed. But this is not specified behavior in the spec. So, I agree with our current approach.)

> Is there a way we keep it mapped once and use HostPointer?

Yes, I'm already working on this to apply HostPointer to structures (Rec/Rd/...) that need to stay mapped.
That PR will be uploaded soon after this PR gets merged.
(How to do this in one sentence:  not call map/unmap in acquire/release, instead call them explicitly)
(+ though, i'll think about a better way to do this)

rmm/monitor/src/host/mod.rs
jinbpark: Reflected:

rmm/monitor/src/host/mod.rs
jinbpark: Fixed:

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -20,16 +21,19 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| {
         let rmi = rmm.rmi;
         let mm = rmm.mm;
-        let params_ptr = arg[1];
-        mm.map(params_ptr, false);
+        let params_ptr = HostPointer::<Params>::new(arg[1] as *const Params, mm);
```
bokdeuk-jeong, line 24: I think "arg[1] as *const Params" is misleading.  (The exising name 'params_ptr' is wrong though.)
XXX as *const YYY seems to indicate that we can access the object with the casting. Meanwhile it can be only accessed as a pointer after mapping to the RMM table.  
To be exact, how about handing over a u64 address value to the new() and coverting it into a struct inside Pointer or PointerGuard? If the implementation is conventional, never mind. :)

jinbpark: > how about handing over a u64 address value to the new() and coverting it into a struct inside Pointer or PointerGuard?

That seems more neat and tidy. I'll try it right away.

bitboom (approved):
Nice to unify interfaces to access data structures of host

rmm/monitor/src/rmi/realm/mod.rs
jinbpark: Reflected:

rmm/monitor/src/rmi/rec/handlers.rs
```diff
@@ -18,13 +19,20 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
         }
-        mm.map(params_ptr, false);
+
+        let params_ptr = HostPointer::<Params>::new(arg[2], mm);
+        let params = params_ptr.acquire();
+        if params.is_none() {
+            error!("access Params fail");
+            return;
+        }
+        let params = params.unwrap();
```
zpzigi754, line 34: [minor]
How about avoiding using `unwrap()` like the below?
```
let params = if let Some(val) = params {
    val
} else {
    error!("access Params fail");
    return;
}
```

jinbpark: Reflected:  (avoiding unwrap() is always good)

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -20,16 +21,19 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
         let mm = rmm.mm;
-        let params_ptr = arg[1];
-        mm.map(params_ptr, false);
+        let params_ptr = HostPointer::<Params>::new(arg[1], mm);
+        let params = params_ptr.acquire();
+        if params.is_none() {
+            error!("access Params fail");
+            return;
+        }
+        trace!("{:?}", *params.unwrap());
```
zpzigi754, line 30: The same comment can be applied to here (removing `unwrap()`)

rmm/monitor/src/host/mod.rs
```diff
@@ -0,0 +1,25 @@
+pub mod pointer;
+
+use crate::rmm::PageMap;
+
+/// This trait is used to enforce security checks for physical region allocated by the host.
+pub trait Accessor {
+    /// Try to do page-relevant stuff (e.g., RMM map).
+    /// returns true only if everything goes well.
+    fn acquire(ptr: usize, page_map: PageMap) -> bool {
+        page_map.map(ptr, false)
```
zpzigi754, line 10: Is there no case that `true` is used as the parameter of `map()`? (` GranuleState::RTT` seems to use mm.map(addr, true);)

rmm/monitor/src/host/pointer.rs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        } else {
+            None
+        }
+    }
+}
+
+/// Guard for `Pointer`
+pub struct PointerGuard<'a, T: HostAccessor> {
+    ptr: &'a Pointer<T>,
```
zpzigi754, line 44: Maybe we can use the name `inner` here to avoid using `ptr.ptr` in other places?

zpzigi754 (commented):
Good to avoid manually unmapping!

rmm/monitor/src/host/pointer.rs
jinbpark: > avoid using ptr.ptr

It seems more correct. I'll do it.

rmm/monitor/src/host/mod.rs
jinbpark: All structs that have to do with `GranuleState` use `true`.
As I mentioned above in response to bokdeuk, I'm working on the follow-up PR to apply HostPointer to such cases.
You can see that soon after this PR gets merged.

rmm/monitor/src/rmi/realm/mod.rs
jinbpark: Reflected:

rmm/monitor/src/host/pointer.rs
jinbpark: Reflected:

zpzigi754 (approved):

bokdeuk-jeong (approved):

PR #110, opened 2023-07-12
Update AOSP support

kanglib:
- Updates the `android-kernel` submodule
- Uses `out/shared` for the shared directory on AOSP
- Fixes the JVM internal error

`out/shared` is mounted at `/mnt/shared` on AOSP. AOSP userdata is stored as a partition on the system image.


scripts/config.py
```diff
@@ -54,7 +54,6 @@
 # directory shared between the pc desktop and the host OS on fvp using 9p filesystem
 SHARED_PATH = os.path.join(OUT, "shared")
 SHARED_EXAMPLES_PATH = os.path.join(OUT, "shared/examples/")
-NW_AOSP_DATA = os.path.join(OUT, "aosp_data")
```
bokdeuk-jeong, line 57: I think it's better to keep the shared data with aosp in a sperate directory. If you meant that you want to reflect the purpose of the share between the host PC and the fvp to the directory name, how about AOSP_SHARED_PATH?

scripts/.config
```diff
@@ -35,6 +39,10 @@ cluster0.stage12_tlb_size=1024
 cluster1.NUM_CORES=4
 cluster1.PA_SIZE=48
 cluster1.check_memory_attributes=0
+cluster1.cpu0.DCZID-log2-block-size=4
+cluster1.cpu1.DCZID-log2-block-size=4
+cluster1.cpu2.DCZID-log2-block-size=4
+cluster1.cpu3.DCZID-log2-block-size=4
```
bokdeuk-jeong, line 45: Could you explain why we need these settings? Does it degrade performance? If not, can you add some comments why we need these?

kanglib: This revives b69a62f that has been removed after the tf-rmm support. JVM may panic without those settings.

scripts/config.py
kanglib: That sounds great.

kanglib:
Note: `islet-asset@3rd-android-kernel-6.2-kanglib` (now deleted) supercedes `@3rd-android-kernel`.

bokdeuk-jeong (approved):

PR #111, opened 2023-07-13
Apply HostPointer and HostPointerMut

jinbpark:
## What this PR does

- Introduce `HostPoinerMut` which are used to access "mutable physical host region".
- `HostPointer` is used for
    - Params for REALM_CREATE
    - Params for REC_CREATE
    - Source data page for DATA_CREATE
- `HostPointerMut` is used for
    - Run  (it needs to be mutable to handle RSI. e.g., HOST_CALL handler modifies Run)
- Note
    - Rec/Rd/Data(target_data)/... --> those that involve granule states should be handled in a different way than `HostPointer`. I'll do update codes to manage those struct in a separate PR that might have to do with #113 .
- Related issue: #107 

## TO-DO

- #112 
- #113    --> this issue is likely the next to go
- #114

### DIFF
--- a/rmm/monitor/src/host/mod.rs
+++ b/rmm/monitor/src/host/mod.rs
@@ -1,8 +1,10 @@
+#[macro_use]
 pub mod pointer;
 
 use crate::rmm::PageMap;
 
 /// This trait is used to enforce security checks for physical region allocated by the host.
+/// This is used for `PointerGuard` which is not able to modify data.
 pub trait Accessor {
     /// Try to do page-relevant stuff (e.g., RMM map).
     /// returns true only if everything goes well.
@@ -23,3 +25,16 @@ pub trait Accessor {
         true
     }
 }
+
+/// DataPage is used to convey realm data from host to realm.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct DataPage([u8; 4096]);
+
+impl DataPage {
+    pub unsafe fn as_ptr(&self) -> *const u8 {
+        self.0.as_ptr() as *const u8
+    }
+}
+
+impl Accessor for DataPage {}
--- a/rmm/monitor/src/host/mod.rs
+++ b/rmm/monitor/src/host/mod.rs
@@ -1,6 +1,7 @@
 #[macro_use]
 pub mod pointer;
 
+use crate::rmm::granule::GRANULE_SIZE;
 use crate::rmm::PageMap;
 
 /// This trait is used to enforce security checks for physical region allocated by the host.
@@ -29,7 +30,7 @@ pub trait Accessor {
 /// DataPage is used to convey realm data from host to realm.
 #[repr(C)]
 #[derive(Copy, Clone)]
-pub struct DataPage([u8; 4096]);
+pub struct DataPage([u8; GRANULE_SIZE]);
 
 impl DataPage {
     pub unsafe fn as_ptr(&self) -> *const u8 {
--- a/rmm/monitor/src/host/pointer.rs
+++ b/rmm/monitor/src/host/pointer.rs
@@ -1,6 +1,6 @@
 use crate::host::Accessor as HostAccessor;
 use crate::rmm::PageMap;
-use core::ops::Deref;
+use core::ops::{Deref, DerefMut};
 
 /// Type for holding an immutable pointer to physical region allocated by the host
 #[repr(C)]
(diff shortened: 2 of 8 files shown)


rmm/monitor/src/rmi/realm/rd.rs
```diff
@@ -1,43 +1,76 @@
...
+use crate::rmm::granule::GranuleState;
+use crate::PageMap;
 
 pub struct Rd {
     realm_id: usize,
     state: State,
 }
 
 impl Rd {
-    pub unsafe fn new(rd_addr: usize, realm_id: usize) -> ManuallyDrop<&'static mut Rd> {
```
bokdeuk-jeong, line 9: For the same reason with the comment I left before, I think it is better to keep new(), but without ManuallyDrop.

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -21,21 +20,10 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
-        let params_ptr = HostPointer::<Params>::new(arg[1], mm);
-        let params = params_ptr.acquire();
-        let params = if let Some(val) = params {
-            val
-        } else {
-            error!("access Params fail");
-            return;
-        };
+        var_host_pointer!(params, Params, arg[1], mm, ret[0]);
+        var_host_pointer_mut!(rd, Rd, arg[0], mm, rmi::REALM_CREATE, ret[0]);
```
bokdeuk-jeong, line 24: There's difference between Params and Rd. params is shared between NW and RW and it already contains information in struct Param when RMM receives the physical address of the shared page, arg[1]. Meanwhile, as you already know, Rd isn't shared between NW and RW. The pysical page of Rd, arg[0], is delegated to RW just before receiving the REALM_CREATE RMI. Once the delegated page address is received with arg[0], RMM starts creating a struct Rd stuffing information that it needs.  

- It doesn't seem to be proper using the term 'host_pointer' for Rd and Rec because it doesn't point an object that a host created. 
- Considering the difference of the memory page usage, I think that holding these two semantics in a single struct type var_host_pointer(_mut) is too much.

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -21,21 +20,10 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
         trace!("{:?}", *params);
 
-        if granule::set_granule(arg[0], GranuleState::RD, mm) != granule::RET_SUCCESS {
```
bokdeuk-jeong, line 34: nice decision. 👍

rmm/monitor/src/rmi/realm/mod.rs
jinbpark: In short: `HostPointer` for Params (or other equivalent ones that do not involve granule state), `GranulePointer (??)` for Rd/Rec/Data/... (or other equivalent ones that involve granule state).

I totally agree with that. We should treat them as different pointers.
What particular name do you have in your mind???? `GranulePointer` is good enough? (it makes enough sense i think. any better name?)

rmm/monitor/src/rmi/realm/rd.rs
jinbpark: You mean, we need to keep the first type of validation?
```
let _ = unsafe { Rd::new(arg[0], id) };  // --> modify Rd
```

with accessor trait,  (maybe GranulePointer?)
```
var_granule_pointer_mut!(rd, Rd, arg[0], mm, rmi::REALM_CREATE, ret[0]);
rd.set_id(id);   // --> modify Rd
```

jinbpark:
**NOTICE**:  I will change this PR a lot (eliminate `host_pointer_mut` part) and do this in a different way in #113.

rmm/monitor/src/host/mod.rs
```diff
@@ -23,3 +28,53 @@ pub trait Accessor {
         true
     }
 }
+
+/// This trait is used to enforce security checks for physical region allocated by the host.
+/// This is used for `PointerMutGuard` which is able to modify data.
+pub trait AccessorMut {
+    /// Try to do page/granule-relevant stuff.
+    /// cmd takes a rmi command and is used to offer further access control.
+    fn acquire(ptr: usize, page_map: PageMap, cmd: usize) -> bool;
```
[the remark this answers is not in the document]
jinbpark: `AccessorMut` --> this will be eliminated when this PR becomes ready for review again.  `Accessor` will be used for both `HostPointer` and `HostPointerMut`.

rmm/monitor/src/rmi/realm/mod.rs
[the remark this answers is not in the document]
jinbpark: > using many parameters

That's what I was worried about too. ("cmd" will be eliminated)
So, the final form would look like `var_host_pointer!(params, Params, arg[1], mm, ret[0]);` --> 5 parameters.
Oh.. I think there is a way to get `mm` and `ret[0]` away. I'll do it to reduce it to 3 parameters.

rmm/monitor/src/host/pointer.rs
```diff
@@ -70,3 +71,98 @@ impl<'a, T: HostAccessor> Drop for PointerGuard<'a, T> {
...
+
+impl<'a, T: HostAccessorMut> Drop for PointerMutGuard<'a, T> {
+    fn drop(&mut self) {
+        T::release(self.inner.ptr as usize, self.inner.page_map, self.inner.cmd);
+    }
+}
+
+#[macro_export]
+macro_rules! var_host_pointer {
+    ($var:ident, $target_type:tt, $ptr:expr, $page_map:expr, $ret:expr) => {
```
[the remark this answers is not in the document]
jinbpark: I'll do it.

rmm/monitor/src/rmi/realm/mod.rs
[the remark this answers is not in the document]
jinbpark: The way I thought doesn't work...  I left TODO comment about this issue. I'll bring this up again in another PR.

jinbpark:
**NOTICE:**  this pr has become ready for review. And I modified the PR description. Please check out the description.

rmm/monitor/src/host/pointer.rs
```diff
@@ -70,3 +70,96 @@ impl<'a, T: HostAccessor> Drop for PointerGuard<'a, T> {
         T::release(self.inner.ptr as usize, self.inner.page_map);
     }
 }
+
+/// Type for holding a mutable pointer to physical region allocated by the host
+#[repr(C)]
+pub struct PointerMut<T: HostAccessor> {
+    /// pointer to phyiscal region
```
bitboom, line 77: minor typo

```suggestion
    /// pointer to physical region
```

rmm/monitor/src/host/mod.rs
```diff
@@ -23,3 +25,16 @@ pub trait Accessor {
         true
     }
 }
+
+/// DataPage is used to convey realm data from host to realm.
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct DataPage([u8; 4096]);
```
bitboom, line 32: How about use `crate::rmm::ganule::GRANULE_SIZE`?

rmm/monitor/src/rmi/rec/handlers.rs
```diff
@@ -25,14 +26,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
 
-        let params_ptr = HostPointer::<Params>::new(arg[2], mm);
-        let params = params_ptr.acquire();
-        let params = if let Some(val) = params {
-            val
-        } else {
-            error!("access Params fail");
-            return;
-        };
+        host_pointer!(params, Params, arg[2], mm, ret[0]);
```
bitboom, line 29: I think it's better to add `_or_ret`. Because the control flow is changed when host_pointer is None.
`host_pointer_mut_or_ret` too.

```suggestion
        host_pointer_or_ret!(params, Params, arg[2], mm, ret[0]);
```

---

Later, if we enhance error handling, below form would be better .

```rust
let params = host_pointer!(Params, arg[2], mm)?;
trace!("{:?}", *params);
```

bitboom (approved):
Minor comments. I agree overall.

rmm/monitor/src/rmi/rec/handlers.rs
jinbpark: > let params = host_pointer!(Params, arg[2], mm)?;

Agree. I'll leave a todo comment to do it. (we can do it after modifying error handling as you said)

jinbpark:
> Minor comments. I agree overall.

All reflected:
`let params = host_pointer_or_ret!(Params, arg[2], mm)?;` --> looks really nice, I left a todo comment to do it next time (maybe with Result as a return value)

zpzigi754 (approved):

bokdeuk-jeong (approved):

Issue #106, opened 2023-07-07
Realm linux kernel booting log  (stuck point)


zpzigi754:
This is the latest log about the stuck point. I leave it for the reference.

```
[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x00000000]
[    0.000000] Linux version 6.2.0-rc1-gc865d47546ee-dirty (changho@changho) (aarch64-none-linux-gnu-gcc (Arm GNU Toolchain 11.3.Rel1) 11.3.1 20220712, GNU ld (Arm GNU Toolchain 11.3.Rel1) 2.38.20220708) #49 SMP PREEMPT Thu Jul  6 18:42:23 KST 2023
[    0.000000] Machine model: linux,dummy-virt
[    0.000000] RME: RSI version 1.0 advertised
[    0.000000] earlycon: pl11 at MMIO 0x000000001c0a0000 (options '')
[    0.000000] printk: bootconsole [pl11] enabled
[    0.000000] efi: UEFI not found.
[    0.000000] NUMA: No NUMA configuration found
[    0.000000] NUMA: Faking a node at [mem 0x0000000080000000-0x000000008fffffff]
[    0.000000] NUMA: NODE_DATA [mem 0x8ff73a00-0x8ff75fff]
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000080000000-0x000000008fffffff]
[    0.000000]   DMA32    empty
[    0.000000]   Normal   empty
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
[    0.000000]   node   0: [mem 0x0000000080000000-0x000000008fffffff]
[    0.000000] Initmem setup node 0 [mem 0x0000000080000000-0x000000008fffffff]
[    0.000000] cma: Reserved 32 MiB at 0x000000008c400000
[    0.000000] psci: probing for conduit method from DT.
[    0.000000] psci: PSCIv1.1 detected in firmware.
[    0.000000] psci: Using standard PSCI v0.2 function IDs
[    0.000000] psci: MIGRATE_INFO_TYPE not supported.
[    0.000000] psci: SMC Calling Convention v1.2
[    0.000000] percpu: Embedded 20 pages/cpu s42728 r8192 d31000 u81920
[    0.000000] Detected PIPT I-cache on CPU0
[    0.000000] CPU features: detected: GIC system register CPU interface
[    0.000000] CPU features: detected: Hardware dirty bit management
[...]

PR #115, opened 2023-07-14
Wfx exception handle

bokdeuk-jeong:
1. Add exception handlers for wfi and wfe in armv9a/src/exception/trap.rs
2. Disable TWI in HCR_EL2 (trap to EL2) and add warning message to the REC_ENTER requests in case that the normal world wants to set it enabled.
    - current design decision: I don't see a good reason we support per realm or per realm execution configuration with TWI in HCR_EL2. It is more feasible setting enble or disable for all realms consistently.
    - As current linux-cca in normal world does not handle wfx exceptions, just disable it. We currently have exception handler for wfx.  When we need such dynamic configuration in the future, let's add the hcr_el2 context management then.

### DIFF
--- a/rmm/armv9a/src/exception/trap.rs
+++ b/rmm/armv9a/src/exception/trap.rs
@@ -116,6 +116,15 @@ pub extern "C" fn handle_lower_exception(
                 advance_pc(vcpu);
                 ret
             }
+            Syndrome::WFX => {
+                debug!("Synchronous: WFx");
+                tf.regs[0] = realmexit::SYNC as u64;
+                tf.regs[1] = esr as u64;
+                tf.regs[2] = unsafe { HPFAR_EL2.get() };
+                tf.regs[3] = unsafe { FAR_EL2.get() };
+                advance_pc(vcpu);
+                RET_TO_RMM
+            }
             undefined => {
                 debug!("Synchronous: Other");
                 tf.regs[0] = realmexit::SYNC as u64;
--- a/rmm/armv9a/src/exception/trap/syndrome.rs
+++ b/rmm/armv9a/src/exception/trap/syndrome.rs
@@ -38,6 +38,7 @@ pub enum Syndrome {
     HVC,
     SMC,
     SysRegInst,
+    WFX,
     Other(u32),
 }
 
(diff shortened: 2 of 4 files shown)


jinbpark:
How far can we get to in realm linux booting? Is it going to face another stuck point than specified in #106 ??

bokdeuk-jeong:
> How far can we get to in realm linux booting? Is it going to face another stuck point than specified in #106 ??

Good question. There is no error due to the wfi instruction, but the linux booting progress stays there because we haven't implemented timer support. next thing to do is enable timer support. :)

Realm kernel will stay idle until it can schedule. For scheduling it require timer interrupt.

ffff80000892af20 <cpu_do_idle>:
ffff80000892af20:   d503233f    paciasp
ffff80000892af24:   d5033f9f    dsb sy
ffff80000892af28:   d503207f    wfi   // realm is still stuck here due to the missing timer support
ffff80000892af2c:   d50323bf    autiasp
ffff80000892af30:   d65f03c0    ret

jinbpark:
> wfi instruction

I got it

jinbpark (approved):

zpzigi754:
I didn't get the relationship between the first commit (adding handler) and the second commit (disabling TWI).
If this PR intends to add the handler for WFI/WFE, why should the relevant bit disabled which would make the handler out of reach?

If the intention is disabling the trap to EL2 regarding WFI/WFE, should we still need the handler?

bokdeuk-jeong:
> I didn't get the relationship between the first commit (adding handler) and the second commit (disabling TWI). If this PR intends to add the handler for WFI/WFE, why should the relevant bit disabled which would make the handler out of reach?
> 
> If the intention is disabling the trap to EL2 regarding WFI/WFE, should we still need the handler?

Currently disabling TWI in HCR_EL2 is enough. 
linux-cca doesn't handle WFx exception for now, but I don't think that's will be a ultimate solution. Usually when core waits for event or interrupt (using wfi, wfe), VM exits and let KVM schdule for another process. As the realm runs in the context of vcpu thread in kvmtool, it would be better to yield to its cpu time if the realm waits for such events (, unless such exit induces a security risk). 
I think  linux-cca will eventually implement wfi handler. So, I left the first commit so as to just re-enable TWI.

zpzigi754:
> Currently disabling TWI in HCR_EL2 is enough. linux-cca doesn't handle the WFx exception at the moment, but I don't think that will be the ultimate solution. Usually when core waits for event or interrupt (using wfi, wfe), VM exits and let KVM schdule for another process. As the realm runs in the context of vcpu thread in kvmtool, it would be better to yield to its cpu time if the realm waits for such events (, unless such exit induces a security risk). I think linux-cca will eventually implement wfi handler. So, I left the first commit so as to just re-enable TWI.

Thank you for the detailed answer! I understood that the handler is for the future usage.

zpzigi754 (approved):

PR #116, opened 2023-07-18
Realm timer support

bokdeuk-jeong:
1. Support realm timer at RMM
2. Deliver write value to the normal world on mmio data abort

### DIFF
--- a/rmm/armv9a/src/exception/trap.rs
+++ b/rmm/armv9a/src/exception/trap.rs
@@ -137,9 +137,10 @@ pub extern "C" fn handle_lower_exception(
         Kind::Irq => {
             debug!("IRQ");
             tf.regs[0] = realmexit::IRQ as u64;
-            tf.regs[1] = esr as u64;
+            // IRQ isn't interpreted with esr. It just hold previsou info. Void them out.
+            tf.regs[1] = 0;
             tf.regs[2] = 0;
-            tf.regs[3] = unsafe { FAR_EL2.get() };
+            tf.regs[3] = 0;
             RET_TO_RMM
         }
         _ => {
--- a/rmm/armv9a/src/realm/registry.rs
+++ b/rmm/armv9a/src/realm/registry.rs
@@ -17,6 +17,7 @@ use crate::realm::mm::stage2_translation::Stage2Translation;
 use crate::realm::mm::translation_granule_4k::RawPTE;
 use monitor::rmi::error::Error;
 use monitor::rmi::error::InternalError::*;
+use crate::realm::timer;
 
 use alloc::boxed::Box;
 use alloc::collections::BTreeMap;
@@ -83,10 +84,11 @@ impl monitor::rmi::Interface for RMI {
         let vttbr = bits_in_reg(VTTBR_EL2::VMID, id as u64)
             | bits_in_reg(VTTBR_EL2::BADDR, page_table as u64);
 
-        let _vcpu = VCPU::new(realm.clone());
-        _vcpu.lock().context.sys_regs.vttbr = vttbr;
+        let vcpu = VCPU::new(realm.clone());
+        vcpu.lock().context.sys_regs.vttbr = vttbr;
+        vcpu.lock().context.timer.cnthctl_el2 = timer::init_cnthctl();
 
-        realm.lock().vcpus.push(_vcpu);
+        realm.lock().vcpus.push(vcpu);
         let vcpuid = realm.lock().vcpus.len() - 1;
         Ok(vcpuid)
     }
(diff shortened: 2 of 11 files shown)


rmm/monitor/src/rmi/rec/run.rs
```diff
@@ -71,6 +71,22 @@ impl Run {
...
+    pub unsafe fn set_cntv_ctl(&mut self, val: u64) {
+        *(&mut (*(*self.exit.inner).cnt.inner).v_ctl) = val;
+    }
+
+    pub unsafe fn set_cntv_cval(&mut self, val: u64) {
+        (*(*self.exit.inner).cnt.inner).v_cval = val;
+    }
+
+    pub unsafe fn set_cntp_ctl(&mut self, val: u64) {
+        (*(*self.exit.inner).cnt.inner).v_ctl = val;
```
bitboom, line 84:
```suggestion
        (*(*self.exit.inner).cnt.inner).p_ctl = val;
```

rmm/monitor/src/rmi/rec/run.rs
```diff
@@ -71,6 +71,22 @@ impl Run {
...
+    }
+
+    pub unsafe fn set_cntp_cval(&mut self, val: u64) {
+        (*(*self.exit.inner).cnt.inner).v_cval = val;
```
bitboom, line 88:
```suggestion
        (*(*self.exit.inner).cnt.inner).p_cval = val;
```

jinbpark:
I don't know much about how modern virtual timer works. So I took a quick look at what it is and this PR code.
Could you confirm if my understanding is correct at a high level?
- RMM should manage Realm timer registers (that belongs to Realm, right? according to the spec, Realm has own set of timer registers)
- When Host issues RMI_REC_ENTER, realm gets executed and *virtual time (time in each Realm)* flows.
- When realm execution exists and gets back to RMM for some reason, RMM has to send *each Realm's timer state* to Host (via `rmi.send_timer_state_to_host()`). It's because Host is something that checks if timer is expired or not.
- Host receives *timer state* and injects a virtual timer interrupt if it is expired into RMM.
- Then, RMM can deliver that timer interrupt to the Realm, and the Realm can get out of *WFI* and keep going forward.

rmm/armv9a/src/realm/registry.rs
```diff
@@ -340,4 +342,63 @@ impl monitor::rmi::Interface for RMI {
...
+
+        let esr_el2 = context.sys_regs.esr_el2;
+        let esr = EsrEl2::new(esr_el2);
+        let isv = esr.get_masked_value(EsrEl2::ISV);
+        let ec = esr.get_masked_value(EsrEl2::EC);
+        let wnr = esr.get_masked_value(EsrEl2::WNR);
+        let rt = esr.get_masked_value(EsrEl2::SRT) as usize;
+
+        // wnr == 1: caused by writing
+        if ec != ESR_EL2_EC_DATA_ABORT || isv == 0 || wnr == 0 {
```
jinbpark, line 360: It seems that for mmio data aborts we leave the task of read/write to Host and wait for Host to complete it.
Is it a behavior in RMM spec (I don' think it is kind of standard stuff imo..)??  How does TF-RMM handle this case??

Apart from that, is there any possibility that some private information in Realm is leaked through gprs in Run?
```
struct _Inner {
    imm: u16,
    gprs: [u64; HOST_CALL_NR_GPRS],  // does this info leak some information to Host??
}
```

bokdeuk-jeong: You are absolutely right. I also had the same concern this morning. 
The existing cca reference project does the same thing, but this doesn't mean it is okay. Delivering an output value which is in a register for writing a value into a target memory address should be only happen if it is mmio. We need to add a check for that. However, I would like to focus on making timer interrupt work in the PR.

bokdeuk-jeong:
I found out that the timer interrupt handler in a realm is not invoked even after a virq injection at the normal world. I need  more time investigate on this. I am taking the status of this PR back to draft.

rmm/armv9a/src/realm/registry.rs
jinbpark: #119 --> I made a separate issue for these security concerns.  We need to look at it in the future. Maybe, it can be a point where islet is better than tf-rmm in terms of security hardening.

bokdeuk-jeong:
I checked that the timer irq handler, arch_timer_handler_virt(), is invoked.

rmm/armv9a/src/exception/trap.rs
```diff
@@ -137,9 +137,10 @@ pub extern "C" fn handle_lower_exception(
         Kind::Irq => {
             debug!("IRQ");
             tf.regs[0] = realmexit::IRQ as u64;
-            tf.regs[1] = esr as u64;
+            // IRQ isn't interpreted with esr. It just hold previsou info. Void them out.
```
zpzigi754, line 140: [minor typo]
previsou -> previous

rmm/monitor/src/rmi/rec/handlers.rs
```diff
@@ -131,6 +131,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
                         run.set_esr(val[1] as u64);
                         run.set_hpfar(val[2] as u64);
                         run.set_far(val[3] as u64);
+                        let _ = rmi.send_mmio_write(rec.rd.id(), rec.id(), &mut run);
```
zpzigi754, line 130: Now we could use try (`?`) operator. 
[suggested change] 
rmi.send_mmio_write(rec.rd.id(), rec.id(), &mut run)?;

zpzigi754 (approved):

rmm/monitor/src/rmi/rec/handlers.rs
bokdeuk-jeong: I missed that. thansk!

jinbpark (approved):

PR #121, opened 2023-07-19
Update granule management

jinbpark:
## What this DRAFT PR for

**This DRAFT PR is meant to share possible approaches to manage GranuleStatusTable (GST) in RMM and get some feedbacks**. Note that the codes in this draft PR are subject to change. (overall, they are not ready-for-review quality).
Among below approaches, this PR implements Approach-2, which is I prefer at this point. (Approach-3 would be nice but it takes more time to implement)

### Approach-1:  a static one-level table

```
struct GranuleStatusTable {
    granules: [SpinLock<Granule>; MAX_NUM_GRANULES],
    // MAX_NUM_GRANULES = RAM_SIZE / 4096;
}
```
- we don't need any locking mechanism in accessing GranuleStatusTable.
- pros: simple implementation. we don't need anything other than per-granule locking.
- cons: a large memory footprint by default. For 4GB RAM and 8-byte Granule entry, it takes up **8mb by default**.

### Approach-2:  a static two-level table (what this PR implements) not considering table deletion

```
struct GranuleStatusTable {  // This is a root table (first-level table)
    sub_tables: [GranuleStatusSubTable; NUM_SUB_TABLES],
    // It contains multiple second-level tables. (sub tables).
    // We can configure sub_table_size. In this PR, I set 8mb to it.
}
struct GranuleStatusSubTable {  // This is a sub table (second-level table)
    active: AtomicBool,
    granules: Option<Vec<Spinlock<Granule>>>,
    // granules are created on demand when the first granule in this sub-table is asked to get active.
}
```
- this is similar to a typical page table lookup method, which means we could extend it to three-level table if needed.
- it requires one more atomic-based concurrency mechanism to check if `granules` in SubTable exists.
- pros: simple implementation. this atomic-based concurrency is simple enough because it doesn't consider table deletion. a reasonable memory overhead:  for 4gb ram and 4mb sub-table and 8byte granule entry,  it takes up **8kb by default** and takes **8kb more** when a new sub table is created.
- cons:  even when no granule in a sub table is used, that sub table still takes up memory. (on-demand table deletion is not supported)

### Approach-3:  a static two-level table considering table deletion

- same as Approach-2 but supports on-demand sub-table deletion
- same memory overhead as Approach-2.
- cons: it requires more atomic variables and more complex codes to ensure concurrency on sub tables. (would be hard to read and formally verify, but possible though)

### Approach-4:  a dynamic BTreeMap with a big lock

```
struct GranuleStatusTable {
    granules: Spinlock<BTreeMap<usize, Granule>>,
}
```
- this is our current implementation.
- pros:  zero memory consumption by default. 
- cons:  it requires a big lock for a whole GranuleStatusTable,  not per-granule locking.  In the worst case, the lookup speed would be slower than Approach-1/2/3.

### Approach-5:  a dynamic lock-free BTreeMap

```
struct GranuleStatusTable {
    granules: LockFreeBTreeMap<usize, SpinLock<Granule>>>,
}
```
- we might be able to use LockFreeBTreeMap to add "per-granule locking" to Approach-4.
- but I couldn't find a rust-based lock free btree online...  even if we find it, I don't prefer this option as LockFreeBTreeMap may be super complicated..

### DIFF
--- a/rmm/board/fvp/src/main.rs
+++ b/rmm/board/fvp/src/main.rs
@@ -9,12 +9,14 @@
 extern crate log;
 
 mod entry;
+mod memory;
 
 use armv9a::allocator;
 use armv9a::cpu;
 use armv9a::helper;
 
 use monitor;
+use memory::FVPGranuleMap;
 
 #[no_mangle]
 pub unsafe fn main() -> ! {
@@ -29,6 +31,9 @@ pub unsafe fn main() -> ! {
     let rmm = armv9a::rmm::MemoryMap::new();
     let monitor = monitor::Monitor::new(rmi, smc, rmm);
     let mut mainloop = monitor::event::Mainloop::new();
+    let granule_map = FVPGranuleMap::new();
+    monitor::rmm::granule::create_gst(granule_map);
+
     mainloop.boot_complete(smc);
     mainloop.run(&monitor);
 
--- a/rmm/board/fvp/src/memory.rs
+++ b/rmm/board/fvp/src/memory.rs
@@ -1,0 +1,42 @@
+const FVP_DRAM0_REGION: core::ops::Range<usize> = core::ops::Range {
+    start: 0x8000_0000,
+    end: 0x8000_0000 + 0x7C00_0000 - 1,
+};
+const FVP_DRAM1_REGION: core::ops::Range<usize> = core::ops::Range {
+    start: 0x8_8000_0000,
+    end: 0x8_8000_0000 + 0x8000_0000 - 1,
+};
+const GRANULE_SIZE: usize = 4096;
+
+#[derive(Debug)]
+pub struct FVPGranuleMap;
+impl FVPGranuleMap {
+    pub fn new() -> &'static FVPGranuleMap {
+        &FVPGranuleMap {}
+    }
+}
+
+impl monitor::rmm::granule::GranuleMemoryMap for FVPGranuleMap {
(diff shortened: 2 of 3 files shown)


bitboom:
I highly prefer static multi-level page table (Approach 2-3) . Could we re-use & enhance previous [page table ](https://github.com/Samsung/islet/blob/main/rmm/monitor/src/mm/page_table/mod.rs)? If it is possible, we can make it to library like `islet/lib/page_table`. And we can apply it both rmm page table and granule table. I like modularization :)

jinbpark:
> I highly prefer static multi-level page table (Approach 2-3) .

As of now, my plan is to implement Approach-2 and get reviews and merged, and update it to Approach-3 when needed. (maybe when our RMM starts dealing with more realistic cases) I thought about how to provider perfect concurrency mechanism for Approach-3 and concluded that is not trivial. 

>  Could we re-use & enhance previous page_table

Nice idea. I took a quick look at this idea to estimate how much time and effort it would take.
I think it's possible with some modification into our current page_table stuff.
Then, 
- move a unified interface into lib:  `islet/lib/page_table`
- three use cases:  (1) RMM page_table,  (2) RTT (realm),   (3) GST (granule status)
- key difference between (1/2) and (3):  Entry type, locking/concurrency mechanism

It would take more time but worth a try.
I'll go this way of reusing our page_table.

zpzigi754:
Currently, I prefer the approach 2, because paging which requires the Host's involvement does not seem to be supported in the near time in linux-cca (even the spec is missing).

PR #127, opened 2023-07-21
script: Print the realm's log in uart1

zpzigi754:
This PR enables to print the realm's log in uart1 (@bokdeuk-jeong's configuration). The right `launch-realm.sh` will be prepared depending on the fvp-cca's rmm mode (`tf-rmm` or `islet`).


bitboom:
> For the default option, which one would be better: `tf-rmm` or `islet`?

I think it's better to have two scripts (`launch-realm-tf-rmm.sh` `launch-realm-islet.sh`).
And copy it to `launch-realm.sh` on `fvp-cca script` according to rmm's type, which mean launching fvp time.
 Then we don't need to consider it.

zpzigi754:
> > For the default option, which one would be better: `tf-rmm` or `islet`?
> 
> I think it's better to have two scripts (`launch-realm-tf-rmm.sh` `launch-realm-islet.sh`). And copy it to `launch-realm.sh` on `fvp-cca script` according to rmm's type, which mean launching fvp time. Then we don't need to consider it.

Thank you for the good suggestion. I wanted to remove the duplicate portion with the one file, yet the suggestion seems better in terms of usability. I will reflect it!

bitboom (approved):

Issue #129, opened 2023-07-24
Fix SMC Context bug


bitboom:
Below [line](https://github.com/Samsung/islet/blob/main/rmm/monitor/src/event/mainloop.rs#L75) should be fixed to set to `x0`, which mean arg0.

```rust
            match self.on_event.get(&ctx.cmd) {
                Some(handler) => {
                    let res = ctx.do_rmi(|arg, ret| handler(arg, ret, monitor));
                    if let Err(val) = res {
                        ctx.set_ret0(val.into()); << arg0
                    } else {
                        ctx.set_ret0(rmi::SUCCESS); << arg0
                    }
                }
                None => {
                    error!("Not registered event: {:X}", ctx.cmd);
                    ctx.init_arg(&[rmi::RET_FAIL]); << like this
                }
```

PR #130, opened 2023-07-25
Fix smc context bug

bitboom:
Fixed https://github.com/Samsung/islet/issues/129

This PR fixes to encode `error code` to `x0` properly.

### DIFF
--- a/rmm/monitor/src/event/mainloop.rs
+++ b/rmm/monitor/src/event/mainloop.rs
@@ -70,11 +70,8 @@ impl Mainloop {
 
             match self.on_event.get(&ctx.cmd) {
                 Some(handler) => {
-                    let res = ctx.do_rmi(|arg, ret| handler(arg, ret, monitor));
-                    if let Err(val) = res {
-                        ctx.set_ret0(val.into());
-                    } else {
-                        ctx.set_ret0(rmi::SUCCESS);
+                    if let Err(code) = ctx.do_rmi(|arg, ret| handler(arg, ret, monitor)) {
+                        ctx.init_arg(&[code.into()]);
                     }
                 }
                 None => {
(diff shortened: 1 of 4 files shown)


jinbpark (approved):

rmm/monitor/src/event/mainloop.rs
```diff
@@ -70,11 +70,8 @@ impl Mainloop {
 
             match self.on_event.get(&ctx.cmd) {
                 Some(handler) => {
-                    let res = ctx.do_rmi(|arg, ret| handler(arg, ret, monitor));
-                    if let Err(val) = res {
-                        ctx.set_ret0(val.into());
-                    } else {
-                        ctx.set_ret0(rmi::SUCCESS);
+                    if let Err(code) = ctx.do_rmi(|arg, ret| handler(arg, ret, monitor)) {
+                        ctx.init_arg(&[code.into()]);
```
zpzigi754, line 74: From the point of semantic, argument should be input while return be output. It looks a bit weird to set the argument (input) at the end of the RMI calls, because it would be more natural to set the output. 

Maybe, we could use the previous return value to set the x0 in another place.

bitboom: I totally agree with you. I will make a refactor PR to improve semantic for this.

bitboom: > Maybe, we could use the previous return value to set the x0 in another place.

1. https://github.com/Samsung/islet/blob/main/rmm/monitor/src/event/mod.rs#L71 sets the ret0.
2. https://github.com/Samsung/islet/blob/main/rmm/monitor/src/event/mod.rs#L80 replace to x0 (arg0)

Issue #131, opened 2023-07-26
A safe interface to read/write page table 


jinbpark:
- Currently, we write a page table entry in a typical way, (e.g., `self.0 = RawPTE::new(0);`), but this is not a safe way to modify page table entries, as it requires a strict order of instructions.
   - e.g., secure way:  `str something to pte;  barrier_instructions;  tlb_instructions;`  it should be read/written atomically and proper barrier or tlb instructions must be followed right after.

Issue #112, opened 2023-07-13
Figure out a better way to associate Rec with Rd


jinbpark:
Currently,
```
pub struct Rec {
    pub rd: &'static Rd,
    vcpuid: usize,
    ripas: Ripas,
}
```

- One Rd can have multiple Recs. So, roughly thinking, we need to have the ability to manage the reference count of the rd. (need to look at RMM spec and TF-RMM, and figure out what would be the best way to associate Rec with Rd)

jinbpark:
More generally speaking, what this issue wants to solve is
- if X is in use, Y must not be freed.

For example, there are multiple RECs in use, the associated RD must not be destroyed even when RMI_REALM_DESTROY called. (error code: RMI_ERROR_IN_USE)

The cases in this category are:
- RD (REALM_DESTROY):  If there are other instances (e.g., Rec) that points to an RD, the RD should not be destroyed.
- REC (REC_DESTROY):  If there are other instances (e.g., Run) that points an REC, the REC should not be destroyed.

jinbpark:
More concrete scenario:
(1) Create Rd
(2) Create Rec
(3) Destroy Rd --> Check if this Rd's reference count is 1 by (2). 

Possible approaches to do:
(1)

Issue #113, opened 2023-07-13
Reinforce granule management


jinbpark:
- Currently, it uses the big lock for the granule table. We have to turn this into per-entry lock in accordance with RMM spec.
- Any other security checks additionally needed?? --> figure it out too.

jinbpark:
--- Work In Progress ---

jinbpark:
things that are worth to consider:
- double-fetch bug in TF-RMM? --> TF-RMM maps Rd/Rec/... and directly accesses those physical areas, where it could become vulnerable to double-fetch bugs, as it is similar to what's happening in kernel(linux)-user interfaces.

PR #132, opened 2023-07-26
initialize Rec.ripas by zero (fix fdt bug)

### LINKED ISSUE #118: Sporadic realm linux boot failure
## Symptom
No boot message from realm linux

## Related code
Linux realm got stuck while processing fdt in setup_arch():
```armasm
ffff800008c1387c <setup_arch>:
....
ffff800008c13924:   94000e7b    bl  ffff800008c17310 <fixmap_remap_fdt>
ffff800008c13928:   aa0003f3    mov x19, x0
ffff800008c1392c:   b50000e0    cbnz    x0, ffff800008c13948 <setup_arch+0xcc> 
// <=== result of fixmap_remap_fdt() is often not zero. 
ffff800008c13930:   aa1303e2    mov x2, x19
ffff800008c13934:   9101c3e1    add x1, sp, #0x70
ffff800008c13938:   d0fff820    adrp    x0, ffff800008b19000 <kallsyms_token_index+0x1250>
ffff800008c1393c:   911de000    add x0, x0, #0x778
ffff800008c13940:   97f42b11    bl  ffff80000891e584 <_printk>
ffff800008c13944:   14000009    b   ffff800008c13968 <setup_arch+0xec>
ffff800008c13948:   b9806fe1    ldrsw   x1, [sp, #108]
ffff800008c1394c:   f9403be0    ldr x0, [sp, #112]
ffff800008c13950:   97d8c0fc    bl  ffff800008243d40 <memblock_reserve>
ffff800008c13954:   aa1303e0    mov x0, x19
ffff800008c13958:   9400cbfc    bl  ffff800008c46948 <early_init_dt_scan>
ffff800008c1395c:   72001c1f    tst w0, #0xff
ffff800008c13960:   54000081    b.ne    ffff800008c13970 <setup_arch+0xf4>  // b.any
ffff800008c13964:   17fffff3    b   ffff800008c13930 <setup_arch+0xb4>
ffff800008c13968:   d503203f    yield    // Realm is stuck here
ffff800008c1396c:   17ffffff    b   ffff800008c13968 <setup_arch+0xec>
``` 

Corresponding linux source code:
```c
static void __init setup_machine_fdt(phys_addr_t dt_phys)
{
    int size;
    void *dt_virt = fixmap_remap_fdt(dt_phys, &size, PAGE_KERNEL);
    const char *name;

    if (dt_virt)
        memblock_reserve(dt_phys, size);

    if (!dt_virt || !early_init_dt_scan(dt_virt)) {  // failed
        pr_crit("\n"
            "Error: invalid device tree blob at physical address %pa (virtual address 0x%px)\n"
            "The dtb must be 8-byte aligned and must not exceed 2 MB in size\n"
            "\nPlease check your bootloader.",
            &dt_phys, dt_virt);

        /*
         * Note that in this _really_ early stage we cannot even BUG()
         * or oops, so the least terrible thing to do is cpu_relax(),
         * or else we could end-up printing non-initialized data, etc.
         */
        while (true)
            cpu_relax();  // <===== here we stuck at this point: corresponds to the 'yield' above
```

## Workaround
Improper cache operation or tlb invalidation seems to cause this. Configure FVP to have a single core.
```
diff --git a/scripts/.config b/scripts/.config
index 512365cf8956..fde2734733c4 100644
--- a/scripts/.config
+++ b/scripts/.config
@@ -14,7 +14,7 @@ bp.secure_memory=1
 bp.smsc_91c111.enabled=1
 bp.ve_sysregs.exit_on_shutdown=1
 cache_state_modelled=0
-cluster0.NUM_CORES=4
+cluster0.NUM_CORES=1
 cluster0.PA_SIZE=48
 cluster0.check_memory_attributes=0
 cluster0.cpu0.DCZID-log2-block-size=4
@@ -36,7 +36,7 @@ cluster0.restriction_on_speculative_execution=2
 cluster0.restriction_on_speculative_execution_aarch32=2
 cluster0.rme_support_level=2
 cluster0.stage12_tlb_size=1024
-cluster1.NUM_CORES=4
+cluster1.NUM_CORES=0
 cluster1.PA_SIZE=48
 cluster1.check_memory_attributes=0
 cluster1.cpu0.DCZID-log2-block-size=4


```

jinbpark:
- This PR fixes #118 .
- In-depth analysis for this bug is [here](https://github.com/Samsung/islet/issues/118#issuecomment-1650894030).
- While debugging this issue, I came across a new issue, #131 , which needs to be handled in the future.
- Plus, we need to have a secure way to manage Rd/Rec/Data/... . They could be handled via #112 .

### DIFF
--- a/rmm/monitor/src/rmi/rec/mod.rs
+++ b/rmm/monitor/src/rmi/rec/mod.rs
@@ -30,6 +30,7 @@ impl Rec {
         let rec: &mut Rec = &mut *(rec_addr as *mut Rec);
         rec.vcpuid = vcpuid;
         rec.rd = rd;
+        rec.set_ripas(0, 0, 0, 0);
         ManuallyDrop::new(rec)
     }
 


jinbpark:
Additional thought:  Not sure, but we might be able to catch this bug using Miri, as it is a kind of undefined behavior.

zpzigi754 (approved):

bokdeuk-jeong:
oops! my bad. Thanks for finding the bug! 

By the way, when the pages are delegated to RMM before it is used for Rec or Rd, the state of the page granule is UNDELEGATED. According to the ARM spec, the state transition of a granule to UNDELEGATED has to be accompanied with wiping out the page. As I know, we also have implemented it too. However, there seems a mistake with the granule implementation.

bokdeuk-jeong (approved):

jinbpark:
> accompanied with wiping out the page. As I know, we also have implemented it too. However, there seems a mistake with the granule implementation.

I got your point. I had a quick look at this.
At this moment, wiping out seems to only happen when making a transition from "Rd/Rec" to "Delegated" based on the below code. This seems to be a mistake as you said.
I'm doing #113 , I'll fix this there as #113 has to do with granule management.
```
pub fn set_granule(addr: usize, state: GranuleState, mm: PageMap) -> usize {
     GranuleState::Delegated => {
            if prev_state != GranuleState::Undelegated
                && prev_state != GranuleState::Delegated
                && prev_state != GranuleState::RTT
            {
                granule.get_mut(&addr).unwrap().zeroize();   // This is only part that calls "zeroize()".
}
```

PR #134, opened 2023-07-31
Update granule management

jinbpark:
## What this PR does (Issue:  #113 )

- Update generic PageTable and Entry to support various types of Entry, including GranuleStatusTable which is far from hardward-associated page tables.
    - `entries: [E; N],`  --> before: N=512 was only supported,  after:  different N (num_entries in a table) is supported.
    -  `impl<A: Address, L: Level, E: Entry, const N: usize> PageTableMethods<A, L, E, N>`
        -  before: `E: Entry + Copy`  --> Entry must implement `Copy` trait.
        -  after: `E: Entry`  --> Entry doesn't have to be `Copy` trait, which allows entry-level locking. e.g., `pub struct Entry(Spinlock<T>)`
    - define `pub enum Error` for a better error handling

- Introduce GranuleStatusTable which is based on entry-level locking, not a big lock.
    - `pub static mut GRANULE_STATUS_TABLE: Option<GranuleStatusTable<'static>> = None;` --> the root table doesn't use a big lock.
    -  instead, `pub struct Entry(Spinlock<Inner>);`  --> each granule entry has spinlock.
    -  The current implementation doesn't consider "unset_page" which eliminates a sub table or invalidates an entry. (it should be added if needed in the future)

- Test for GranuleStatusTable's entry-level locking
    - I haven't implemented additional test cases to test its entry-level locking mechanism.
    - Because, tf-a-tests already has a test case that changes GranuleStatusTable through multi CPUs. The result was,
        - GranuleStatusTable with a big lock:  Success
        - GranuleStatusTable with no lock: Fail
        - GranuleStatusTable with entry-level lock (this PR):  Success

- Safety and TODO
    - As for concurrency safety of GranuleStatusTable (or PageTable in general), I added some comments with "Safety" tag. (some with "Note" tag)
    - I left some TODO works with the prefix of "TODO"

- This PR is based on [this discussion](https://github.com/Samsung/islet/pull/121).

## Thought

I think we need to apply entry-level locking to RTT/RMM as well in the future, as some RSI calls affect the page table mapping. (currently, we don't update page tables when RTT_SET_RIPAS called, it is one of TODO items)


rmm/monitor/src/mm/page_table/mod.rs
```diff
@@ -196,38 +278,44 @@ where
...
+            });
+            if res.is_err() {
+                warn!("unset_page fail");
+            }
         }
     }
 }
 
-impl<A: Address, L: HasSubtable, E: Entry> PageTable<A, L, E>
+impl<A: Address, L: HasSubtable, E: Entry, const N: usize> PageTable<A, L, E, N>
```
bokdeuk-jeong, line 309: Deriving the number of entries from the size of a table and the size of a single entry seems more natural to me.
What if a developer provide the incorrect numbers to it by mistake?

rmm/armv9a/src/rmm/page_table/mod.rs
```diff
@@ -11,6 +11,7 @@ pub enum L0Table {}
 impl Level for L0Table {
     const THIS_LEVEL: usize = 0;
     const TABLE_SIZE: usize = PAGE_SIZE;
+    const TABLE_ALIGN: usize = PAGE_SIZE;
```
bokdeuk-jeong, line 14: We can't say L0Table is always aligned with a page size. L0Table can have consecutive multiple pages and its alignment is upto how many consecutive pages it consists of.

jinbpark: The final align value for RMM page table is `PAGE_SIZE * 2` which is the same as before.
++ The reason why I newly added `TABLE_ALIGN` is GranuleStatusTable doesn't need to be aligned with TABLE_SIZE. (previously, the alignment was determined by TABLE_SIZE)

rmm/monitor/src/mm/page_table/mod.rs
jinbpark: > Deriving the number of entries from the size of a table and the size of a single entry

I tried this way first but it didn't work out, as long as we keep using a static array `entries[E; N]`, because rust compiler emits compile errors.

In detail, what I tried was as follows.
```
struct Entries<E, const N: usize>([E; N]);

pub struct PageTable<A, L: Level, E>
where Entries<E, { L::NUM_ENTRIES }>: Sized  // --> this hint is required for const generic expression.
{
    entries: [E; L::NUM_ENTRIES],   // derive NUM_ENTRIES from trait "L".
    level: PhantomData<L>,
    address: PhantomData<A>,
   // in this case, "L::NUM_ENTRIES" --> this is called const generic expression, not const generic.
}
```

I almost made it, but ended up being failed, because of the below line. (I couldn't figure out how to fix it, not sure it's possible to fix it.)
```
impl<A: Address, L: HasSubtable, E: Entry> PageTable<A, L, E>
where
    Entries<E, { L::NUM_ENTRIES }>: Sized
{
fn subtable<S: PageSize>(&self, page: Page<S, A>) -> &mut PageTable<A, L::NextLevel, E> {
    unsafe { &mut *(subtable_addr.as_usize() as *mut PageTable<A, L::NextLevel, E>) }
}
// I think the reason is..
// -- the hint for PageTable is Entries<E, { L::NUM_ENTRIES }>, which targets current table.
// -- but it returns "mut PageTable<A, L::NextLevel, E>" which is the next table.
// -- imo, this mismatch was something that is impossible to solve.
}
```

> What if a developer provide the incorrect numbers to it by mistake?

An alternative way for this is to add `assert` into new*() functions. I'll do this.

jinbpark: ++ FYI:  one fun thing was:  when I tried this const_generic_expression stuff, I faced a rustc compiler bug (seg fault in rustc)
```
thread 'rustc' panicked at 'Box<dyn Any>', /rustc/b70baa4f922a1809d79caeaeb902800c3be283b9/compiler/rustc_errors/src/lib.rs:1576:9
```

rmm/armv9a/src/realm/mm/page_table/mod.rs
```diff
@@ -13,8 +13,12 @@ pub enum L0Table {}
 impl Level for L0Table {
     const THIS_LEVEL: usize = 0;
     const TABLE_SIZE: usize = PAGE_SIZE;
+    const TABLE_ALIGN: usize = PAGE_SIZE;
     const NUM_ENTRIES: usize = (Self::TABLE_SIZE / core::mem::size_of::<Entry>());
 }
+impl L0Table {
+    pub const NUM_ENTRIES: usize = (PAGE_SIZE / core::mem::size_of::<Entry>());
```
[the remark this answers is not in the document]
jinbpark: sorry, it was just an oversight. I'll remove it.

bokdeuk-jeong (approved):

bitboom (approved):

PR #135, opened 2023-08-02
acs: Pass granule delegation case

bitboom:
This PR covers negative TCs related RMI_GRANULE_DELEGATE.

## Changes
1. Change GranuleState to Delegated with PA translated by IPA given RMI 
2. Handle RMI_ERROR_INPUT according to RMM Spec
3. Set granule state only when un/delegation succeed

## Negative cases
| idx | type | addr |
|:---:|:---:|:---:|
|1| ADDR_UNALIGNED| 0x88300001|
|2| ADDR_DEV_MEM_MMIO| 0x1C0B0000|
|3|ADDR_OUTSIDE_OF_PERITTED_PA| 0x1000000001000|
|4| ADDR_DELEGATED| 0x8830C000|
|5| ADDR_REC| 0x88315000|
|6| ADDR_DATA| 0x88348000|
|7| ADDR_RTT| 0x88306000|
|8| ADDR_RD| 88303000|
|9| ADDR_PAS_REALM| 8834E000|
|10| ADDR_PAS_SECURE| 6000000|

### DIFF
--- a/rmm/monitor/src/rmi/gpt.rs
+++ b/rmm/monitor/src/rmi/gpt.rs
@@ -19,12 +19,11 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
 
 pub fn mark_realm(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error> {
     let cmd = smc.convert(smc::Code::MarkRealm);
-
-    if granule::set_granule(addr, GranuleState::Delegated) != granule::RET_SUCCESS {
+    if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
         return Err(Error::RmiErrorInput);
     }
 
-    if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
+    if granule::set_granule(addr, GranuleState::Delegated) != granule::RET_SUCCESS {
         return Err(Error::RmiErrorInput);
     }
 
@@ -33,13 +32,12 @@ pub fn mark_realm(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error>
 
 pub fn mark_ns(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error> {
     let cmd = smc.convert(smc::Code::MarkNonSecure);
+    if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
+        panic!("A delegated granule should only be undelegated on request from RMM.");
+    }
 
     if granule::set_granule(addr, GranuleState::Undelegated) != granule::RET_SUCCESS {
         return Err(Error::RmiErrorInput);
-    }
-
-    if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
-        panic!("A delegated granule should only be undelegated on request from RMM.");
     }
 
     Ok(())
--- a/rmm/monitor/src/rmi/gpt.rs
+++ b/rmm/monitor/src/rmi/gpt.rs
@@ -18,6 +18,10 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
 }
 
 pub fn mark_realm(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error> {
+    if !granule::validate_addr(addr) {
+        return Err(Error::RmiErrorInput);
+    }
+
     let cmd = smc.convert(smc::Code::MarkRealm);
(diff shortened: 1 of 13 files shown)


bitboom:
This PR should be merged after https://github.com/Samsung/islet/pull/134. 
The reason is that [this line](https://github.com/Samsung/islet/pull/135/files#diff-446c01718ff9fdfe067c3793b4ee94bafdda0f2e8bce376d07d506529819171e) makes panic according to granule state translation but current implement does not handle it properly.

rmm/monitor/src/rmi/gpt.rs
```diff
@@ -9,57 +9,45 @@ use crate::smc;
...
         return Err(Error::RmiErrorInput);
-    } else {
-        let smc_ret = smc.call(cmd, &[addr]);
-        // XXX: Should we use ret[0] for smc return value?
-        //      For RMI calls, ret[0] is mostly used for RMI's result value
-        ret[0] = smc_ret[0];
     }
+
+    if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
+        panic!("A delegated granule should only be undelegated on request from RMM.");
```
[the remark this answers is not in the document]
bitboom: Thank you for suggestion.

Actually i referenced [tf-rmm](https://github.com/TF-RMM/tf-rmm/blob/main/runtime/rmi/granule.c#L48-L54) which means the granule of `DELEGATED state` should be always undelegated by RMM successfully.

So, we have to check the addr is `DELEGATED state` before calling smc. I let this work for next PR (undelegation TCs).

If you still want to change this message, feel free to let me know :smile:  I will reflect.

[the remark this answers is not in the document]
zpzigi754: Thank you for the reference! `smc_granule_undelegate (Granule Undelegate)`  itself is initiated by the Host, so I thought that the granule undelegation is the request from the Host. Leaving it as it is would be totally fine :)

zpzigi754 (approved):

rmm/monitor/src/rmi/gpt.rs
```diff
@@ -4,62 +4,52 @@ use crate::rmi;
...
 }
 
-pub fn mark_realm(
-    smc: smc::SecureMonitorCall,
-    _mm: PageMap,
-    addr: usize,
-    ret: &mut [usize],
-) -> Result<(), Error> {
+pub fn mark_realm(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error> {
+    if !granule::validate_addr(addr) {
```
bokdeuk-jeong, line 21: What if the state of a granule is in a state other than undelegated? We need to check it ahead of this smc call.

rmm/monitor/src/rmi/gpt.rs
```diff
@@ -4,62 +4,52 @@ use crate::rmi;
...
-    addr: usize,
-    ret: &mut [usize],
-) -> Result<(), Error> {
+pub fn mark_ns(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error> {
+    if !granule::validate_addr(addr) {
+        return Err(Error::RmiErrorInput);
+    }
+
     let cmd = smc.convert(smc::Code::MarkNonSecure);
+    if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
```
bokdeuk-jeong, line 43: Similar comment above goes here too. We make sure current state is undelegated before we actually mark_ns.

rmm/monitor/src/rmi/gpt.rs
bitboom: Yes. You're right. We have to validate the granule state and we need the way what gets the state by given PA.

I recently raised the issue https://github.com/Samsung/islet/issues/138 to handle it.

Could I fix it at following PR? (Since it needs more flexible locking related with PA and Entry.)

rmm/monitor/src/rmi/gpt.rs
bitboom: I agree with you. I replied it on above comment.

rmm/monitor/src/rmi/gpt.rs
jinbpark: I'm doing "https://github.com/Samsung/islet/issues/112", which aims to add "reference counting" into granule. I'm considering `get_granule_state()` in a concurrency safe way there. I'll add this API in a PR that solves #112 .

jinbpark (approved):

bokdeuk-jeong (approved):

PR #142, opened 2023-08-07
add get_granule_state()

jinbpark:
This is an interim implementation of `get_granule_state()` for ACS works that @bitboom is working on.
This implementation will likely be modified in a future PR that deals with #112.
See `test_get_granule_state()` to know how to use i.t


bitboom (approved):
Thanks for the quick interim implementation 👍

PR #143, opened 2023-08-09
Support granule refcount and graunule transaction

jinbpark:
First of all, sorry for a huge PR.. but I thought it would be better to put them in all at once.

#### What it means and examples

"granule transaction" means typical stuff that requires "transaction management".
Take the example of "REALM_CREATE".
```
listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| { 
    let mut granule = get_granule_if!(arg[0], GranuleState::Delegated);  // (1) get_granule
    ....  // (2) do something with granule
    set_granule(&mut granule, GranuleState::RD)   // (3) set_granule at the end of this handler when everything is ok.
}
```
We can say that (1)~(3) is a granule transaction to access a granule in a concurrency-safe manner.

1. `let mut granule = get_granule_if!(addr, state)`: 
      Get a granule if the state of it equals to the given state.
2. `set_granule(&mut granule, GranuleState::RD)`:
      At the end of a function (in more general, a point where you want to terminate a transaction),
      you can set the state of the granule.

The reason why it is needed is Granule must be accessed in a unit of transaction.
For example, we can say that the below code is problematic when some error comes out.
```
// current implementation
listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| {
    set_granule(arg[0], GranuleState::RD);  // (1) set_granule first.
    ...
       return Err(..);  // (2) error comes out, then REALM_CREATE fails but the granule is still GranuleState::RD.
                                // this is semantically wrong.
    ...
    Ok(())
}
```

Note for @bitboom :  `mark_realm()` and `mark_ns()` have been rewritten considering this granule transaction.

#### Secure API designs

Take the example of REALM_CREATE again.
```
listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| { 
    let mut granule = get_granule_if!(arg[0], GranuleState::Delegated);  // (1) get_granule
     let rd = granule.content_mut::<Rd>();     // (2) get a right to access content 
     rd.init(id);   // (3) write something to the physical address of RD securely.
    set_granule(&mut granule, GranuleState::RD)   // (4) set_granule at the end of this handler when everything is ok.
}
```
- (1) `let mut granule` represents `EntryGuard`, which means this API tries to acquire a granule lock. And if successful,
   it can have a guard to access a granule behind the lock. The lock is of course automatically unlocked.
- Every operation can only be done through **this guard**. So, we can ensure that read/write from/to a granule is done under a proper lock.
- Also, (2)/(3):  "content" (placed at the address of a granule) can only be read/written under the lock. As this is a way to securely access Rd/Rec/..., `ManuallyDrop` is no longer used for Rd/Rec.
- See `granule/mod.rs` Line_64 for more detail about this design.

## Feature-2:  refcount

At this point, the only point where refcount matters lies in REC_CREATE.
```
listen!(mainloop, rmi::REC_CREATE, |arg, ret, rmm| {
    let rd_granule = get_granule_if!(arg[1], GranuleState::RD);
    let mut rec_granule = get_granule_if!(arg[0], GranuleState::Delegated);
    ....
    set_granule_with_parent(rd_granule.clone(), &mut rec_granule, GranuleState::Rec)?;  // --> rd is parent and rec is child.
    Ok(())
}
```

If REALM_DESTROY is called when there is one or more than two RECs,
it returns `Err(Error::MmIsInUse)`. The code to check this is in `granule/entry.rs` Line_144.

Note that I haven't tested calling "REALM_DESTROY" when REC is still there.
We need a system test case to do this.

## TODO

- Refactor "PageTable" (more precisely, move "array-based entries" from a core interface to each impl)
- Improve the API design of granule transaction considering "RMM map". 
- Add a document page about what this PR has done.


rmm/monitor/src/mm/page_table/mod.rs
```diff
@@ -58,6 +63,15 @@ pub trait Entry {
         }
     }
     fn index<L: Level>(addr: usize) -> usize;
+    fn map<F: FnOnce(usize) -> Result<(), Error>>(
+        &mut self,
+        _index: usize,
```
hihi-wang, line 68: Why `map` to single `Entry` needs index? I guess it's safe to remove it.

jinbpark: Thanks for pointing it out. I think it was an oversight.  There is no point index needs to be used as you said.

jinbpark: Oh sorry,, "index" is needed in `granule/entry.rs` but it could get eliminated after the mentioned refactoring done (Refactor `PageTable` interfaces).  I'll do it there.

rmm/monitor/src/rmm/granule/entry.rs
```diff
@@ -15,16 +19,25 @@ use spinning_top::Spinlock;
...
-    fn set_state(&mut self, addr: usize, state: u64) -> Result<(), Error> {
+    fn set_state<F>(&mut self, addr: usize, state: u64, destroy_callback: F) -> Result<(), Error>
+    where
+        F: Fn() -> Result<(), Error>,
+    {
         let prev = self.state;
 
         if prev == GranuleState::Undelegated {
-            if state != GranuleState::Delegated {
+            if state != GranuleState::Delegated && state != GranuleState::Undelegated {
```
bitboom, line 40: How about change this whole if-else statement to match?

```suggestion
let valid = match prev {
    GranuleState::Delegated => state != GranuleState::Delegated,
    _ => state == GranuleState::Delegated,
};

if !valid {
    return Err(Error::MmStateError);
}
```

rmm/monitor/src/rmm/granule/mod.rs
```diff
@@ -59,33 +61,124 @@ impl GranuleState {
...
+                guard
+            } else {
+                return to_rmi_result(Err(MmError::MmErrorOthers));
+            };
+
+            let guard = match guard_res {
+                Some(g) => g,
+                _ => return to_rmi_result(Err(MmError::MmErrorOthers)),
+            };
+            guard
```
bitboom, line 124: Could we return Result<Guard, MMError> on `get_granule` and `get_granule_if`?
If it is changed, we can use like `let granule = get_granule!()?;`

```suggestion
            if !validate_addr($addr) {
                    Err(MmError::MmInvalidAddr);
            } else {
                 if let Some(gst) = unsafe { &mut GRANULE_STATUS_TABLE } {
                     let pa = Page::<GranuleSize, PhysAddr>::including_address(PhysAddr::from($addr));
                     match gst.root_pgtlb.entry(pa, |e| page_table::Entry::lock(e)).map(|g|  .. Ok(g) )
            } else {
                Err(MmError::MmErrorOthers))
            }
```

jinbpark: `let granule = get_granule!()?;`  --> I really love it... many thanks. I'll do it.

rmm/monitor/src/rmm/granule/entry.rs
jinbpark: Reflected:

rmm/monitor/src/rmm/granule/mod.rs
jinbpark: Reflected:

jinbpark:
NOTICE:  I found that realm-linux booting goes something wrong with this PR. I'll get it back to Draft and look into this issue and back to ready for review again once it gets done.

rmm/monitor/src/rmm/granule/entry.rs
```diff
@@ -15,40 +19,63 @@ use spinning_top::Spinlock;
...
     addr: usize,
+    /// parent that this granule points to
+    /// the only case at this point is "Rd(parent) - Rec(child)"
+    /// Notice: do not put self-reference into this field, which may cause undefined behaviors.
+    parent: Option<Inner>,
 }
 
 impl Granule {
-    fn set_state(&mut self, addr: usize, state: u64) -> Result<(), Error> {
+    fn set_state<F>(&mut self, addr: usize, state: u64, destroy_callback: F) -> Result<(), Error>
```
bokdeuk-jeong, line 33: Sorry for not following though the code for my laziness. 
Can we make sure that the page is  zeroized with current implementation whenever a granule goes into the UNDELEGATED state?
That includes when ISLET rmm receives a new page from NW with the DELEGATE RMI and the state transition from other states to UNDLETATED.

rmm/monitor/src/mm/page_table/mod.rs
```diff
@@ -322,7 +337,25 @@ where
...
-        unsafe { &mut *(subtable_addr.as_usize() as *mut PageTable<A, L::NextLevel, E, N>) }
+        match self.entries[index].subtable(index, L::THIS_LEVEL) {
+            Ok(table_addr) => unsafe {
+                &mut *(table_addr as *mut PageTable<A, L::NextLevel, E, N>)
+            },
+            Err(_) => panic!(),
+        }
+    }
+
+    fn subtable_and_set_page<S: PageSize>(
```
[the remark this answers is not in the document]
jinbpark: Not necessary from functional perspective.
It's just a helper function that does two things at once: get a table address and set_page.
++ it returns `Result<(), Error>` that match with the return type of caller functions. This is another reason.

rmm/monitor/src/rmi/gpt.rs
```diff
@@ -18,21 +18,20 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
 
-    granule::set_granule(addr, GranuleState::Delegated)?;
+    set_granule(&mut granule, GranuleState::Delegated)?;
     Ok(())
 }
 
 pub fn mark_ns(smc: smc::SecureMonitorCall, addr: usize) -> Result<(), Error> {
-    if !granule::validate_addr(addr) {
+    let mut granule = get_granule!(addr)?;
+    if granule.state() != GranuleState::Delegated && granule.state() != GranuleState::Undelegated {
```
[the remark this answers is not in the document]
jinbpark: `get_granule_if()` currently only supports checking with one state while this checks with two states. We can implement this using closure for a more general implementation.
(FYI:  I found that `mark_ns()` is called even when it is already in the state of Undelegated, in tf-a-tests)

rmm/monitor/src/rmm/granule/entry.rs
jinbpark: > whenever a granule goes into the UNDELEGATED state?

It wont' wipe out it in this case. I'll update it to handle this case.

rmm/monitor/src/mm/guard.rs
```diff
@@ -0,0 +1,49 @@
+use core::ops::{Deref, DerefMut};
+use spinning_top::SpinlockGuard;
+
+/// EntryGuard provides a secure interface to access Entry while holding the corresponding lock.
+/// Also, it is used as a means of accessing "content" placed at the address of Entry under the lock.
```
zpzigi754, line 5: [minor typo]
a means->a mean

bitboom (approved):

rmm/monitor/src/mm/page_table/mod.rs
```diff
@@ -322,7 +337,25 @@ where
         assert!(L::THIS_LEVEL < S::MAP_TABLE_LEVEL);
 
         let index = E::index::<L>(page.address().into());
-        let subtable_addr = self.entries[index].address(L::THIS_LEVEL).unwrap();
-        unsafe { &mut *(subtable_addr.as_usize() as *mut PageTable<A, L::NextLevel, E, N>) }
+        match self.entries[index].subtable(index, L::THIS_LEVEL) {
+            Ok(table_addr) => unsafe {
+                &mut *(table_addr as *mut PageTable<A, L::NextLevel, E, N>)
+            },
+            Err(_) => panic!(),
```
zpzigi754, line 344: How about returning the error value or `None` to the caller instead of panicking by modifying the return type?

rmm/monitor/src/rmm/granule/entry.rs
jinbpark: Reflected:

rmm/monitor/src/mm/page_table/mod.rs
```diff
@@ -58,6 +63,15 @@ pub trait Entry {
         }
     }
     fn index<L: Level>(addr: usize) -> usize;
+    fn map<F: FnOnce(usize) -> Result<(), Error>>(
+        &mut self,
+        _index: usize,
+        level: usize,
+        func: F,
+    ) -> Result<(), Error> {
+        let subtable_addr = self.address(level).unwrap();
```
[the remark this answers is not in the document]
jinbpark: Reflected: but I couldn't use `ok_or` as the type of Ok does not equal to the type of Option.
Instead, I just used `match` to eliminate `unwrap()` as you said.

rmm/monitor/src/mm/guard.rs
jinbpark: means --> this is a singular word.

zpzigi754: Oh, my fault. Thanks!

rmm/monitor/src/mm/page_table/mod.rs
jinbpark: Reflected:

rmm/monitor/src/mm/guard.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl<'a, E> DerefMut for EntryGuard<'a, E> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut *self.inner
+    }
+}
+
+pub trait Content {
+    const FLAGS: u64;
+}
```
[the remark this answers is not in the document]
jinbpark: > Maybe we could change the name of this field.

Agreed. I'm planning to do this in the follow-up PR (refactoring the core PageTable interface).
That confusion mainly comes from that GranuleTable is trying to fit in the current PageTable interface.
I'll handle this in that PR.

> Content's FLAGS is only intended to contain GranuleState

I designed this to be  common to all PageTable implementations.
To explain it better, let me give you one example.
Actually, the reason why I used this was for a validation check.  (but this check is not performed as of now)
```
// use
let rd = granule.content_mut::<Rd>();   // we want to check if this granule is Rd.

// check
pub fn content_mut<T: Content>(&mut self) -> &mut T {
    if T::FLAGS != self.flags {
         // T::FLAGS == Rd 
         // self.flags --> the state of the granule at run-time.
         error;  // 
    }
}
```

We can do this check for another kind of PageTable. (e.g., HW Table)
```
// check
// For HW page table, "Content" represents a data page that looks like `array: [u8; 4096]`
pub fn content_mut<T: Content>(&mut self) -> &mut T {
    if T::FLAGS != self.flags {
        // T::FLAGS == "read-only"
        // self.flags == "read-write"
        error;  // --> we can do this kind of validation check.
    }
}
```

zpzigi754 (approved):

rmm/monitor/src/mm/guard.rs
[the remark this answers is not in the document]
zpzigi754: Thank you for the detailed explanation! Now I got your intention.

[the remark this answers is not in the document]
jinbpark: I added a note (comment in the code) for what I said in this thread. (how flags can be used here)

PR #145, opened 2023-08-10
Remove unsafe & ManuallyDrop about shared data structrues 

bitboom:
This PR
- removes `unsafe & ManuallyDrop`
- adds tests for spec compliance

`Unsafe & ManuallyDrop` was introduced by union type. (the way of tf-rmm implement) 
 I change this by adding padding.

---

I will apply this way to other structures continuously

### DIFF
--- a/rmm/monitor/src/rmi/realm/params.rs
+++ b/rmm/monitor/src/rmi/realm/params.rs
@@ -1,4 +1,5 @@
 use crate::host::Accessor as HostAccessor;
+use crate::rmm::granule::GRANULE_SIZE;
 
 #[repr(C)]
 #[derive(Debug)]
@@ -36,25 +37,17 @@ impl Default for Params {
     }
 }
 
+const_assert_eq!(core::mem::size_of::<Params>(), GRANULE_SIZE);
+
 impl HostAccessor for Params {}
 
 #[cfg(test)]
 pub mod test {
     use super::*;
 
-    macro_rules! offset_of {
-        ($type:ty, $field:tt) => {{
-            let dummy = core::mem::MaybeUninit::<$type>::uninit();
-            let dummy_ptr = dummy.as_ptr();
-            let member_ptr = unsafe { ::core::ptr::addr_of!((*dummy_ptr).$field) };
-
-            member_ptr as usize - dummy_ptr as usize
-        }};
-    }
-
     #[test]
     fn spec_params() {
-        assert_eq!(core::mem::size_of::<Params>(), 4096);
+        assert_eq!(core::mem::size_of::<Params>(), GRANULE_SIZE);
 
         assert_eq!(offset_of!(Params, features_0), 0x0);
         assert_eq!(offset_of!(Params, hash_algo), 0x100);
--- a/rmm/monitor/src/rmi/realm/params.rs
+++ b/rmm/monitor/src/rmi/realm/params.rs
@@ -2,7 +2,6 @@ use crate::host::Accessor as HostAccessor;
 use crate::rmm::granule::GRANULE_SIZE;
 
 #[repr(C)]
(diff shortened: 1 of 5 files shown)


jinbpark (approved):

jinbpark:
Does it have something to do with ACS tests? (it seems not I guess)

rmm/monitor/src/rmi/rec/params.rs
```diff
@@ -1,101 +1,68 @@
...
-struct _Inner {
-    num_aux: u64,
-    aux: [u64; 16],
-}
-
-#[repr(C)]
-union Inner {
-    val: core::mem::ManuallyDrop<_Inner>,
-    reserved: [u8; 0x1000 - 0x800],
+        assert_eq!(offset_of!(Params, flags), 0x0);
```
jinbpark, line 61: I like this idea. These are not checked at compile-time but are ensured by CI test. I think this pattern is useful for a wide range of places.

rmm/monitor/src/rmi/realm/params.rs
```diff
@@ -1,87 +1,73 @@
...
+use crate::rmm::granule::GRANULE_SIZE;
 
 #[repr(C)]
 pub struct Params {
-    features0: Features0,
-    hash_algo: HashAlgo,
-    rpv: RPV,
-    inner: Inner,
+    pub features_0: u64,
+    padding0: [u8; 248],
```
HeeillWang, line 7: How about define those padding size as a `constant` as it's repeated on `drop()`

bitboom:
> Does it have something to do with ACS tests? (it seems not I guess)

It is not related with ACS directly.

I am currently debugging below issue. 
The first trial was checking structure offset.
It was not the problem but i thought that it needs refactoring and testing.

---

## Current issue
I found that RMM reads dummy values (0xCFDFDFDF) with ACS tests.
- Read failed with ACS: Params 0x88306000
- Read succeed with TFTF: Params 0x88C01000

rmm/monitor/src/rmi/realm/params.rs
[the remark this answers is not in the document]
zpzigi754: Maybe we can also define a macro (`SET_MEMBER_RMI`) with union like `tf-rmm` to maintain the padding with more straightforward hex value (e.g., `SET_MEMBER_RMI(unsigned char hash_algo, 0x100, 0x400)`)?

rmm/monitor/src/rmi/rec/params.rs
zpzigi754: Later, we might be able to use the compile-time checks too (like `COMPILER_ASSERT(_condition)` in tf-rmm)!

zpzigi754 (approved):

rmm/monitor/src/rmi/rec/params.rs
bitboom: @zpzigi754 Yes right. This PR only applied compile-time check to `struct size`(`const_assert_eq!(core::mem::size_of::<Params>(), GRANULE_SIZE);`).

`offset_of` was hard to make at compile-time, because of pointer operation :(
I hope we will make this too at compile-time :)

rmm/monitor/src/rmi/realm/params.rs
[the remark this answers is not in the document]
bitboom: @hihi-wang Reflected :)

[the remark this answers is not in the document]
bitboom: @zpzigi754 Nice suggestion! I don't want to use union type, because it makes code dirty (ManuallyDrop & unsafe). 

So. I tried to make it as macro like what you said. But It was hard to make it :(

Issue #146, opened 2023-08-11
[Cache] Garbage value in shared region (NS, Realm)


bitboom:
I found that RMM reads dummy values (0xCFDFDFDF) with ACS tests.
- Read failed with ACS: Params 0x88306000
- Read succeed with TFTF: Params 0x88C01000

_Originally posted by @bitboom in https://github.com/Samsung/islet/issues/145#issuecomment-1672644130_

jinbpark:
@bitboom ,
Not sure if [this PR](https://github.com/Samsung/islet/pull/143) helps solving this,
But I added some points to wipe out a page by zero. 
1. a transition from Undelegated to Delegated
2. a transition from Delegated to Undelegated

I think it's worth a try running the same ACS test with this PR. (not that high probability though)

bitboom:
@jinbpark Thank you for the suggestion.

I found the cause of this issue.
The issue is that ACS set the shared region as `Normal, Inner/Outer WB/WA/RA`, but islet map the region as `Device` in `set_pages_for_rmi`.

When i changed ACS to `Device` temporarily, i could read the memory.

But I think our islet should change to `Normal` in `set_pages_for_rmi`. I tried it but failed :(

jinbpark:
> When i changed ACS to Device temporarily, i could read the memory.

This sounds quite likely a TLB issue. (not 100% sure though)
It might be worth trying to run the below code.

```
pub fn set_pages_for_rmi(addr: usize, secure: bool) {
    let device_flags = helper::bits_in_reg(PTDesc::INDX, attr::mair_idx::RMM_MEM);
    ....
    page_table.set_pages(va, phys, PAGE_SIZE, rw_flags | device_flags | secure_flags);
    ....
    unsafe {
        asm!("tlbi alle2", "dsb sy", "isb",);   // TLB flush
    }
}
```

bitboom:
@jinbpark Thanks for the suggestion! I tried it but i does't work :(

The problem seems to relate with cache. When i turn off cache with `cache_state_modelled=0` of acs's script, it work.
Since this is not the tfundamental solution, I detached this issue from origin PR https://github.com/Samsung/islet/pull/148.

Issue #147, opened 2023-08-14
Refactor PageTable interfaces


jinbpark:
- Move "entires: [E; N]" from core PageTable interfaces to each implementations.

jinbpark:
-- WORK IN PROGRESS --

PR #148, opened 2023-08-17
acs: Pass granule undelegation cases

bitboom:
This PR covers negative TCs related RMI_GRANULE_UNDELEGATE.
I have recorded the ACS status to https://github.com/Samsung/islet/issues/149

## Changes
1. Manage RTT pages
2. Fix mem attributes of shared pages for RMI -> detach cache issue https://github.com/Samsung/islet/issues/146

## Negative cases
| idx | type | addr |
|:---:|:---:|:---:|
|1| ADDR_UNALIGNED| 0x88300001|
|2| ADDR_DEV_MEM_MMIO| 0x1C0B0000|
|3|ADDR_OUTSIDE_OF_PERITTED_PA| 0x1000000001000|
|4| ADDR_UNDELEGATED| 0x8830C000|
|5| ADDR_REC| 0x88315000|
|6| ADDR_DATA| 0x88348000|
|7| ADDR_RTT| 0x88306000|
|8| ADDR_RD| 88303000|

### DIFF
--- a/rmm/monitor/src/event/mainloop.rs
+++ b/rmm/monitor/src/event/mainloop.rs
@@ -55,7 +55,10 @@ impl Mainloop {
                 ctx.resize_ret(ret_num);
                 self.queue.lock().push_back(ctx);
             },
-            || {},
+            || {
+                let ctx = Context::new(rmi::NOT_SUPPORTED_YET);
+                self.queue.lock().push_back(ctx);
+            },
         );
     }
 
--- a/rmm/monitor/src/rmi/gpt.rs
+++ b/rmm/monitor/src/rmi/gpt.rs
@@ -1,11 +1,13 @@
 use crate::event::Mainloop;
 use crate::listen;
+use crate::mm::error::Error as MmError;
 use crate::rmi;
 use crate::rmi::error::Error;
 use crate::rmm::granule::{set_granule, GranuleState};
 use crate::rmm::PageMap;
 use crate::smc;
-use crate::{get_granule, get_granule_if, get_granule_not_if, set_state_and_get_granule};
+use crate::{get_granule, get_granule_if, set_state_and_get_granule};
+
 extern crate alloc;
 
 pub fn set_event_handler(mainloop: &mut Mainloop) {
@@ -18,17 +20,11 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     });
 }
 
-fn check_state_not_if(addr: usize, state: u64) -> Result<(), Error> {
-    get_granule_not_if!(addr, state)?;
-    Ok(())
-}
-
-pub fn mark_realm(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
-    if check_state_not_if(addr, GranuleState::Undelegated).is_ok() {
-        // returns error if this is not in the state of Undelegated.
-        return Err(Error::RmiErrorInput);
-    }
-    let mut granule = set_state_and_get_granule!(addr, GranuleState::Undelegated)?;
+fn mark_realm(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
+    let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
+        Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated),
+        other => other,
+    }?;
 
     let cmd = smc.convert(smc::Code::MarkRealm);
     if smc.call(cmd, &[addr])[0] != smc::SMC_SUCCESS {
@@ -44,7 +40,7 @@ pub fn mark_realm(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Resu
     Ok(())
 }
 
-pub fn mark_ns(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
+fn mark_ns(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
     let mut granule = get_granule_if!(addr, GranuleState::Delegated)?;
 
     let cmd = smc.convert(smc::Code::MarkNonSecure);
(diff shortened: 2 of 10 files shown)


rmm/monitor/src/rmi/gpt.rs
```diff
@@ -50,9 +45,6 @@ pub fn mark_realm(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Resu
 }
 
 pub fn mark_ns(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
-    if check_state_if(addr, GranuleState::Undelegated).is_ok() {
-        return Ok(());
-    }
```
bokdeuk-jeong, line 55: I have a question that might not be related to this commit.
Why we call both check_state_not_if() and set_state_and_granule!() in mark_realm() while we only call get_granule_if!() in mark_ns()? 
Can't we just have a single call like get_granule_if!(addr, GranuleState::None) in mark_realm()?

bitboom: I refactor it. Could you check this https://github.com/Samsung/islet/pull/148/commits/0e99cd9b4c03725fbdd7a5f4b96319fbe81db626?

rmm/monitor/src/rmi/gpt.rs
```diff
@@ -18,22 +20,11 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
-
-pub fn mark_realm(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
-    if check_state_not_if(addr, GranuleState::Undelegated).is_ok() {
-        // returns error if this is not in the state of Undelegated.
-        return Err(Error::RmiErrorInput);
-    }
-    let mut granule = set_state_and_get_granule!(addr, GranuleState::Undelegated)?;
+fn mark_realm(smc: smc::SecureMonitorCall, mm: PageMap, addr: usize) -> Result<(), Error> {
+    let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
+        Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated),
```
jinbpark, line 25: FYI: `Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated),`
This is one of unnecessary inefficiency, and I expect this can be eliminated by getting #147 done. (refactor pagetable interfaces)

rmm/monitor/src/event/mainloop.rs
```diff
@@ -55,7 +55,10 @@ impl Mainloop {
                 ctx.resize_ret(ret_num);
                 self.queue.lock().push_back(ctx);
             },
-            || {},
+            || {
+                let ctx = Context::new(rmi::NOT_SUPPORTED_YET);
+                self.queue.lock().push_back(ctx);
+            },
```
zpzigi754, line 61: Isn't it better to call an explicit panic when unsupported commands have been reached? (as we are easy to miss important log messages).

zpzigi754 (approved):

bokdeuk-jeong (approved):

rmm/monitor/src/rmi/gpt.rs
jinbpark: Just doing `let mut granule = get_granule_if!(addr, GranuleState::Delegated)?;` in `mark_ns()` works for both realm-linux booting and tf-a-tests?

The reason why I did `if check_state_if(addr, GranuleState::Undelegated).is_ok() {` was, there were some calls to `mark_ns` when it was already in the state of Undelegated. (I don't remember it was realm-linux booting or tf-a-tests)  And when I returned a error code in that case, the system falled in a failure case.

rmm/monitor/src/rmi/realm/rd.rs
```diff
@@ -1,15 +1,18 @@
 use crate::mm::guard::Content;
 use crate::rmm::granule::GranuleState;
 
+// TODO: Integrate with our `struct Realm`
```
jinbpark, line 4: Agree with this TODO. Rd and Realm have some redundancy.

rmm/monitor/src/event/mainloop.rs
bitboom: If we panic unsupported commands like `RMI_RTT_FOLD`, it is hard to progress one TC fully even it is not related directly. And we returns errors to host when it comes to come unsupported commands, i think it's better for now.

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -36,17 +36,18 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+            rd.init(id, params.rtt_base as usize);
+            ret[1] = id;
+        })?;
 
         // set Rd state only when everything goes well.
         set_granule(&mut granule, GranuleState::RD)?;
+
+        let rd = granule.content::<Rd>();
+        let mut granule = get_granule_if!(rd.rtt_base(), GranuleState::Delegated)?;
+        set_granule(&mut granule, GranuleState::RTT)?;
```
jinbpark, line 49: Is there any reason we have to do "set_granule for RTT" after "set_granule for RD"?
I think doing "set_granule for RTT" first and then "set_granule for RD" seems more correct one.

rmm/monitor/src/rmi/gpt.rs
bitboom: I checked `realm-linux` is same with https://github.com/Samsung/islet/issues/106#issuecomment-1641917963 and `tf-a-tests` worked well.
I guess that [this line](https://github.com/Samsung/islet/pull/148/files#diff-a8f89f6017f8009dee1d13b922981665d13e030606ee35210e77993b24ce58beR131) fixed it. Previous code returned  error early by `?` in macro.

rmm/monitor/src/rmi/realm/mod.rs
bitboom: I agreed and reflected it :)

jinbpark (approved):

hihi-wang:
nice job :)

PR #152, opened 2023-08-23
Validate PA before copy from host to realm

bitboom:
This PR is intermediate status of `realm creation`. 
I share this PR to disable cache state modelling temporarily.

### DIFF
--- a/rmm/monitor/src/host/mod.rs
+++ b/rmm/monitor/src/host/mod.rs
@@ -2,8 +2,8 @@
 pub mod pointer;
 
 use crate::mm::guard::Content;
-use crate::rmm::granule::GranuleState;
-use crate::rmm::granule::GRANULE_SIZE;
+use crate::rmm::granule::validate_addr;
+use crate::rmm::granule::{GranuleState, GRANULE_SIZE};
 use crate::rmm::PageMap;
 
 /// This trait is used to enforce security checks for physical region allocated by the host.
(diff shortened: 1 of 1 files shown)


jinbpark (approved):

scripts/build-tf-rmm.sh
```diff
@@ -6,7 +6,7 @@ ROOT=$(git rev-parse --show-toplevel)
 
 cd $ROOT/third-party/tf-rmm
 export CROSS_COMPILE=$1
-cmake -DRMM_CONFIG=fvp_defcfg -S . -B build
+cmake -DRMM_CONFIG=fvp_defcfg -DLOG_LEVEL=50 -S . -B build
```
zpzigi754, line 9: Is this log level the maximum value to choose? I didn't try with this log level and wonder whether this log level includes the log about realm entrance, as I personally prefer not leaving logs about realm entrance which bloat the log.

rmm/monitor/src/host/mod.rs
```diff
@@ -12,6 +12,9 @@ pub trait Accessor {
     /// Try to do page-relevant stuff (e.g., RMM map).
     /// returns true only if everything goes well.
     fn acquire(ptr: usize, page_map: PageMap) -> bool {
+        if !validate_addr(ptr) {
```
hihi-wang, line 15: I have no context understanding here, but is it impossible to wrap this value with `Option` or use [pointer::is_null](https://doc.rust-lang.org/std/primitive.pointer.html#method.is_null), instead of passing `usize`?

scripts/build-tf-rmm.sh
bitboom: I want to print `memory map` but there are additional logs about measurement. So i reverted it as your suggestion.

zpzigi754 (approved):

PR #153, opened 2023-08-23
dcache flush to fix the problem that reads dummy value

### LINKED ISSUE #146: [Cache] Garbage value in shared region (NS, Realm)

jinbpark:
This PR can fix #146 by dcache-flush

### DIFF
--- a/rmm/armv9a/src/helper/asm.rs
+++ b/rmm/armv9a/src/helper/asm.rs
@@ -27,3 +27,17 @@ pub fn hvc<const IMM: u16>() {
         asm!("hvc #{}", const IMM);
     }
 }
+
+#[inline(always)]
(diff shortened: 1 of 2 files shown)


bitboom (approved):
Really nice !! 👍

I still wonder why tf-rmm works without dcache flush. I think i missed some points..
https://github.com/Samsung/islet-asset/blob/3rd-tf-rmm/lib/xlat/src/xlat_tables_utils.c#L413

zpzigi754 (approved):
Great!

PR #155, opened 2023-08-24
acs: Pass realm creation cases 

bitboom:
This PR covers negative TCs related RMI_REALM_CREATE.

## Changes
- Add validation logic of realm_params and features_reg_0
- Prevent double locking => Raise https://github.com/Samsung/islet/issues/162
- Manage vmid

## Negative cases
| idx | type |
|:---:|:---:|
|1| PARAMS_UNALIGNED |
|2| PARAMS_DEV_MEM |
|3|PARAMS_OUTSIDE_OF_PERMITTED_PA |
|4| PARAMS_PAS_REALM |
|5| PARAMS_PAS_SECURE |
|6| HASH_ALGO_INVALID TA|
|7| HASH_ALGO_UNSUPPORTED |
|8| RTT_BASE_RD_ALIASED |
|9| RD_UNALIGNED |
|10| RD_DEV_MEM  |
|11| RD_OUTSIDE_OF_PERMITTED_PA |
|12| RD_STATE_UNDELEGATED |
|13| RD_STATE_RD |
|14| RD_STATE_REC |
|15| RD_STATE_RTT |
|16| RD_STATE_DATA |
|17| RTT_UNALIGNED |
|18| FEAT_REG_INVALID |
|19| RTT_START_INVALID |
|20| RTT_BASE_UNDELEGATED  |
|21| VMID_INVALID |
|22| VMID_USED |

### DIFF
--- a/rmm/monitor/src/rmi/realm/mod.rs
+++ b/rmm/monitor/src/rmi/realm/mod.rs
@@ -3,6 +3,7 @@ pub(crate) mod rd;
 
 use self::params::Params;
 pub use self::rd::Rd;
+use super::error::Error;
 use crate::event::Mainloop;
 use crate::host::pointer::Pointer as HostPointer;
 use crate::listen;
@@ -21,6 +22,10 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     listen!(mainloop, rmi::REALM_CREATE, |arg, ret, rmm| {
         let rmi = rmm.rmi;
         let mm = rmm.mm;
+
+        if arg[0] == arg[1] {
+            return Err(Error::RmiErrorInput);
+        }
 
         // get the lock for granule.
         let mut rd_granule = get_granule_if!(arg[0], GranuleState::Delegated)?;
@@ -42,6 +47,9 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             ret[1] = id;
         })?;
 
+        if arg[0] == rd.rtt_base() {
+            return Err(Error::RmiErrorInput);
+        }
         let mut rtt_granule = get_granule_if!(rd.rtt_base(), GranuleState::Delegated)?;
         set_granule(&mut rtt_granule, GranuleState::RTT)?;
 
--- a/rmm/monitor/src/rmi/realm/mod.rs
+++ b/rmm/monitor/src/rmi/realm/mod.rs
@@ -30,6 +30,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // read params
         let params = copy_from_host_or_ret!(Params, arg[1], mm);
         trace!("{:?}", params);
+        params.validate()?;
 
         // TODO:
         //   Manage vmid
--- a/rmm/monitor/src/rmi/realm/mod.rs
+++ b/rmm/monitor/src/rmi/realm/mod.rs
@@ -32,7 +32,9 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         rmm.mm.map(rd, true);
 
         let params = copy_from_host_or_ret!(Params, params_ptr, rmm.mm);
-        params.validate(rd)?;
+        if params.rtt_base as usize == rd {
+            return Err(Error::RmiErrorInput);
+        }
 
         // revisit rmi.create_realm() (is it necessary?)
         rmm.rmi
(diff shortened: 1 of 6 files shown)


rmm/monitor/src/rmi/realm/params.rs
```diff
@@ -21,6 +24,37 @@ pub struct Params {
 
 const_assert_eq!(core::mem::size_of::<Params>(), GRANULE_SIZE);
 
+impl Params {
+    pub fn validate(&self, rd: usize) -> Result<(), Error> {
```
jinbpark, line 28: I thought this `validate()` can be replaced with `HostAccessor.validate()`, but it seems it can't, as `HostAccessor.validate()` doesn't take an argument. (in this case, "rd").
So, I think we have no choice but to keep this `validate()` function.

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -18,36 +19,38 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
-
-        // set Rd state only when everything goes well.
-        set_granule(&mut rd_granule, GranuleState::RD)?;
-
-        Ok(())
+    listen!(mainloop, rmi::REALM_CREATE, |arg, _, rmm| {
+        let rd = arg[0];
+        let params_ptr = arg[1];
+
+        if rd == params_ptr {
```
jinbpark, line 26: Just thought:
we might be able to check this constraint at a higher level, (e.g., RMI dispatch function)
by adding a rule like "all arguments indicating "address" must not equal to each other",
if this is true in every case in the spec.

rmm/monitor/src/rmi/realm/mod.rs
```diff
@@ -18,36 +19,38 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+            return Err(Error::RmiErrorInput);
+        }
+
+        let mut rd_granule = get_granule_if!(rd, GranuleState::Delegated)?;
+        let rd_obj = rd_granule.content_mut::<Rd>();
+        rmm.mm.map(rd, true);
```
jinbpark, line 32: It seems that we need to call `unmap(rd)` when something goes wrong.
I think `unmap` can be added into `epilog.map_err()`.
PS:  a todo comment relevant to this is in "granule/mod.rs". (i.e., putting map/unmap into a granule transaction)

rmm/monitor/src/rmi/realm/params.rs
bitboom: Oh. I forgot `HostAccessor::validate`. I thought it is more proper place to validate. I reflected it and checked `rd == rtt_base` after copying.

rmm/monitor/src/rmi/realm/mod.rs
bitboom: It sounds great!  Spec noted `Failure Conditions` of commands. I thinks it's really good to check in constraint even it is hard job :)

rmm/monitor/src/rmi/realm/mod.rs
bitboom: You are right. I reflected it.

rmm/monitor/src/rmi/realm/params.rs
```diff
@@ -54,7 +56,34 @@ impl core::fmt::Debug for Params {
...
-impl HostAccessor for Params {}
+impl HostAccessor for Params {
+    fn validate(&self) -> bool {
+        trace!("{:?}", self);
+        if !features::validate(self.features_0 as usize) {
+            return false;
+        }
+
+        // Check misconfigurations between IPA size and SL
+        let ipa_bits = features::ipa_bits(self.features_0 as usize);
```
zpzigi754, line 67: Good to extract ipa bits from `features_0` parameter. This could be connected with https://github.com/Samsung/islet/pull/156/commits/d48bab786c706b034a8284e55c040f5e3cea51d5 later.

zpzigi754 (approved):

rmm/monitor/src/rmi/realm/mod.rs
zpzigi754: If `unmap(rd)` should be called in any error case, I guess that there could exist a conflict in using the `? (try)` operator in realm creation, as `unmap` would be missed.

jinbpark (approved):

rmm/monitor/src/rmi/realm/mod.rs
bitboom: @zpzigi754 Yes right.  I hope below job resolves it !

>  PS: a todo comment relevant to this is in "granule/mod.rs". (i.e., putting map/unmap into a granule transaction)

PR #156, opened 2023-08-28
Add support for virtio

zpzigi754:
This PR enables realm to use virtio (e.g., file system access) by allowing shared pages to be accessed by the Host. With this PR, realm-linux will be able to boot to the shell with islet.

### Memory sharing between Realm and the Host

Realm can modify its private region into a shared region by invoking `RSI_IPA_STATE_SET` with RIPAS:empty (Figure D2.1 Realm shared memory protocol flow in rmm-spec). With the following `RMI_RTT_SET_RIPAS`, this triggers a set of actions by the Host (e.g., `RTT_SET_RIPAS`->`RTT_READ_ENTRY`->`DATA_DESTROY`) according to the value of RTT entry for freeing regions (For more about detail, please refer to realm_tear_down_rtt_range() in nw-linux). This PR fills this flow. 

### About fault redirection

`linux-cca` controls the accesses to the encrypted/decrypted page with the bit 33 in Realm's page table entry (the bit location can be configured by the user). The previous `RSI_REALM_CONFIG` returns an incorrect value which resulted in improper fault generation. For example, the access to a decrypted page (e.g., 0x184c00000) always triggered a fault for its encrypted page (e.g., 0x84c00000), while it should have been treated as the decrypted page. With the fix, the two different kind of accesses can be distinguished and handled properly. 

### Using level in page table traversing

In page table traversing API, (input) level and (output) level are added which is useful for the fine-grained traversing. (input) level indicates the intended (maximum) level to reach, while (output) level is the actual level reached (e.g., If the user passes 3 for (input) level and there are upto level 2 page tables for the given address, it would return 2 for the (output) level).

### Additional Notes

Currently, `islet`'s page table traversing (e.g., ipa_to_pa()) in default checks its validity with bit 0. However, this cannot be done in stage 2 page table, as bit [0-1] is used as [description type](https://github.com/Samsung/islet-asset/blob/708eb9d6ef16700c38aa89232e63e8bdea18c486/lib/realm/src/s2tt.c), and valid page table entry has that bit on. 
In this PR, `ipa_to_pte()` is added and this function does not conduct the bit 0 based validity check (here `no_valid_check`: true is passed in `entry()`) and returns the whole entry's value.

Also, page table entry modification in default attaches unintended flags which is not compatible with stage 2 table entry (e.g., RIPAS, HIPAS field). To avoid incorrect bits to be set, I've added `is_raw` parameter in PTE modification APIs, which when set as true, only sets the flags that the users have provided.

### Remaining issue
- TF-A failure case (will be resolved in another PR)

### DIFF
--- a/rmm/armv9a/src/realm/registry.rs
+++ b/rmm/armv9a/src/realm/registry.rs
@@ -15,11 +15,14 @@ use crate::realm;
 use crate::realm::context::Context;
 use crate::realm::mm::page_table::pte;
 use crate::realm::mm::stage2_translation::Stage2Translation;
+use crate::realm::mm::stage2_tte::invalid_ripas;
+use crate::realm::mm::stage2_tte::S2TTE;
 use crate::realm::mm::translation_granule_4k::RawPTE;
 use crate::realm::timer;
 use monitor::realm::config::RealmConfig;
 use monitor::rmi::error::Error;
 use monitor::rmi::error::InternalError::*;
+use monitor::rmi::rtt_entry_state;
 
 use alloc::boxed::Box;
 use alloc::collections::BTreeMap;
@@ -449,5 +452,53 @@ impl monitor::rmi::Interface for RMI {
         } else {
             Err(Error::RmiErrorInput)
         }
+    }
+
+    fn rtt_read_entry(&self, id: usize, ipa: usize, level: usize) -> Result<[usize; 4], Error> {
+        let (s2tte, last_level) = get_realm(id)
+            .ok_or(Error::RmiErrorOthers(NotExistRealm))?
+            .lock()
+            .page_table
+            .lock()
+            .ipa_to_pte(GuestPhysAddr::from(ipa), level)
+            .ok_or(Error::RmiErrorRtt)?;
+        let r1 = last_level;
+        let (mut r2, mut r3, mut r4) = (0, 0, 0);
+
+        let s2tte = S2TTE::from(s2tte as usize);
+
+        if s2tte.is_unassigned() {
+            let ripas = s2tte.get_masked_value(S2TTE::INVALID_RIPAS);
+            r2 = rtt_entry_state::RMI_UNASSIGNED;
+            r4 = ripas as usize;
+        } else if s2tte.is_destroyed() {
+            r2 = rtt_entry_state::RMI_DESTROYED;
+        } else if s2tte.is_assigned() {
+            r2 = rtt_entry_state::RMI_ASSIGNED;
+            r3 = s2tte.address(last_level).ok_or(Error::RmiErrorRtt)?.into();
+            r4 = invalid_ripas::EMPTY as usize;
+        } else if s2tte.is_valid(last_level, false) {
+            r2 = rtt_entry_state::RMI_ASSIGNED;
+            r3 = s2tte.address(last_level).ok_or(Error::RmiErrorRtt)?.into();
+            r4 = invalid_ripas::RAM as usize;
+        } else if s2tte.is_valid(last_level, true) {
+            r2 = rtt_entry_state::RMI_VALID_NS;
+            let addr_mask = match level {
+                1 => S2TTE::ADDR_L1_PAGE,
+                2 => S2TTE::ADDR_L2_PAGE,
+                3 => S2TTE::ADDR_L3_PAGE,
+                _ => {
+                    return Err(Error::RmiErrorRtt);
+                }
+            };
+            let mask = addr_mask | S2TTE::MEMATTR | S2TTE::AP | S2TTE::SH;
+            r3 = (s2tte.get() & mask) as usize;
+        } else if s2tte.is_table(last_level) {
+            r2 = rtt_entry_state::RMI_TABLE;
+            r3 = s2tte.address(3).ok_or(Error::RmiErrorRtt)?.into();
+        } else {
+            error!("Unexpected S2TTE value retrieved!");
+        }
+        Ok([r1, r2, r3, r4])
     }
 }
--- a/rmm/armv9a/src/realm/registry.rs
+++ b/rmm/armv9a/src/realm/registry.rs
@@ -15,8 +15,8 @@ use crate::realm;
 use crate::realm::context::Context;
 use crate::realm::mm::page_table::pte;
 use crate::realm::mm::stage2_translation::Stage2Translation;
-use crate::realm::mm::stage2_tte::invalid_ripas;
 use crate::realm::mm::stage2_tte::S2TTE;
+use crate::realm::mm::stage2_tte::{desc_type, invalid_hipas, invalid_ripas};
 use crate::realm::mm::translation_granule_4k::RawPTE;
 use crate::realm::timer;
 use monitor::realm::config::RealmConfig;
@@ -185,7 +185,8 @@ impl monitor::rmi::Interface for RMI {
                 PhysAddr::from(phys),
                 size,
                 flags as usize,
-            );
+                false,
+            )?;
 
         Ok(())
     }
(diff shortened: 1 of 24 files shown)


jinbpark:
Not ready for review, but I want to say that this is a huge step forward! Great work in advance!

zpzigi754:
> Not ready for review, but I want to say that this is a huge step forward! Great work in advance!

Thank you for the comment. Now, it gets ready!

rmm/monitor/src/rmi/rtt.rs
```diff
@@ -51,19 +52,44 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+        let rd_granule = get_granule_if!(arg[0], GranuleState::RD)?;
+        let rd = rd_granule.content::<Rd>();
+        let ipa = arg[1];
+        let level = arg[2];
+
+        let res = rmi.rtt_read_entry(rd.id(), ipa, level)?;
+        ret[1] = res[0];
+        ret[2] = res[1];
+        ret[3] = res[2];
+        ret[4] = res[3];
```
bitboom, line 91: [minor] How about this?

```suggestion
        ret[1..5].clone_from_slice(&res[0..4]);
```

bitboom (approved):
What a great PR !!

rmm/armv9a/src/realm/registry.rs
```diff
@@ -15,10 +15,14 @@ use crate::realm;
...
 use crate::realm::mm::page_table::pte;
 use crate::realm::mm::stage2_translation::Stage2Translation;
+use crate::realm::mm::stage2_tte::S2TTE;
+use crate::realm::mm::stage2_tte::{desc_type, invalid_hipas, invalid_ripas};
 use crate::realm::mm::translation_granule_4k::RawPTE;
 use crate::realm::timer;
+use monitor::realm::config::RealmConfig;
 use monitor::rmi::error::Error;
 use monitor::rmi::error::InternalError::*;
+use monitor::rmi::rmi_rtt_entry_state;
```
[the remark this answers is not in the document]
zpzigi754: Good to point it out! I've renamed it.

rmm/monitor/src/rmi/rtt.rs
zpzigi754: Good suggestion. I've changed it with `copy_from_slice()`.

rmm/monitor/src/realm/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    // in parsing the following kernel cmdline argument:
+    // `console=ttyS0 root=/dev/vda rw  console=pl011,mmio,0x1c0a0000 console=ttyAMA0 printk.devkmsg=on`.
+    // So, we get back to use the same kernel argument with TF-RMM's one (uart0 & uart3).
+    pub unsafe fn new(
+        config_addr: usize,
+        ipa_width: usize,
+    ) -> ManuallyDrop<&'static mut RealmConfig> {
+        let config: &mut RealmConfig = &mut *(config_addr as *mut RealmConfig);
+        config.ipa_width = ipa_width;
+        ManuallyDrop::new(config)
```
jinbpark, line 24: I think it would be better to do this write without ManuallyDrop, like the following.
(because we're trying to avoid ManuallyDrop if possible)
```
init() {
   let config: &mut RealmConfig = &mut *(config_addr as *mut RealmConfig);
   config.ipa_width = ipa_width;
}
ipa_width() -> usize {
  let config: & RealmConfig = &*(config_addr as *RealmConfig);
   config.ipa_width
}
```

Plus, in a long run, I think we need a secure interface, like we did for HostPointer` to access "realm space" from RMM. (I'm not sure there are how many cases like this)

rmm/monitor/src/rmi/rtt.rs
```diff
@@ -51,19 +52,40 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+            return Err(Error::RmiErrorInput);
+        }
+
         let map_size = level_to_size(level);
+        if ipa as u64 + map_size > rec.ripas_end() {
+            return Err(Error::RmiErrorInput);
+        }
+
+        if ripas as u64 == RIPAS_EMPTY {
+            let ret = rmi.make_shared(rd.id(), ipa, level);
```
jinbpark, line 70: Just question:  is the given level always "3" (the last level == page) throughout the realm-linux booting?

rmm/monitor/src/rmi/rtt.rs
```diff
@@ -148,15 +170,14 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
         let rd = rd_granule.content::<Rd>();
         let realm_id = rd.id();
         let ipa = arg[1];
 
-        // TODO: Fix to get PA by rtt_walk
-        let pa = rmm.rmi.unmap(realm_id, ipa, GRANULE_SIZE)?;
+        let pa = rmm.rmi.data_destroy(realm_id, ipa)?;
 
         // data granule lock and change state
         let _ = set_state_and_get_granule!(pa, GranuleState::Delegated);
```
jinbpark, line 183: Sorry, but could you change this code?
It seems my mistake from the PR of granule transaction. There is no reason to ignore an error.
```
// before
let _ = set_state_and_get_granule!(...);
Ok(())

// after
set_state_and_get_granule!(...)
```

rmm/monitor/src/rmi/rtt.rs
```diff
@@ -167,19 +188,15 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
         let realm_id = rd.id();
 
-        // islet stores rd as realm id
-        let granule_sz = GRANULE_SIZE;
-        let mut prot = rmi::MapProt(0);
-        prot.set_bit(rmi::MapProt::NS_PAS);
-        let _ret = rmi.map(realm_id, ipa, ns_pa, granule_sz, prot.get());
+        let level = arg[2];
+        let host_s2tte = arg[3];
+        rmi.rtt_map_unprotected(realm_id, ipa, level, host_s2tte)?;
```
jinbpark, line 199: Just question:
doing RTT_MAP_PROTECTED through `rtt_map_unprotected()` but RTT_UNMAP_UNPROTECTED through the existing `unmap()`. Is it ok? (I guess it would be ok because "unset_page" typically set 0)

scripts/fvp-cca
```diff
@@ -310,7 +310,8 @@ def place_realm_at_shared(rmm, realm_ip, fvp_tap_ip, host_ip):
     run(["cp", "%s/lkvm" % OUT, SHARED_PATH], cwd=ROOT)
     run(["cp", "-R", "%s/kvm-unit-tests" % OUT, SHARED_PATH], cwd=ROOT)
     if rmm == "islet":
-        run(["cp", LAUNCH_REALM_ISLET, LAUNCH_REALM], cwd=ROOT)
+        # Use the TF-RMM's one until the issue has been resolved 
+        run(["cp", LAUNCH_REALM_TFRMM, LAUNCH_REALM], cwd=ROOT)
```
jinbpark, line 314: Is it to get around "tf-a-failure"?

rmm/monitor/src/realm/config.rs
zpzigi754: Good suggestion! I will remove `ManuallyDrop`, as suggested. I totally agree with using a secure interface between RMM and Realm (not sure either how many there are).

scripts/fvp-cca
zpzigi754: Oh, not really. That one is related to the comment I left in `rmm/monitor/src/realm/config.rs` like the below.

```
// For some reason, using 33 for ipa_width causes a problem (format string bug?)
// in parsing the following kernel cmdline argument:
// `console=ttyS0 root=/dev/vda rw  console=pl011,mmio,0x1c0a0000 console=ttyAMA0 printk.devkmsg=on`.
// So, we get back to use the same kernel argument with TF-RMM's one (uart0 & uart3).
```

`parse_early_param()` in realm-linux caused a hang.

rmm/monitor/src/rmi/rtt.rs
zpzigi754: Frankly speaking, I am not sure, because I've not tested `unmap_unprotected`. Actually, there seems no call to `RTT_UNMAP_PROTECTED` in realm-linux booting, while there are several calls to `RTT_MAP_PROTECTED`.

rmm/monitor/src/rmi/rtt.rs
zpzigi754: Reflected like the below, as `set_state_and_get_granule`  returns () type.
```
set_state_and_get_granule!(...)?
Ok(())
```

rmm/monitor/src/realm/config.rs
zpzigi754: Reflected :)

scripts/fvp-cca
jinbpark: I see. It is just a command line configuration for lkvm.

jinbpark (approved):

Issue #163, opened 2023-08-29
KVM_CAP_RME(KVM_CAP_ARM_RME_CREATE_RD): No such device or address


zpzigi754:
Realm-linux booting currently fails in the early stage.

The below is the log with [the latest commit](https://github.com/Samsung/islet/commit/73819b6ae4c14e736344ca0bfc0438fe62896a20).

```
./launch-realm.sh
  # lkvm run -k linux.realm -m 256 -c 1 --name guest-702^M
  Info: (arm/aarch64/kvm.c) kvm__get_vm_type:194: max_ipa 8fffffff ipa_bits 33 max_ipa_bits 48^M
KVM_CAP_RME(KVM_CAP_ARM_RME_CREATE_RD): No such device or address
```

PR #164, opened 2023-08-29
rmm: Remove an invalid feature check

zpzigi754:
This PR resolves https://github.com/Samsung/islet/issues/163. The previous check did not consider the case when FEAT_LPA2 is implemented and caused a failure in realm linux booting.

### DIFF
--- a/rmm/monitor/src/rmi/features.rs
+++ b/rmm/monitor/src/rmi/features.rs
@@ -10,6 +10,7 @@ const S2SZ_WIDTH: usize = 8;
 const S2SZ_VALUE: usize = 48;
 
 const LPA2_SHIFT: usize = 8;
+#[allow(unused)]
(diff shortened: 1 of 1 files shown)


bitboom (approved):
I am sorry that i didn't check realm-linux booting :(
After realm-linux booting PR is merged, we have to test it on CI.

--- 

Anyway we have to read `AA64MMFR_EL1` in monitor crate. I didn't add it because we cannot run `cargo test` if we add `asm` code to `monitor`.

So, i think it's time to refactor whole structures.

1. Detach `RMI data structure & commands` to another crate
2. Divide some modules in `monitor` for testing
3. **Merge `armv9a` and `monitor`**

zpzigi754:
I also agree with merging `armv9a` and `monitor`!

Issue #167, opened 2023-09-01
ACS test failure (delegation) on latest commit


bitboom:
## Cause
ACS tries a maximum walk and check where to create the next table entry by `rtt_read_entry` before mapping addr. But our rmm returns failure `RMI_RTT_ERROR`.

## Failure log
```
 23 ^M..Check 6 : gran_state; intent id : 0x5
 24 ^M..ReadEntry query failed!.(Check failed at:/home/sangwan/islet/third-party/cca-rmm-acs/test/command/common/command_common_host.c ,line:261)
 25 ^M..Couldn't create the assigned protected mapping 0x1000
 26 ^M..(Check failed at:/home/sangwan/islet/third-party/cca-rmm-acs/test/command/common/command_common_host.c ,line:377)
 27 ^MResult => Failed (Error Code=3)
```

PR #168, opened 2023-09-01
rmi: Fix RMI_ERROR_RTT condition

bitboom:
This PR fixes https://github.com/Samsung/islet/issues/167

* Spec 3.4.22 RmiStatusCode type RMI_ERROR_RTT: An RTT walk terminated before reaching the target RTT level, or reached an RTTE with an unexpected value.


zpzigi754 (approved):
Thank you for fixing the bug!

PR #169, opened 2023-09-05
ci: Add cross test & Update rust toolchain

bitboom:
This PR is pre-step to re-organize `islet-rmm` dependencies.  _(`monitor <- armv9a` => `monitor -> armv9a`)_

BTW, monitor should be tested by `cargo test`. So, i add below commits.
- Add [cross testing](https://docs.rs/crate/cross/0.2.5) to test `monitor crate` as aarch64
- Update rust toolchain to latest (1.74)


zpzigi754:
Not about this PR, but according to the PR description, it is mentioned that you are thinking of changing the current dependency from `monitor <- armv9a` to `monitor -> armv9a`. Why should `monitor` depend on `armv9a` (the reverse order)?

BTW, I was thinking of removing the entire dependency between `armv9a` and `monitor` by merging them as one.

zpzigi754:
How about changing the below files too?
```
./Cargo.toml
./sdk/Cargo.toml
./cli/Cargo.toml
./lib/rsi-el0/Cargo.toml
```

bokdeuk-jeong:
Can we continue discussing the reorganization at https://github.com/Samsung/islet/discussions/170? This topic requires participation and opinions of all of us.

.github/workflows/ci.yml
```diff
@@ -71,11 +71,19 @@ jobs:
...
+      run: |
+        ./scripts/deps/rust.sh
+        ./scripts/deps/docker.sh
 
     - name: Run crates test
       run: ./scripts/tests/crates.sh
 
+    - name: Run crates test (aarch64)
+      run: |
+        sudo systemctl start docker
```
zpzigi754: Which docker image does this line use (for the cross testing)?

bitboom:
@zpzigi754 @bokdeuk-jeong 
Sure, i am happy to discuss it. 😄 

BTW, I think this PR could be merged regardless of that. Because testing on aarch64 is better. And latest toolchain is also better :)

.github/workflows/ci.yml
bitboom: Now we uses default image of cross: `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:0.2.5`.
We can change it.

bitboom:
> Not about this PR, but according to the PR description, it is mentioned that you are thinking of changing the current dependency from `monitor <- armv9a` to `monitor -> armv9a`. Why should `monitor` depend on `armv9a` (the reverse order)?

I will update my opinion to https://github.com/Samsung/islet/discussions/170 tomorrow !

zpzigi754 (approved):
It's interesting to run `aarch64-unknown-linux-gnu` target without emulation.

bokdeuk-jeong (approved):

nook1208:
I have a problem to build islet project because of the low version of rust toolchain. 
So this PR is really helpful to me to fix the build error. Thank you !

PR #172, opened 2023-09-05
[draft] ci: Add a check for realm-linux booting

zpzigi754:


scripts/tests/realm-boot.sh
```diff
@@ -0,0 +1,35 @@
...
+	ps -ef | grep FVP_Base_RevC-2xAEMvA | grep -v grep | awk '{print $2}' | xargs kill
+
+	# report
+	LOGIN=$(tail -30 $UART | grep "login" | awk '{print $2}')
+	LOGIN="${LOGIN//[$'\t\r\n ']/}"
+
+	if [ "$LOGIN" == "" ]; then
+		echo "[-] Test result: Realm booting failed with the following log"
+		tail -30 $UART
+		echo "Try increasing the time for sleep"
```
bitboom, line 22: FYI, I wrote shell script like this purpose before.

```sh
prev=0
curr=1

while [ "$prev" -ne "$curr" ]; do
	prev=$curr
	sleep 30
	curr=$(wc -l $UART | awk '{print $1}')
	echo "prev = $prev, curr = $curr"
done
```

zpzigi754: Thank you for the reference. It would be good to call sleep repeatedly to find some condition.

zpzigi754:
I will reopen this PR after resolving the issue regarding docker environment.

Issue #176, opened 2023-09-08
nw-linux booting does not succeed in a docker image


zpzigi754:
`nw-linux` fail to boot with the following log. 

```
...
[    0.115567] hw-breakpoint: found 16 breakpoint and 16 watchpoint registers.
[    0.116545] ASID allocator initialised with 65536 entries
[    0.118215] Serial: AMBA PL011 UART driver
[    0.140566] 1c090000.serial: ttyAMA0 at MMIO 0x1c090000 (irq = 14, base_baud = 0)
```
The used kernel parameter is `console=tty0 console=ttyAMA0,115200 earlycon=pl011,0x1c090000 root=/dev/disk/by-partlabel/system rootwait rw ignore_loglevel efi=noruntime cpuidle.off=1`.

I've confirmed that this symptom occurs only in a docker environment including github actions. Both of `tf-rmm` and `islet` have this issue.

PR #178, opened 2023-09-11
ci: Add a check for realm-linux booting

zpzigi754:
This PR enables realm-linux booting check in ci. It also includes
- Using modified `boot.img` to execute `launch-realm.sh` script at the end of `nw-linux`'s booting
- Reducing output size (from 1GB to 130MB) to avoid a github action's storage limitation
- Skipping a slow building procedure by using prebuilt binary
- Disabling telnet in FVP to address a docker [issue](https://github.com/Samsung/islet/issues/176)


bitboom (approved):
Really nice !!

PR #181, opened 2023-09-13
Refactoring paging related modules

bitboom:
This PR is about refactoring.
There are many files changed but no functional changes.

### Main purpose of this PR
There is no need to expose several APIs from `rmm` to `plat/fvp`. (except for configuration APIs)
So, this PR exposes `rmm_entry()` to `plat/fvp` and removed extra APIs from `plat/fvp`.
_After this PR merged, i will fix API visibility of `rmm`._

```rust
// rmm crate
pub fn rmm_entry() {
    unsafe {
        setup_mmu_cfg();
        setup_el2();
        enable_mmu_el2();
        setup_gst();
    }

    Monitor::new().run();
}

// plat/fvp crate
[no_mangle]
pub unsafe fn main() -> ! {
    ...
    islet_rmm::rmm_entry();
    ...
}
```

### About paging
There are three page_tables (RMM, GranuleStateTable, Realm).
So I detached paging module to `lib/paging` and use it.
And I move RMM's page_table to `Monitor`. So we can update page_table of RMM in RMI handler.

```rust
pub struct Monitor {
    pub rmi: RMI,
    pub rsi: RsiHandle,
    pub page_table: PageTable,
}

listen!(mainloop, rmi::GRANULE_DELEGATE, |arg, _, rmm| {
   ...
   rmm.page_table.map(addr, true);
   ...
}
```


rmm/src/lib.rs
```diff
@@ -85,10 +92,9 @@ pub unsafe fn init_el2() {
     SCTLR_EL2.set(SCTLR_EL2::I | SCTLR_EL2::M | SCTLR_EL2::EOS);
     CPTR_EL2.set(CPTR_EL2::TAM);
     ICC_SRE_EL2.set(ICC_SRE_EL2::ENABLE | ICC_SRE_EL2::DIB | ICC_SRE_EL2::DFB | ICC_SRE_EL2::SRE);
-    activate_stage2_mmu();
 }
 
-unsafe fn activate_stage2_mmu() {
+unsafe fn enable_mmu_el2() {
```
zpzigi754, line 115: I think that the original name conveys the meaning more clearly (stage 2)

rmm/src/lib.rs
```diff
@@ -107,6 +113,42 @@ unsafe fn activate_stage2_mmu() {
     core::arch::asm!("tlbi alle2", "dsb ish", "isb",);
 }
 
+unsafe fn setup_mmu_cfg() {
```
zpzigi754, line 134: Is there any reason for not using the previous `set_register_mm()` which had almost the same code?

lib/paging/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "paging"
```
zpzigi754, line 2: Paging usually refers to `storing and retrieving data from a device’s secondary storage (disk) to the primary storage (ram)`. So, I am a bit worried that having this name might give out a confusion.

rmm/src/lib.rs
```diff
@@ -39,35 +40,50 @@ extern crate lazy_static;
...
+    core::arch::asm!(
+        "bl rmm_exit",
+        inlateout("x0") args[0] => ret[0],
+        inlateout("x1") args[1] => ret[1],
+        inlateout("x2") args[2] => ret[2],
+        inlateout("x3") args[3] => ret[3],
+    );
+
+    ret
+}
```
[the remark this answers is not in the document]
bitboom: I wanted to expose APIs(`pub fn`) at top of module.

lib/paging/Cargo.toml
bitboom: > Paging usually refers to `storing and retrieving data from a device’s secondary storage (disk) to the primary storage (ram)`.

I thought that sentence means `swapping`

> So, I am a bit worried that having this name might give out a confusion.

Google also uses paging name in [aarch64-paging](https://github.com/google/aarch64-paging).
Do you want to use other name? (eg. mm)

rmm/src/lib.rs
bitboom: `set_register_mm` located in `rmm/mm/translation`. I had to remove `rmm` module and i collected `setup functions` here.
The reason of renaming is that i though `setup_mmu_cfg` is more intuitive. _(tf-rmm used)_ 

To be honest, I like rhyming..
```sh
setup_mmu_cfg()
enable_mmu()
```

lib/paging/Cargo.toml
bokdeuk-jeong: I second to zpzigi754's opinion.

rmm/src/host/mod.rs
```diff
@@ -1,30 +1,31 @@
...
     /// Try to do page-relevant stuff (e.g., RMM map).
     /// returns true only if everything goes well.
-    fn acquire(ptr: usize, page_map: PageMap) -> bool {
+    fn acquire(ptr: usize) -> bool {
         if !validate_addr(ptr) {
             return false;
         }
         // TODO: check if the granule state of `ptr` is Undelegated.
-        page_map.map(ptr, false)
+        PageTable::new().map(ptr, false)
```
bokdeuk-jeong, line 20: Isn't this line for creating a new mapping to a page table? It is a page table entry that it newly created while the page table might be already existing. The name "new()" is generally used when we create a new object or data structure. Does rust commonly use the term "new()" even when we get the reference to an existing object?

rmm/src/mm/translation.rs
```diff
@@ -35,17 +56,18 @@ fn get_page_table() -> u64 {
 pub const NUM_ROOT_PAGE: usize = 1;
 pub const ALIGN_ROOT_PAGE: usize = 2;
 
-pub struct RmmPageTable<'a> {
+pub struct Inner<'a> {
```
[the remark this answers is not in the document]
bitboom: Thanks for pointing out! I remove `pub`.

rmm/src/lib.rs
bitboom: I restored it !

rmm/src/host/mod.rs
bitboom: I fixed it to `get_ref`.

lib/paging/Cargo.toml
zpzigi754: > I thought that sentence means `swapping`

paging is actually used interchangeably with swapping (Reference: https://en.wikipedia.org/wiki/Memory_paging). 
```
memory paging (or swapping on some Unix-like systems) is a memory management scheme 
by which a computer stores and retrieves data from secondary storage for use in main memory
```

> Google also uses paging name in [aarch64-paging](https://github.com/google/aarch64-paging). Do you want to use other name? (eg. mm)

Paging sometime seems to indicate things related to page table like the one mentioned in google's library. 
using `mm` or `page-table` would be fine I guess.

bitboom: `vmsa` looks cool. I reflected it :) I hope we could move VMSA related code more from `rmm` to `/lib/vmsa`.

rmm/src/host/mod.rs
bitboom: > Does rust commonly use the term "new()" even when we get the reference to an existing object?

It was my miss. rust uses `get_ref`.

lib/paging/Cargo.toml
bitboom: Thanks for the reference!

> Paging sometime seems to indicate things related to page table like the one mentioned in google's library. using `mm` or `page-table` would be fine I guess.

Oh,,, i just saw this comment. `page_table` is in the same opinion with you. @zpzigi754 Could you want to use `page_table` more than `vmsa`? Feel free to tell me :)

zpzigi754: using vmsa would be also fine :)

zpzigi754 (approved):

rmm/src/lib.rs
```diff
@@ -38,30 +40,24 @@ extern crate lazy_static;
...
 
 use armv9a::{bits_in_reg, regs::*};
 
-pub struct Monitor {
-    pub rmi: RMI,
-    pub rsi: RsiHandle,
-    pub mm: PageMap,
-}
+/// RMM entry point
+pub unsafe fn rmm_entry() {
```
bokdeuk-jeong, line 51: Can we change rmm_entry() to rmm_setup()?
We already have the function name rmm_exit() which is used for exiting the rmm at EL2 in realm world. 
As the term 'entry' often pairs with 'exit'. rmm_entry() might be misunderstood in the way that it is called evertime when cpu enters rmm.

bitboom: We used `rmm_enter` & `rmm_exit` when switch between EL1 & EL2. So i choose `entry` word which used at linker script ususally.

I agree with your opinion. So I changed it to `start`. Because this function include running mainloop too.

bokdeuk-jeong (approved):

Issue #166, opened 2023-08-31
Identify the cause of the increased RMM memory usage


bokdeuk-jeong:
We recently had to increase the heap size in RMM from 8MB to 16MB. 
Identify which module requires more memory for what purpose and how much it demands.

nook1208:
I think it's good to make a memory stat feature to check the RMM memory status. 
Let me check the RMM code and then I'll push a PR for it

nook1208:
I found that 'RMI::GRANULE_DELEGATE' command use RMM Heap memory the most.
I uploaded the memory stat result below wiki page: 
https://code.sec.samsung.net/confluence/display/SYSSECC/Stat

Let me finish #182 PR and then go further with RMI::GRANULE_DELEGATE command

bokdeuk-jeong:
Could you share the result here? The link that you shared cannot be accessed in public.

jinbpark:
This is mainly because,
we're managing GranuleStatusTable (GST) with a 2-level page table. (We manage RTT with a 3-level page table, that might be the reason why GST takes up a larger portion of memory than RTT)

One optimization for this is to use a 3-level page table for GST. But, to do so, https://github.com/Samsung/islet/issues/147 --> this refactoring task might be needed to get done first.

PR #182, opened 2023-09-15
Add Stat module to check memory used size per RMI, RSI

nook1208:
This PR is related with #166

This `Stats` structure collects memory used size of RMM's heap per RMI, RSI command with `lock`.
Because of the `lock`, this module must be configured by something like `debug` build or `stat` feature.
At the moment, printing format like below: 
image

It could be extended to checks the `latency` of all RMI, RSI commands later


zpzigi754 (commented):
Good to add stat module!

I vote for using `feature`.

.gitignore
```diff
@@ -12,3 +12,8 @@ third-party/*
 cli/islet
 
 docs/
+
+# tags, cscope files
+cscope*
+tags*
+rusty-tags.vi*
```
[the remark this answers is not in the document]
nook1208: Actually I used terminal and vscode with rust-analyzer both. 
But as you said, I agree that handling cscope, tags in .gitignore is not necessary. I'll remove those two lines :)

bokdeuk-jeong (approved):

rmm/src/stat.rs
```diff
@@ -0,0 +1,302 @@
...
+#[repr(C)]
+#[derive(Copy, Clone, Debug, Default)]
+struct Stat {
+    kind: Kind,
+    overflowed: bool,
+    cur_cmd: Option<usize>,
+    call_cnt: [u64; MAX_CMD_CNT],
+
+    mem_used_before: Option<i64>,
+    total_mem_used: [i64; MAX_CMD_CNT],
```
zpzigi754, line 121: I guess that `u64` or `usize` could be used, if negative integer are unused for `total_mem_used`.

rmm/src/stat.rs
```diff
@@ -0,0 +1,302 @@
...
+            return Ok(());
+        }
+
+        if self.total_mem_used[idx] == 0 {
+            self.print_mem_stat(idx, cmd, 0);
+            return Ok(());
+        }
+
+        let avg_mem_used =
+            self.total_mem_used[idx].checked_div(self.call_cnt[idx].try_into().unwrap());
```
zpzigi754, line 163: If the type of `total_mem_used` is changed as `[u64; MAX_CMD_CNT]`, `self.call_cnt[idx].try_into().unwrap()` could be shortened as `self.call_cnt[idx]`.

zpzigi754 (approved):

rmm/src/event/mod.rs
```diff
@@ -69,8 +69,31 @@ impl Context {
...
+
+                if let Err(e) = stats.start(self.cmd) {
+                    info!("stats.start is failed: {:?} with cmd {:x}", e, self.cmd);
+                }
+                if let Err(code) = handler(&self.arg[..], &mut self.ret[..]) {
+                    self.ret[0] = code.into();
+                }
+                if let Err(e) = stats.end(self.cmd) {
+                    info!("stats.end is failed: {:?} with cmd {:x}", e, self.cmd);
+                }
```
bitboom, line 87: How about expose only `fn measure` in stat module like below? If then, we can abstract locking and others function.

```suggestion
stat::measure(self.cmd, || {
  if let Err(code) = handler(&self.arg[..], &mut self.ret[..]) {
    self.ret[0] = code.into();
   }
});
```

rmm/src/stat.rs
nook1208: Thank you for the review! 
But I'm afraid that I think there would be a case that the command which deallocate heap more. (e.g, REALM_DESTROY)
also I saw the case when I tested like this: 
```bash
[INFO]islet_rmm::stat -- RMI::REALM_DESTROY             TOTAL MemUsed:       -12416 byte, AVG MemUsed:      -3104 byte, CallCnt:        4
```

But it's good point ! thank you

rmm/src/event/mod.rs
nook1208: That's a good idea! I'll try it. thank you :) !

rmm/src/stat.rs
zpzigi754: If i64 is used for `total_mem_used`, how about using `self.call_cnt[idx] as i64` instead (just avoid using `unwrap`)?

nook1208: the same reply can be answered with previous one :)

nook1208: okay I'll accept it 👍

rmm/src/event/mod.rs
nook1208: I reflected it :)

rmm/src/stat.rs
nook1208: I reflected it :)

rmm/src/stat.rs
```diff
@@ -0,0 +1,313 @@
...
+    } else {
+        b
+    }
+}
+
+lazy_static! {
+    pub static ref STATS: Mutex<Stats> = Mutex::new(Stats::new());
+}
+
+#[inline(always)]
```
hihi-wang, line 35: Do we need this? `rustc` can decide whether or not to inline function depending on optimization level.

nook1208: Thank you for the comment !
I just want to do inlining regardless of the optimization level :)

hihi-wang: Just wonders is there any benefit for manually inline this.

rustc considers binary size if optimization level for "z" . i.e. rustc inlines function only if it reduces binary size
Otherwise, rustc considers performance. i.e. rustc inlines function only if it's faster

nook1208: In my opinion, Inline function can save some stack operations and it helpful for performance.
that's the main reason I used it. 
Do you think that there is benefit for not inlining those function except reducing the code size ?

nook1208: Oh thank you for the good reference!
I agreed that too many inlining is not good especially for really long function.
But in my case, those functions are good to use 'inline' because it's not too big (it's just one condition).
so I think that's `inline functions might make it faster` case according to the reference that you shared

FYI, It's just a simple condition but I made it to function for readability. that's my style :)

zpzigi754:
FYI, I recommend to use `rebase and merge` instead of `create a merge commit` for the linear git history and consistency.

nook1208:
Oops, that's my mistake. I actually tried to use `rebase` because the base commit is same. I'll use rebase next time !

PR #183, opened 2023-09-15
acs: Pass whole realm activate cases  & some data creation cases

bitboom:
This PR passes `whole realm activate cases` & `some data creation cases`.
The rest of data creation cases are related with `RTT walk`.
_@zpzigi754 is fixing about `RTT walk` now, so I will re-visit after his work :)_

The main patches of this PR are two
- Validate whether IPA, passed by host, is in PAR.
  - Because IPA of Realm should be in PAR(Protected Address Range)
- Implement `REALM_ACTIVATE`

### DIFF
--- a/rmm/src/rmi/rtt.rs
+++ b/rmm/src/rmi/rtt.rs
@@ -11,6 +11,8 @@ use crate::rmi;
 use crate::rmi::error::Error;
 use crate::{get_granule, get_granule_if, set_state_and_get_granule};
 
+use vmsa::error::Error as MmError;
+
 const RIPAS_EMPTY: u64 = 0;
 const RIPAS_RAM: u64 = 1;
 
@@ -92,12 +94,17 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rmi = rmm.rmi;
         // target_pa: location where realm data is created.
         let target_pa = arg[0];
+        let rd = arg[1];
         let ipa = arg[2];
         let src_pa = arg[3];
         let _flags = arg[4];
 
+        if target_pa == rd || target_pa == src_pa || rd == src_pa {
+            return Err(Error::RmiErrorInput);
+        }
+
         // rd granule lock
-        let rd_granule = get_granule_if!(arg[1], GranuleState::RD)?;
+        let rd_granule = get_granule_if!(rd, GranuleState::RD)?;
         let rd = rd_granule.content::<Rd>();
         let realm_id = rd.id();
 
@@ -106,6 +113,13 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         if !rd.at_state(State::New) {
             return Err(Error::RmiErrorRealm);
         }
+
+        validate_ipa(ipa, rd.ipa_bits())?;
+
+        match get_granule_if!(src_pa, GranuleState::Undelegated) {
+            Ok(_) | Err(MmError::MmNoEntry) => Ok(()),
+            _ => Err(Error::RmiErrorInput),
+        }?;
 
         // data granule lock for the target page
         let mut target_page_granule = get_granule_if!(target_pa, GranuleState::Delegated)?;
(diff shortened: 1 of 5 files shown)


bitboom:
> [minor] there is a typo in git commit message obtion -> option

Thanks! I fixed it.

rmm/src/rmi/rtt.rs
```diff
@@ -208,3 +222,23 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         Ok(())
     });
 }
+
+fn realm_ipa_size(ipa_bits: usize) -> usize {
+    1 << ipa_bits
+}
+
+fn realm_par_size(ipa_bits: usize) -> usize {
+    realm_ipa_size(ipa_bits) / 2
```
bokdeuk-jeong, line 231: why is this a half of ipa?

rmm/src/rmi/rtt.rs
```diff
@@ -208,3 +222,23 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+}
+
+fn validate_ipa(ipa: usize, ipa_bits: usize) -> Result<(), Error> {
+    if ipa % GRANULE_SIZE != 0 {
+        return Err(Error::RmiErrorInput);
+    }
+
+    if ipa >= realm_ipa_size(ipa_bits) || ipa >= realm_par_size(ipa_bits) {
```
bokdeuk-jeong, line 239: realm_ipa_size = realm_par_size * 2
Isn't the line 239 is equivalent to "ipa >= realm_par_size(ipa_bits)"?

rmm/src/rmi/rtt.rs
bitboom: The IPA space of a Realm is divided into two halves. `Un/Protected IPA Space`. `PAR` means protected address range.

rmm/src/rmi/rtt.rs
bitboom: You're right. I fixed it.

rmm/src/rmi/rtt.rs
```diff
@@ -107,6 +114,13 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             return Err(Error::RmiErrorRealm);
         }
 
+        validate_ipa(ipa, rd.ipa_bits())?;
+
+        match get_granule_if!(src_pa, GranuleState::Undelegated) {
+            Ok(_) | Err(MmError::MmNoEntry) => Ok(()),
```
zpzigi754, line 120: Is this a temporary workaround? (Return Ok in MnNoEntry case)

rmm/src/rmi/rtt.rs
```diff
@@ -107,6 +114,13 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             return Err(Error::RmiErrorRealm);
         }
 
+        validate_ipa(ipa, rd.ipa_bits())?;
```
zpzigi754, line 117: Good to add ipa validation!

zpzigi754 (commented):
Good to add ipa validation and realm activate!

rmm/src/rmi/rtt.rs
bitboom: It is not temporary but it needs refactoring.
For now, we can check whether the region is in NS in two ways.
1. `Undelegated`: it is mapped region before.
2. `MmNoEntry`: It hasn't mapped before.

I think `get_granule_if` should be fixed to return GranuleState::NS when it hasn't mapped before.
Or make `addr_in_ns` function.

bokdeuk-jeong: The purpose of this check is to inspect whether src_pa has been delegated ahead of this call rather than checking which pas src_pa is in.
Doesn't Err(MmError::MmNoEntry) mean that there is no mapping for src_pa in current granule management table.
So, shouldn't it be as below?
 Ok(_)  => Ok(()),
 _ => Err(Error::RmiErrorInput)

bokdeuk-jeong: Oh it's src_pa not target_pa. Sorry about the confusion.

bokdeuk-jeong (approved):

PR #184, opened 2023-09-15
rmm: correct cache settings

bokdeuk-jeong:
1. For RMM S1 page table
 - It's important to set attribute index porpery for both rmm memory and uart
 - Re-enable cacheability in SCTLR_EL2

2. For Realm S2 page table
- change cachability properties to NORMAL_FWB

Note: 
The purpose of this PR is only correcting the cache properties in page table descriptors.
there is redundancy between stage2_tte.rs and translation_granule_4k.rs. This is out of scope of this PR.


zpzigi754 (approved):
Good to fix the cache issue!

jinbpark (approved):

bitboom (approved):

PR #185, opened 2023-09-21
acs: Pass 35/36 rec creation cases 

bitboom:
This PR passes 35/36 in REC_CREATE by
-  Validate `rmi arguments` and `rec params` in REC_CREATE
-  Add RmiRecMpidr
- Add rec_index to RD

## Failed case
Failed case is related with `B3.4.11 RmiRecCreateFlags`.
The reserved fields of `RmiRecCreateFlags` must be zero.
But when i patched it like spec, it makes failures of other cases. 
tf-rmm is also failed in this case.
I need more analysis of this case.

### DIFF
--- a/rmm/src/granule/mod.rs
+++ b/rmm/src/granule/mod.rs
@@ -6,6 +6,7 @@ use self::translation::{
     GranuleStatusTable, GRANULE_STATUS_TABLE, L0_TABLE_ENTRY_SIZE_RANGE, L1_TABLE_ENTRY_SIZE_RANGE,
 };
 use crate::rmi::error::Error as RmiError;
+
 use vmsa::address::PhysAddr;
 use vmsa::error::Error;
 use vmsa::page_table::{HasSubtable, Level};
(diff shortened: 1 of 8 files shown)


rmm/src/granule/mod.rs
```diff
@@ -242,6 +243,13 @@ pub fn to_rmi_result(res: Result<(), Error>) -> Result<(), RmiError> {
     }
 }
 
+pub fn is_in_ns(addr: usize) -> bool {
+    match get_granule_if!(addr, GranuleState::Undelegated) {
```
bokdeuk-jeong, line 247: Isn't "is not in realm'?
What if provided physical address is in secure  or root security state? RMM cannot tell which state it is in other than the realm security state.

bitboom: It is more correct function name. I reflected it.

bokdeuk-jeong (approved):
It's nice that islet passed all the test cases for rec creation!

nook1208 (approved):
It looks good to me :)

hihi-wang:
nice to increase pass rate!

Issue #188, opened 2023-09-22
ACS: mm_unprotected_ipa_boundary testcase is failed


nook1208:
`mm_unprotected_ipa_boundary` testcase is failed because of Data abort in realm. 
let me check the issue in this page

```bash
// eom: mm_unprotected_ipa_boundary start
[TRACE]islet_rmm::event -- let's get STATS.lock() with cmd RTT_INIT_RIPAS
[TRACE]islet_rmm::stat -- start Stat with command RTT_INIT_RIPAS
[TRACE]islet_rmm::rmi -- Dummy implementation.
[INFO]islet_rmm::stat -- RMI::RTT_INIT_RIPAS            TOTAL MemUsed:            0 byte, AVG MemUsed:          0 byte, CallCnt:     6418
[TRACE]islet_rmm::event -- RMI: RTT_INIT_RIPAS       [88405000, 7FFFFFF000, 3] > [0, 0]
[TRACE]islet_rmm::event -- let's get STATS.lock() with cmd GRANULE_DELEGATE
[TRACE]islet_rmm::stat -- start Stat with command GRANULE_DELEGATE
[INFO]islet_rmm::stat -- RMI::GRANULE_DELEGATE          TOTAL MemUsed:       208944 byte, AVG MemUsed:         30 byte, CallCnt:     6909
[TRACE]islet_rmm::event -- RMI: GRANULE_DELEGATE     [8844C000] > [0]
[TRACE]islet_rmm::event -- let's get STATS.lock() with cmd DATA_CREATE
[TRACE]islet_rmm::stat -- start Stat with command DATA_CREATE
[INFO]islet_rmm::stat -- RMI::DATA_CREATE               TOTAL MemUsed:       147456 byte, AVG MemUsed:         22 byte, CallCnt:     6417
[TRACE]islet_rmm::event -- RMI: DATA_CREATE          [8844C000, 88405000, 7FFFFFF000, 8844B000, 0] > [0]
[TRACE]islet_rmm::event -- let's get STATS.lock() with cmd REALM_ACTIVATE
[TRACE]islet_rmm::stat -- start Stat with command REALM_ACTIVATE
[INFO]islet_rmm::stat -- RMI::REALM_ACTIVATE            TOTAL MemUsed:            0 byte, AVG MemUsed:          0 byte, CallCnt:       25
[TRACE]islet_rmm::event -- RMI: REALM_ACTIVATE       [88405000] > [0]
[...]

PR #190, opened 2023-09-25
rmm: Add an owner field to REC 

bitboom:
This PR is related with ACS (REC_ENTER).
REC holds `owner` according to Spec **A2.2.2 Granule ownership and A2.3.2 REC attributes**.
>  A REC has an “owner” attribute which points to the RD of the owning Realm.
>  A REC is not mapped at a Protected IPA. Its ownership therefore needs to be recorded explicitly.

The next testcase is related with https://github.com/Samsung/islet/issues/188.
So, i will continue `REC_ENTER` after fixing that.


zpzigi754 (approved):
I also think that many fields should be added in `REC` including `owner`.

rmm/src/rmi/rec/mod.rs
```diff
@@ -13,6 +13,8 @@ use vmsa::guard::Content;
 
 pub struct Rec {
     pub rd: Rd,
```
zpzigi754, line 15: If we use `owner` field, this `rd` field could be duplicate with `owner` (previously we have used this `rd` field for the usage of `owner`). Maybe we can remove this field later.

bokdeuk-jeong (approved):

PR #193, opened 2023-10-04
Change the way stage 2 page table is managed

zpzigi754:
This PR is the implementation of [the previous discussion](https://github.com/Samsung/islet/discussions/180). The stage 2 page table will be constructed in the location and timing where the Host has intended with the APIs.

## Page table difference handling ## 
The host can determine the type of the stage 2 table. For example, realm-linux's stage 2 page table walking starts from level 2 instead of level 0 while using 8 concatenated tables. This PR handles the table's difference with `realm-linux` feature and configurable `VTCR` (Virtualization Translation Control Register). A refactoring work would be needed to remove the feature in the future.

## Result summary ##
Regarding `tf-a-tests`, 
This PR will pass all rmi-related tests (We don't conduct rsi checks yet). 

Regarding `ACS tests`, this will 
pass 2 additional checks in `rtt_read_entry`
pass 13 additional checks in `rtt_init_ripas`
pass 16 additional checks in `rtt_create`
pass 4 additional checks in `cmd_data_create`
miss 1 additional checks in `realm_activate`
pass all checks in `mm_rtt_level_start`


rmm/src/realm/registry.rs
```diff
@@ -503,35 +589,85 @@ impl crate::rmi::Interface for RMI {
...
-            | bits_in_reg(S2TTE::SH, pte::shareable::INNER)
-            | bits_in_reg(S2TTE::DESC_TYPE, desc_type::L3_PAGE);
-        let s2tte = S2TTE::from(new_s2tte as usize);
-        let pa = s2tte.get_masked(S2TTE::ADDR_FULL);
-        let flags = s2tte.get_masked(S2TTE::PAGE_FLAGS);
+        if level == RTT_PAGE_LEVEL {
+            new_s2tte |= bits_in_reg(S2TTE::NS, 1)
+                | bits_in_reg(S2TTE::XN, 1)
+                | bits_in_reg(S2TTE::AF, 1)
+                | bits_in_reg(S2TTE::DESC_TYPE, desc_type::L3_PAGE);
```
bokdeuk-jeong, line 625: What's the reason of removing bits_in_reg(S2TTE::MEMATTR, pte::attribute::NORMAL_FWB) and bits_in_reg(S2TTE::SH, pte::shareable::INNER)?
Without this, doesn't it make the access non-cacheable and non-sharable?

rmm/src/realm/mm/stage2_tte.rs
```diff
@@ -20,10 +25,44 @@ pub mod desc_type {
...
+    pub const NS: u64 = 0x0;
+    pub const RESERVED: u64 = 0x1;
+    pub const OS: u64 = 0x2;
+    pub const IS: u64 = 0x3;
+}
+
+pub mod ap {
+    pub const RW: u64 = 0x3;
+}
+
```
bokdeuk-jeong, line 38: I cannot find where the struct sh and struct ap are used.

rmm/src/realm/mm/stage2_translation.rs
```diff
@@ -28,31 +30,59 @@ pub mod tlbi_ns {
...
 }
 
 impl<'a> Stage2Translation<'a> {
-    pub fn new() -> Self {
+    #[cfg(feature = "realm_linux")]
+    pub fn new(rtt_base: usize) -> Self {
+        let root_pgtlb = unsafe {
+            &mut *PageTable::<
+                GuestPhysAddr,
+                L2Table,
```
bokdeuk-jeong, line 62: I don't understand why the root has to start with L2. Could you point the source code location in nw-linux where it requires L2 root ?

rmm/src/realm/mm/stage2_tte.rs
bokdeuk-jeong: You know that we already have defined them in realm/mm/page_table/pte.rs?
To me, the existing implementation style looks better.

rmm/src/realm/registry.rs
bokdeuk-jeong: I found that host_s2tte, which is one of prameters for RMI_RTT_MAP_UNPROTECTED, already includes the properties. So it eventually inculde chachability(FWB), sharability(inner ), access permission(RW)
https://github.com/Samsung/islet-asset/blob/604bc2c405be75be6645ddf1dcfeeca8bbd86b9f/arch/arm64/kvm/rme.c#L646

rmm/src/realm/mm/stage2_translation.rs
bokdeuk-jeong: The reason looks very much acceptable!

rmm/src/realm/registry.rs
zpzigi754: Thank you for the investigation!

rmm/src/realm/mm/stage2_tte.rs
zpzigi754: I see. I will change to use the existing one instead.

zpzigi754: I've used ones in realm/mm/page_table/pte.rs like the below.
```
new_s2tte |= bits_in_reg(S2TTE::AP, permission::RW);
new_s2tte |= bits_in_reg(S2TTE::SH, shareable::INNER);
```

bokdeuk-jeong (approved):
nice work!

PR #197, opened 2023-10-06
Patch for the setup of realm in ACS

nook1208:
This PR is related with #188 issue.
With this PR, we can see the realm is boot successfully in ACS test.

### DIFF
--- a/rmm/src/realm/registry.rs
+++ b/rmm/src/realm/registry.rs
@@ -57,7 +57,7 @@ fn exit() {
     }
 }
 
-fn get_realm(id: usize) -> Option<RealmMutex> {
+pub fn get_realm(id: usize) -> Option<RealmMutex> {
     RMS.lock().1.get(&id).map(|realm| Arc::clone(realm))
 }
 
--- a/rmm/src/realm/registry.rs
+++ b/rmm/src/realm/registry.rs
@@ -300,7 +300,7 @@ impl crate::rmi::Interface for RMI {
                 3 => 0xffffffff_ffffffff, // double word
                 _ => unreachable!(),      // SAS consists of two bits
             };
-            let val = unsafe { run.entry_gpr0() } & mask;
+            let val = unsafe { run.entry_gpr(0)? } & mask;
             let sign_extended = esr.get_masked_value(EsrEl2::SSE);
             if sign_extended != 0 {
                 // TODO
--- a/rmm/src/realm/registry.rs
+++ b/rmm/src/realm/registry.rs
@@ -736,9 +736,8 @@ impl crate::rmi::Interface for RMI {
         if s2tte.is_valid(level, false) {
             // This condition is added with no-op for handling the `else` case
         } else if s2tte.is_unassigned() || s2tte.is_assigned() {
-            let pa: usize = s2tte.address(level).ok_or(Error::RmiErrorRtt(0))?.into(); //XXX: check this again
             let flags = bits_in_reg(S2TTE::INVALID_RIPAS, invalid_ripas::RAM);
-            let new_s2tte = pa as u64 | flags;
+            let new_s2tte = s2tte.get() | flags;
 
             get_realm(id)
                 .ok_or(Error::RmiErrorOthers(NotExistRealm))?
--- a/rmm/src/realm/registry.rs
+++ b/rmm/src/realm/registry.rs
@@ -377,7 +377,7 @@ impl crate::rmi::Interface for RMI {
         Ok(())
     }
 
-    fn realm_config(&self, id: usize, config_ipa: usize) -> Result<(), Error> {
+    fn realm_config(&self, id: usize, config_ipa: usize, ipa_bits: usize) -> Result<(), Error> {
         let res = get_realm(id)
             .ok_or(Error::RmiErrorOthers(NotExistRealm))?
             .lock()
@@ -386,9 +386,7 @@ impl crate::rmi::Interface for RMI {
             .ipa_to_pa(GuestPhysAddr::from(config_ipa), 3);
         if let Some(pa) = res {
             let pa: usize = pa.into();
-            // TODO: Receive ipa width (== 33) from the Host's argument
-            //       and store it inside REC as well
-            unsafe { RealmConfig::init(pa, 33) };
+            unsafe { RealmConfig::init(pa, ipa_bits) };
             Ok(())
         } else {
             Err(Error::RmiErrorInput)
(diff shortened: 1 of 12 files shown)


rmm/src/rmi/rec/handlers.rs
```diff
@@ -48,7 +48,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         match rmi.create_vcpu(rd.id()) {
             Ok(vcpuid) => {
                 ret[1] = vcpuid;
-                rec.init(owner, rd.id(), rd.state(), vcpuid);
+                rec.init(owner, *rd, vcpuid);
```
bitboom, line 51: Below 68 `rd.inc_rec_index()` does not work properly because rd was already copied. I think we have to avoid copying whole `RD`.
Instead, we can access `rd members` in `rec` with `owner` field.  (owner is rd's address)

```rust
let rd = get_granule_if!(rec.owner(), GranuleState::RD)?;
let rd = rd.content::<Rd>();
// rd.id(), rd.state()
```

rmm/src/realm/registry.rs
```diff
@@ -736,9 +736,8 @@ impl crate::rmi::Interface for RMI {
         if s2tte.is_valid(level, false) {
             // This condition is added with no-op for handling the `else` case
         } else if s2tte.is_unassigned() || s2tte.is_assigned() {
-            let pa: usize = s2tte.address(level).ok_or(Error::RmiErrorRtt(0))?.into(); //XXX: check this again
             let flags = bits_in_reg(S2TTE::INVALID_RIPAS, invalid_ripas::RAM);
-            let new_s2tte = pa as u64 | flags;
+            let new_s2tte = s2tte.get() | flags;
```
zpzigi754, line 740: Uoops, it's my fault. Thank you for finding and fixing it!

rmm/src/rmi/rec/handlers.rs
nook1208: thanks, Fixed!

rmm/src/rsi/mod.rs
```diff
@@ -62,10 +91,10 @@ pub fn set_event_handler(rsi: &mut RsiHandle) {
 
     listen!(rsi, REALM_CONFIG, |_arg, ret, rmm, rec, _| {
         let rmi = rmm.rmi;
-        let realmid = rec.rd.id();
+        let realmid = rec.realm_id()?;
         let vcpuid = rec.id();
         let config_ipa = rmi.get_reg(realmid, vcpuid, 1)?;
-        rmi.realm_config(realmid, config_ipa)?;
+        rmi.realm_config(realmid, config_ipa, rec.ipa_bits()?)?;
```
bokdeuk-jeong, line 97: Rec.ipa_bits() should inquire Rd to get this. This makes another wrapper.
I think Rd should directly expose the ipa_bits() method because Rd holds the information.
Why don't we get it as following?
    let rd = rec.rd()
    rm.realm_config(rd.id(), config_ipa, rd.ipa_bits());

rmm/src/rsi/mod.rs
```diff
@@ -33,9 +34,17 @@ pub fn set_event_handler(rsi: &mut RsiHandle) {
...
         let vcpuid = rec.id();
         let ipa = rmi.get_reg(realmid, vcpuid, 1).unwrap_or(0x0);
-        let pa: usize = ipa;
+
+        let pa = crate::realm::registry::get_realm(realmid)
+            .ok_or(Error::RmiErrorOthers(NotExistRealm))?
+            .lock()
+            .page_table
+            .lock()
+            .ipa_to_pa(crate::realm::mm::address::GuestPhysAddr::from(ipa), 3)
```
zpzigi754: [minor]
How about using const value instead of 3 like the below?
```
.ipa_to_pa(crate::realm::mm::address::GuestPhysAddr::from(ipa), RTT_PAGE_LEVEL)
```

rmm/src/rmi/rec/handlers.rs
```diff
@@ -102,16 +103,10 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
-            // TODO: copy rec::entry gprs to host_call gprs
-            let ipa: u64 = 0x800088e00000;
-            if run.entry_gpr0() == ipa {
-                // TODO: Get ipa from rec->regs[1] and map to pa
-                let pa: usize = 0x88b0_6000;
-                let host_call = rsi::hostcall::HostCall::parse_mut(pa);
-                host_call.set_gpr0(ipa);
-            }
+        if rec.host_call_pending() {
+            do_host_call(&arg, ret, rmm, rec, &mut run)?;
```
zpzigi754, line 107: Good to remove the hardcoded values!

rmm/src/rsi/mod.rs
```diff
@@ -28,31 +31,48 @@ pub const VERSION: usize = (1 << 16) | 0;
...
-            .page_table
-            .lock()
-            .ipa_to_pa(crate::realm::mm::address::GuestPhysAddr::from(ipa), 3)
-            .ok_or(Error::RmiErrorInput)?;
+    let pa = crate::realm::registry::get_realm(realmid)
+        .ok_or(Error::RmiErrorOthers(NotExistRealm))?
+        .lock()
+        .page_table
+        .lock()
+        .ipa_to_pa(crate::realm::mm::address::GuestPhysAddr::from(ipa), 3)
```
zpzigi754, line 51: [minor]
How about using the const value instead of 3 (the same comment as before)?

rmm/src/rsi/hostcall.rs
```diff
@@ -14,8 +15,13 @@ impl HostCall {
...
 
-    pub unsafe fn set_gpr0(&mut self, val: u64) {
-        (*self.inner.val).gprs[0] = val;
+    pub unsafe fn set_gpr(&mut self, idx: usize, val: u64) -> Result<(), Error> {
+        if idx >= HOST_CALL_NR_GPRS {
+            error!("out of index: {}", idx);
+            return Err(Error::RmiErrorInput);
+        }
+        (*self.inner.val).gprs[idx] = val;
+        Ok(())
```
zpzigi754, line 24: How about modifying `set_gpr0()` in `./src/rmi/rec/run.rs` altogether?

zpzigi754 (approved):
Thank you for debugging and fixing many issues!

bokdeuk-jeong (approved):

rmm/src/rsi/mod.rs
nook1208: Thank you for the review! I'll make a new PR to apply your comments :)

rmm/src/rsi/mod.rs
nook1208: I tried to make a rec.rd() after getting the granule of the rd in rec.rd().
But in rust, it's impossible to return a reference to a variable created in a function.
So maybe I can write like this : 
```rust
let rd = get_granule_if!(rec.owner(), GranuleState::RD)?;
rm.realm_config(rd.id(), config_ipa, rd.ipa_bits());
```
l'll make a new PR to apply your comment. Thank you :)

PR #200, opened 2023-10-11
Keep both tf-a and tf-a-rss

bitboom:
ACS uses uart2 for reporting. TF-A-RSS(TF-A patched for RSS) also seems to use uart2.
So. it's hard to use unified tf-a without modifying.

To move forward, I think it's better to keep both tf-a & tf-a-rss before resolving the issue.

## To use HES
`--hes` option is needed.

```sh
./scripts/fvp-cca -nw=tf-a-tests -rmm=tf-rmm --hes 
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,17 @@ jobs:
       - name: Test ACS
         run: ./scripts/tests/acs.sh
 
+  hes:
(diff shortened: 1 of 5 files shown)


zpzigi754 (approved):

zpzigi754:
I also agree with separately maintaining `tf-a` and `tf-a-rss`!

PR #202, opened 2023-10-12
rmm: Do not drop the stage 2 table as RAII

bitboom:
Stage 2 table should be drop by RTT_DESTROY


zpzigi754:
Good to see that it pass two additional tests :)
Rebasing the main branch will pass the acs test!

zpzigi754 (approved):

PR #204, opened 2023-10-12
Add measurements

L0czek:
This PR add the ability to take and read measurements of realms. To accomplish this, I added appropriate fields inside the Realm and rd objects and modified several RMI handler to measure the passed parameters. Additionally, I add 2 missing RSI call allowing for reading and extending the measurements, basically reading RIM and extending REMs. This yields measurements that are binary compatible with beta0 version of tf-rmm provided in islet-assets repo.

### DIFF
--- a/rmm/src/measurement/ctx.rs
+++ b/rmm/src/measurement/ctx.rs
@@ -1,0 +1,114 @@
+use super::{
+    Hasher, Measurement, MeasurementError, MEASUREMENTS_SLOT_RIM, MEASURE_DESC_TYPE_DATA,
+    MEASURE_DESC_TYPE_REC, MEASURE_DESC_TYPE_RIPAS, RMI_MEASURE_CONTENT,
+};
+use crate::rmi::rec::params::Params as RecParams;
+use crate::{
+    event::RsiHandle,
+    host::DataPage,
+    rmi::realm::{params::Params as RealmParams, Rd},
+    rsi::{self, Interface},
+};
+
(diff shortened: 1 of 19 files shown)


zpzigi754:
Good to see the building blocks for the attestation!

Running `cargo fmt` would help to resolve the CI failure (rule-checker).

rmm/src/measurement/ctx.rs
```diff
@@ -0,0 +1,96 @@
+use crate::{event::RsiHandle, rmi::realm::{params::Params as RealmParams, Rd}, rsi::{Interface, self}, host::DataPage};
+use crate::rmi::rec::params::Params as RecParams;
+use super::{Hasher, MeasurementError, MEASUREMENTS_SLOT_RIM, Measurement, RMI_MEASURE_CONTENT, MEASURE_DESC_TYPE_DATA, MEASURE_DESC_TYPE_REC, MEASURE_DESC_TYPE_RIPAS};
+
+pub struct HashContext<'a> {
+    hasher: Hasher,
+    rsi: &'a RsiHandle,
+    realmid: usize
```
bokdeuk-jeong, line 8: Storing realmid sperately is not desirable, Could you change this to have a reference to Rd and call rd.id() when it needs to access realmid?

bokdeuk-jeong (approved):
To avoid the delay due to the different time zone, I am approving this in advance.
I would like to leave the final confirmation to Piotr.

PR #208, opened 2023-10-17
Add RSI::IPA_STATE_GET command

nook1208:
Some ACS testcases are failed because the islet don't have RSI::IPA_STATE_GET command.
This PR makes it pass on below testcases :

```bash
cmd_ipa_state_get
mm_ripas_change
mm_ripas_change_reject
mm_ripas_change_partial
```

### DIFF
--- a/rmm/src/event/rsihandle.rs
+++ b/rmm/src/event/rsihandle.rs
@@ -59,7 +59,8 @@ impl RsiHandle {
                 };
 
                 let realm_id = g_rd.content::<Rd>().id();
-                drop(g_rd);
+                drop(g_rd); // manually drop to reduce a lock contention
+
                 // TODO: handle the error properly
                 let _ = rmi.set_reg(realm_id, rec.id(), 0, RsiHandle::NOT_SUPPORTED);
                 error!(
--- a/rmm/src/event/rsihandle.rs
+++ b/rmm/src/event/rsihandle.rs
@@ -8,7 +8,10 @@ use crate::rsi;
 use crate::rsi::psci;
 use crate::Monitor;
 // TODO: Change this into rsi::error::Error
+use crate::granule::GranuleState;
 use crate::rmi::error::Error;
+use crate::rmi::realm::Rd;
+use crate::{get_granule, get_granule_if};
 
 use alloc::boxed::Box;
 use alloc::collections::btree_map::BTreeMap;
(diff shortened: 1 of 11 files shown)


rmm/src/event/rsihandle.rs
```diff
@@ -46,13 +49,19 @@ impl RsiHandle {
...
+                let g_rd = match res {
+                    Ok(g_rd) => g_rd,
+                    Err(e) => {
+                        error!("failed to get rd: {:?}", e);
+                        return RsiHandle::RET_FAIL;
+                    }
+                };
+
+                let realm_id = g_rd.content::<Rd>().id();
+                drop(g_rd);
```
zpzigi754, line 62: Is manual drop necessary?
(the same comment can be applied to other places too)

rmm/src/rsi/constraint.rs
```diff
@@ -17,8 +17,8 @@ lazy_static! {
         m.insert(rsi::ABI_VERSION, Constraint::new(rsi::ABI_VERSION, 2, 1));
         m.insert(rsi::REALM_CONFIG, Constraint::new(rsi::REALM_CONFIG, 2, 1));
         m.insert(
-            rsi::IPA_STATE_SET,
-            Constraint::new(rsi::IPA_STATE_SET, 2, 1),
+            rsi::IPA_STATE_GET,
+            Constraint::new(rsi::IPA_STATE_GET, 2, 1),
```
zpzigi754, line 21: Good to find and fix duplicate definition!

rmm/src/rmi/rec/handlers.rs
```diff
@@ -93,6 +93,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // grab the lock for Rec
         let mut rec_granule = get_granule_if!(arg[0], GranuleState::Rec)?;
         let mut rec = rec_granule.content_mut::<Rec>();
+        let realm_id;
```
hihi-wang, line 96: May better to declare var like

~~~
let realm_id = {
    ...
    re.id()
}
~~~

nook1208: To do like that, then we should get the granule of rd again.
because it's already in 103,104 lines. 
so I thought that getting the granule and lock again was not good.

If I misunderstand your intention, please let me know :)

rmm/src/event/rsihandle.rs
nook1208: There are two reasons.
The first reason is that trying to reduce a lock contention is necessary.
The second reason is that realm_id and ipa_bits won't changed before the realm is destroyed.
that's why I manually used 'drop'.

zpzigi754 (approved):

rmm/src/event/rsihandle.rs
bokdeuk-jeong: In that case, how about adding an inline comment about why drop() should be called manually? Others will also wonder why this is necessary and might try to remove this.

nook1208: Fixed, Thank you !

bokdeuk-jeong (approved):

rmm/src/rmi/rec/handlers.rs
hihi-wang: What I meant is like this : 

~~~
        let realm_id = 
        {
            let rd = get_granule_if!(rec.owner(), GranuleState::RD)?;
            let rd = rd.content::<Rd>();
            match rd.state() {
                State::Active => {}
                State::New => {
                    return Err(Error::RmiErrorRealm(0));
                }
                State::SystemOff => {
                    return Err(Error::RmiErrorRealm(1));
                }
                _ => {
                    panic!("Unexpected realm state");
                }
            }
            rd.id()
        };
~~~

I think it's bit awkward to use uninitialized declaration in Rust.

nook1208: Oh, I got your point. 
I fixed. thank you !

PR #210, opened 2023-10-18
Add rsi failure contions

nook1208:
Add some failure conditions on RSI_REALM_CONFIG, RSI_IPA_STATE_SET
It can pass the below test cases:
```bash

Suite=command : cmd_realm_config
REALM : 
	Check 1 : addr_align; intent id : 0x0 
	Check 2 : addr_bound; intent id : 0x1 
Result => Passed

Suite=command : cmd_ipa_state_set
REALM : 
	Check 1 : addr_align; intent id : 0x0 
	Check 2 : size_bound; intent id : 0x1 
	Check 3 : rgn_bound; intent id : 0x2 
	Check 4 : ripas_valid; intent id : 0x3 
Result => Passed
```

### DIFF
--- a/rmm/src/rsi/mod.rs
+++ b/rmm/src/rsi/mod.rs
@@ -195,6 +195,12 @@ pub fn set_event_handler(rsi: &mut RsiHandle) {
         drop(g_rd); // manually drop to reduce a lock contention
 
         let config_ipa = rmi.get_reg(realmid, vcpuid, 1)?;
+        if validate_ipa(config_ipa, ipa_bits).is_err() {
+            rmi.set_reg(realmid, vcpuid, 0, ERROR_INPUT)?;
+            ret[0] = rmi::SUCCESS_REC_ENTER;
+            return Ok(());
+        }
+
         rmi.realm_config(realmid, config_ipa, ipa_bits)?;
 
         if rmi.set_reg(realmid, vcpuid, 0, SUCCESS).is_err() {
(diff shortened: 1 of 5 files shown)


bokdeuk-jeong (approved):

rmm/src/rsi/mod.rs
```diff
@@ -257,17 +264,41 @@ pub fn set_event_handler(rsi: &mut RsiHandle) {
...
+        let ipa_start = rmi.get_reg(realmid, vcpuid, 1)?;
+        let ipa_size = rmi.get_reg(realmid, vcpuid, 2)?;
         let ipa_state = rmi.get_reg(realmid, vcpuid, 3)? as u8;
+        let ipa_end = ipa_start + ipa_size;
+
+        if ipa_end <= ipa_start {
+            return Err(Error::RmiErrorInput); // integer overflows or size is zero
+        }
+
+        if !is_granule_aligned(ipa_start)
```
jinbpark, line 281: This comment is not limited to this part, but rather, to broader parts that have to do with simple input validation codes.
I think such input validations (i.e., align check, size check, type check (ipa_state)) can be done in a unified place.

We already have a unified place to do this job, "rmm/src/rsi/constraint.rs".  (see [this page](https://samsung.github.io/islet/platform-development/secure-interaction.html) for more detail on RMI/RSI validation)
Currently, what "constraint.rs" does is only enforce the number of parameters used in RMI/RSI commands.
But we can extend it to deal with these command-specific yet simple input validation codes.
This is what I left "TODO comment" on, in "constraint.rs - validate()".

But this kind of refactoring task is not necessary I think, as there are not many checks at this point.

jinbpark (approved):

hihi-wang (approved):

PR #213, opened 2023-10-23
doc: Add an initial SDK design document

bitboom:
This PR adds an initial SDK design document.
Please refer [here](https://github.com/Samsung/islet/blob/sdk-sdd/doc/sdk-sdd.md) to see the document.


bitboom:
> Nice documentation-: particularly I like all the diagrams used here. Is this going to get into our github page, right?

Absolutely! 

> * One suggestion: it would be nice to note what part has not yet been done. (maybe, "process delegated attestation" is one thing that needs to get done later, right?)

I noted it like below. Thanks for the suggestion.

---

*NOTE: Delegated attestation module of Islet RMM works in progress. SDK uses [the reference RMM](https://www.trustedfirmware.org/projects/tf-rmm/) now, which means SDK is loosely coupled with RMM.*

jinbpark (approved):

hihi-wang (approved):

PR #215, opened 2023-10-24
Handle realm exit data abort

nook1208:
It handles an exception for a data abort from the realmexit
according to 'A4.3.4.3 REC exit due to Data Abort'.

it constructs 'Exit' fields differently especially
for the 'exit.esr' register based by the below two types of data abort:
1. Emulatable Data Abort
2. Non-Emulatable Data Abort

It can pass below testcases :
```bash
exception_emulatable_da
exception_non_emulatable_da_1
exception_non_emulatable_da_2
mm_hipas_destroyed_ripas_ram_da
mm_protected_ipa_boundary

TOTAL TESTS     : 51
TOTAL PASSED    : 35
```

### DIFF
--- a/rmm/src/event/realmexit.rs
+++ b/rmm/src/event/realmexit.rs
@@ -13,33 +13,79 @@ use crate::{get_granule, get_granule_if};
 use crate::{rmi, rsi};
 use armv9a::{EsrEl2, EMULATABLE_ABORT_MASK, HPFAR_EL2, NON_EMULATABLE_ABORT_MASK};
 
+const RMI_REC_EXIT_REASON_MASK: usize = 15; // 0b1111
+
+const EXIT_SYNC_TYPE_SHIFT: usize = 4;
+const EXIT_SYNC_TYPE_MASK: usize = 15 << EXIT_SYNC_TYPE_SHIFT; // 0b1111_0000
+
 #[derive(Debug)]
-pub enum Kind {
-    RSI = 0,
+#[repr(usize)]
+pub enum RmiRecExitReason {
+    Sync(ExitSyncType),
     IRQ = 1,
     FIQ = 2,
-    SERROR = 3,
-    InstAbort = 4,
-    DataAbort = 5,
-    UndefinedSync,
+    PSCI = 3,
+    RipasChange = 4,
+    HostCall = 5,
+    SError = 6,
+    Undefined = RMI_REC_EXIT_REASON_MASK, // fixed, 0b1111
 }
 
-impl Into<u64> for Kind {
+impl Into<u64> for RmiRecExitReason {
+    fn into(self) -> u64 {
+        match self {
+            RmiRecExitReason::Sync(exit_sync_type) => exit_sync_type.into(),
+            RmiRecExitReason::IRQ => 1,
+            RmiRecExitReason::FIQ => 2,
+            RmiRecExitReason::PSCI => 3,
+            RmiRecExitReason::RipasChange => 4,
+            RmiRecExitReason::HostCall => 5,
+            RmiRecExitReason::SError => 6,
+            RmiRecExitReason::Undefined => 7,
+        }
+    }
+}
+
+impl From<usize> for RmiRecExitReason {
+    fn from(num: usize) -> Self {
+        match num & RMI_REC_EXIT_REASON_MASK {
+            0 => RmiRecExitReason::Sync(ExitSyncType::from(num)),
+            1 => RmiRecExitReason::IRQ,
+            2 => RmiRecExitReason::FIQ,
+            3 => RmiRecExitReason::PSCI,
+            4 => RmiRecExitReason::RipasChange,
+            5 => RmiRecExitReason::HostCall,
+            6 => RmiRecExitReason::SError,
+            _ => RmiRecExitReason::Undefined,
+        }
+    }
+}
+
+/// Fault Status Code only for the 'RmiRecExitReason::Sync'
+/// it's a different from trap::Syndrome
+#[derive(Debug)]
+#[repr(usize)]
+pub enum ExitSyncType {
+    RSI = 1 << EXIT_SYNC_TYPE_SHIFT,
+    DataAbort = 2 << EXIT_SYNC_TYPE_SHIFT,
+    InstAbort = 3 << EXIT_SYNC_TYPE_SHIFT,
+    Undefined = EXIT_SYNC_TYPE_MASK, // fixed, 0b1111_0000
+}
+
+impl Into<u64> for ExitSyncType {
     fn into(self) -> u64 {
         self as u64
     }
 }
 
-impl From<usize> for Kind {
+impl From<usize> for ExitSyncType {
     fn from(num: usize) -> Self {
-        match num {
-            0 => Kind::RSI,
-            1 => Kind::IRQ,
-            2 => Kind::FIQ,
-            3 => Kind::SERROR,
-            4 => Kind::InstAbort,
-            5 => Kind::DataAbort,
-            _ => Kind::UndefinedSync,
+        let masked_val = (num & EXIT_SYNC_TYPE_MASK) >> EXIT_SYNC_TYPE_SHIFT;
+        match masked_val {
+            1 => ExitSyncType::RSI,
+            2 => ExitSyncType::DataAbort,
+            3 => ExitSyncType::InstAbort,
+            _ => ExitSyncType::Undefined,
         }
     }
 }
@@ -51,8 +97,8 @@ pub fn handle_realm_exit(
     run: &mut Run,
 ) -> Result<(bool, usize), Error> {
     let mut return_to_ns = true;
-    let ret = match Kind::from(realm_exit_res[0]) {
-        Kind::RSI => {
+    let ret = match RmiRecExitReason::from(realm_exit_res[0]) {
+        RmiRecExitReason::Sync(ExitSyncType::RSI) => {
             trace!("REC_ENTER ret: {:#X?}", realm_exit_res);
             let rsi = &rmm.rsi;
             let cmd = realm_exit_res[1];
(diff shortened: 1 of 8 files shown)


rmm/src/event/realmexit.rs
```diff
@@ -1,5 +1,158 @@
...
+    let esr_el2 = EsrEl2::new(esr_el2);
+    let rt = esr_el2.get_masked_value(EsrEl2::SRT) as usize;
+    let write_val = match rt == 31 {
+        true => 0, // xzr
+        false => rmi.get_reg(realm_id, vcpu_id, rt)? as u64 & esr_el2.get_access_size_mask(),
+    };
+    Ok(write_val)
+}
+
+fn handle_sync_exception(
```
bokdeuk-jeong, line 98: We have a  directory reserved for handling exceptions taken from lower exception levels at rmm/src/exception/lower.
Why don't you first handle dabt exceptions itself at rmm/src/exception/lower/synchronous and then let the rmm, whose last context was at entering a realm, resume its previous context and handle the realm exit?

rmm/src/event/realmexit.rs
```diff
@@ -1,5 +1,158 @@
...
 pub const IRQ: usize = 1;
 pub const FIQ: usize = 2;
 pub const SERROR: usize = 3;
 pub const SYNC: usize = 4;
+const NON_EMULATABLE_ABORT_MASK: u64 =
+    EsrEl2::EC | EsrEl2::SET | EsrEl2::FNV | EsrEl2::EA | EsrEl2::DFSC;
+const EMULATABLE_ABORT_MASK: u64 =
+    NON_EMULATABLE_ABORT_MASK | EsrEl2::ISV | EsrEl2::SAS | EsrEl2::SF | EsrEl2::WNR;
+
+pub fn handle_realm_exit(
```
bokdeuk-jeong, line 27: I think this code needs to move some other place. The purpose of this 'event' directory is to provide interface to register each RMI or RSI commands and invoke them instead of actually handling them.

rmm/src/event/realmexit.rs
nook1208: It exactly follows the flow that you mentioned :)
The `realmexit::handle_sync_exception()` will handle the data abort and set the esr field in `Exit` to pass it to the host with the retricted information based by the abort type (emulatable or not).

bokdeuk-jeong: Thanks for explaining what handle_sync_exception() does here and explaining it in person too.

rmm/src/event/realmexit.rs
```diff
@@ -1,5 +1,158 @@
...
+    let esr_el2 = realm_exit_res[1] as u64;
+    let hpfar_el2 = realm_exit_res[2] as u64;
+    let far_el2 = realm_exit_res[3] as u64;
+
+    unsafe {
+        run.set_exit_reason(rmi::EXIT_SYNC);
+        run.set_hpfar(hpfar_el2);
+    }
+
+    match Syndrome::from(esr_el2 as u32) {
```
bokdeuk-jeong, line 119: We already deliver exit reason in realm_exit_res[0]. Re-interperting syndrome from esr_els looks duplication of a same work. The exit reasons in realmexit.rs are intended to be shared between exception vector and the rest of RMM such as RMI handlers.
Why don't you add new definitions or update existing definitions in realmexit.rs?

nook1208: I think that re-interpreting is necessary after rec_enter is done. because the rmm should handle the abort based by ESR_EL2.
And I used 'Syndrome' in trap.rs to reuse it.
You means, define a new `Syndrome` struct in realmexit.rs to avoid confusion between trap.rs and realmexit.rs ?

I'm not sure I understand your intent correctly.

rmm/src/event/realmexit.rs
nook1208: Fixed, thank you !

rmm/src/event/realmexit.rs
nook1208: Fixed, it looks better!
Thank you :)

zpzigi754 (commented):
Good to pass 5 additional acs tests by distinguishing emulatable abrot from non-emulatable abort!

The last commit can be removed by fixing `GRANULE_MASK` into `!((1 << GRANULE_SHIFT) - 1)` or `!0xfff`. It would be able to successfully boot linux-realm.

rmm/src/event/realmexit.rs
```diff
@@ -5,11 +5,36 @@ use crate::rmi::rec::Rec;
...
+#[derive(Debug)]
+pub enum Kind {
+    RSI = 0,
+    IRQ = 1,
+    FIQ = 2,
+    SERROR = 3,
+    InstAbort = 4,
+    DataAbort = 5,
+    UndefinedSync,
+}
```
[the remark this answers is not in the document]
nook1208: Actually, it's the encoding of `RmiRecExitReason`.
`InstAbort`, `DataAbort` and `UndefinedSync` are part of `RMI_EXIT_SYNC`. 
I think those fields are needed to define to see what kinds of synchronous exception

rmm/src/realm/registry.rs
```diff
@@ -296,14 +296,7 @@ impl crate::rmi::Interface for RMI {
...
         if wnr == 0 && rt != 31 {
-            let sas = esr.get_masked_value(EsrEl2::SAS);
-            let mask: u64 = match sas {
-                0 => 0xff,                // byte
-                1 => 0xffff,              // half-word
-                2 => 0xffffffff,          // word
-                3 => 0xffffffff_ffffffff, // double word
-                _ => unreachable!(),      // SAS consists of two bits
-            };
+            let mask = esr.get_access_size_mask();
```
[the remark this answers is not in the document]
nook1208: You're right. `send_mmio_write()` is removed in cfa3e71 commit :)

rmm/src/event/realmexit.rs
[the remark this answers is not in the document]
nook1208: > The last commit can be removed by fixing GRANULE_MASK into !((1 << GRANULE_SHIFT) - 1) or !0xfff. It would be able to successfully boot linux-realm.
Isn't it the same as !(GRANULE_MASK as u64)??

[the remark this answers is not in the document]
zpzigi754: > Oh, Isn't it the same as `!(GRANULE_MASK as u64)`??

What I meant was 

[current]
```
pub const GRANULE_MASK: usize = (1 << GRANULE_SHIFT) - 1;
...
let exit_far =  (far_el2 & !(GRANULE_MASK as u64));  // this will mask the top 52 bits
```


[fix]
```
pub const GRANULE_MASK: usize = !((1 << GRANULE_SHIFT) - 1);
...
let exit_far =  (far_el2 & !(GRANULE_MASK as u64)); // this will mask the bottom 12 bits
```

With this change, https://github.com/Samsung/islet/pull/215/commits/18aa8f16a2286bf55fbdf290547d1a9bcad688f2 can be removed.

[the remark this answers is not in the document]
nook1208: Oh, That's a bug. let me fix it. 
Thank you !!

bitboom (approved):

rmm/src/event/realmexit.rs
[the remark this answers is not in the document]
nook1208: Fixed. Thank you :)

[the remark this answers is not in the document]
zpzigi754: > Actually, it's the encoding of `RmiRecExitReason`. `InstAbort`, `DataAbort` and `UndefinedSync` are part of `RMI_EXIT_SYNC`. There are so many synchronous exceptions. The synchronous exceptions are classified already in `trap.rs` and then passed to the rmm. so we don't need to classify it again in `realmexit.rs` That's the review given by @bokdeuk-jeong in a previous comment.
> 
> I think those fields are need to be defined to see what kinds of synchronous exception occured removing duplicated classification code.

Now I get your intention (not adding duplicate classification). As you mentioned, there are many different kinds of synchronous exception. I think that those do not need to be classified in trap.rs, but can be classified in realmexit while handling synchronous exception. 

I think that if `Kind` is our synonym type for `RmiRecExitReason`, the subcateogory of the exit reason should not be classified with the higher category. In that case, not only the entry but also the encoding value becomes different. 
Also, if InstAbort and DataAbort are added in `Kind`, there is no reason not to include other synchronous exception types such as `WFX`, `HVC`, and `SysReg`. 

If classifying the synchronous exception in trap.rs is unavoidable, maybe `Kind` can be defined as follows.
```
pub enum RmiRecExitReason {   // I personally prefer using `RmiRecExitReason` instead of `Kind`
    SYNC(SyncType) = 0,
    IRQ = 1,
    FIQ = 2,
    PSCI = 3,
    RIPAS_CHANGE = 4,
    HOST_CALL = 5,
    SERROR = 6,
}

pub enum SyncType {
    INST_ABORT,
    DATA_ABORT,
    ...   
}
```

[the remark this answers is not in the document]
nook1208: > If classifying the synchronous exception in trap.rs is unavoidable, maybe Kind can be defined as follows.

I think it's unavoidable. so it seems to look better!
Thank you for the example code. let me try to make the `RmiRecExitReason` enums type. :)

rmm/src/event/realmexit.rs
```diff
@@ -1,5 +1,171 @@
...
+    realm_exit_res: [usize; 4],
+    rmm: &Monitor,
+    rec: &mut Rec,
+    run: &mut Run,
+) -> Result<usize, Error> {
+    let g_rd = get_granule_if!(rec.owner(), GranuleState::RD)?;
+    let rd = g_rd.content::<Rd>();
+    let realm_id = rd.id();
+    let ipa_bits = rd.ipa_bits();
+    drop(g_rd); // manually drop to reduce a lock contention
```
jinbpark, line 135: Not related to this PR itself,

I saw this manual drop after it got merged. I think the right way to deal with this (i.e., to reduce a lock contention) is to make use of a read-write lock because the rationale behind the fact that this manual drop is safe is there will be many reads but no write.

We can benefit from read-write locks the same way we're getting from this manual drop. But, read-write locks seem a more correct and generic approach.
With read-write locks, we don't need to have additional drop() calls.

jinbpark (approved):

rmm/src/event/realmexit.rs
nook1208: Thank you for the comment !
As I know, the rwlock uses a bit more complicated atomic operation. 
so if the reader critical section is not that long, then spinlock is more recommended [in this reference](https://docs.kernel.org/locking/spinlocks.html#lesson-2-reader-writer-spinlocks)
> NOTE! reader-writer locks require more atomic memory operations than simple spinlocks. Unless the reader critical section is long, you are better off just using spinlocks.

I think in many cases, `realm_id` and `ipa_bits` are used.
So actually I want to make new fields for `realm_id` and `ipa_bits` in our Rec structure if we can confirm that they won't be changed in runtime after activating the realm.
then we can remove the manual `drop`.

rmm/src/event/realmexit.rs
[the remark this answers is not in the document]
nook1208: Fixed. Thank you !

zpzigi754 (approved):

rmm/src/event/realmexit.rs
jinbpark: I made an issue for this.  (https://github.com/Samsung/islet/issues/222)

Plus, I found something is wrong with the current code base. Currently,
`REC_ENTER` tries to grab both rec-lock and rd-lock in this order, REC first and RD next. But, semantically, it's not correct as RD should live longer than REC. So, if somewhere a code tries to grab locks in the order of "RD -- REC", it would face a dead-lock situation.
This problem also needs to be fixed through the above issue (#222 )

PR #216, opened 2023-10-25
Add the third-party/kvmtool-rim-measurer submodule

p-sawicki2:
This PR adds the kvmtool-rim-measurer tool for calculating Realm Initial Measurement.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -53,3 +53,7 @@
 	path = third-party/tf-a-rss
 	url = https://github.com/Samsung/islet-asset
 	branch = 3rd-tf-a-rss
+[submodule "third-party/kvmtool-rim-measurer"]
+	path = third-party/kvmtool-rim-measurer
+	url = https://github.com/Samsung/islet-asset
+	branch = 3rd-kvmtool-rim-measurer


jinbpark:
Good to see this change, which is a big step forward to support attestation!
I took a quick look at the changes you've made to "3rd-kvmtool-rim-measurer" in islet-asset.

It seems like "kvmtool" now can work in two different modes-- (1) the original mode, which populates a realm and runs it, (2) the RIM calculation mode that you made.
I get why you designed this way because this design allows us to reuse many parts of "kvmtool" codes, which is a good thing, I like this point.
(If I was mistaken at any point, please correct me)

Apart from that, could you add some documents that briefly describe the design of the RIM measurer, why it's designed this way, and how to use it considering the part to verify RIM in FVP? (added as a markdown file in islet/doc/)
That document doesn't have to be added to this PR, when to upload it is up to you.

p-sawicki2:
A brief description is here https://github.com/Samsung/islet-asset/blob/3rd-kvmtool-rim-measurer/BUILD-RIM-MEASURER

I can elaborate more how it is designed, but basically it re-uses the existing kvmtool source code and adds a layer that takes care of measurements.

jinbpark:
> A brief description is here https://github.com/Samsung/islet-asset/blob/3rd-kvmtool-rim-measurer/BUILD-RIM-MEASURER

Checked. I think this is good enough at this point. I'll copy this explanation onto [our github page](https://samsung.github.io/islet) as well, which could be a more proper place I think.

jinbpark (approved):

bokdeuk-jeong (approved):

PR #220, opened 2023-10-27
Add validity checks in rtt_map_unprotected

zpzigi754:
This PR adds missing checks in rtt_map_unprotected. All the tests in ACS's rtt_map_unprotected will be passed after this change. 

[before]
```
TOTAL TESTS     : 51

TOTAL PASSED    : 29

TOTAL FAILED    : 19
```

[after]
```
TOTAL TESTS     : 51

TOTAL PASSED    : 30

TOTAL FAILED    : 18
```

### DIFF
--- a/scripts/fvp-cca
+++ b/scripts/fvp-cca
@@ -459,18 +459,32 @@ def place_realm_at_shared(rmm, realm_ip, fvp_tap_ip, host_ip, no_kvm_unit_tests,
         run(["cp", "-R", "%s/lib" % PREBUILT_EXAMPLES, SHARED_EXAMPLES_PATH], cwd=ROOT)
         run(["tar", "-zxvf", "%s/lib/libtensorflowlite.tar.gz" % SHARED_EXAMPLES_PATH, "-C", "%s/lib/" % SHARED_EXAMPLES_PATH], cwd=ROOT)
 
-def clean_repo():
-    run(["make", "distclean"], cwd=TF_A)
-    run(["make", "distclean"], cwd=TF_A_TESTS)
-    run(["make", "clean"], cwd=REALM_LINUX)
-    run(["make", "clean"], cwd=KVMTOOL)
+def clean_repo(target):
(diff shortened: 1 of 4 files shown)


scripts/fvp-cca
```diff
@@ -570,6 +570,9 @@ if __name__ == "__main__":
             prepare_nw_aosp(args.no_prebuilt_initrd)
             prepare_bootloaders(args.rmm, PREBUILT_EDK2, args.hes)
         elif args.normal_world == "acs":
+            # Clear out build directory so that the current target is 
+            # not confused by the previous target
+            run(["rm", "-rf", "build"], cwd=ACS)
```
bitboom, line 575: [IMHO] I think it's enough to clean only when with `--clean option` or remove manually than removing every time.

bitboom (approved):

scripts/fvp-cca
zpzigi754: Good opinion! I also think that it might be enough to explicitly clean with `./scripts/fvp-cca --clean` or manually deleting `third-party/cca-rmm-acs/build`. 

Currently, `./scripts/fvp-cca --clean` not only clean `ACS`, but also clean other third-party submodules (tf-a, kvmtool realm-linux,  etc) altogether which I wanted to avoid (e.g., building realm-linux after clearing would takes a lot of time). For the second option, it requires an internal knowledge about the ACS's build system which users might be aware of. 

In my personal experience, building acs from the initial state (after removing build dir) did not add much time (a few seconds) in the testing because the module size relatively small. That's the reason I chose building ACS from the initial state every time.

zpzigi754: I will make `scripts/fvp-cca --clean` into `scripts/fvp-cca --clean <target>` (e.g., `scripts/fvp-cca --clean acs`) and do not remove build directory every time after leaving some comment that acs should be cleaned if `selected-tests` or `excluded-tests` are used!

bokdeuk-jeong (approved):

Issue #222, opened 2023-11-01
A secure yet efficient way to access from child (Rec) to parent (Rd)


jinbpark:
e.g., REC_ENTER
```
// grab the lock for Rec
let mut rec_granule = get_granule_if!(arg[0], GranuleState::Rec)?;
let mut rec = rec_granule.content_mut::<Rec>();

rec.rd.id();  // --> how to implement reading parents' variables in a secure manner? (Rd.id) 
rec.rd.ipa_bits();
```

- Rules that can apply to child-to-parent access
  - Rule-1: Child-Parent relationship has to be managed through reference-counting mechanisms. (we already doing it for Rec-Rd). This is because, at the moment of Child trying to access Parent, the Parent has to be guaranteed alive.
  - Rule-2: Child is allowed to access read-only variables that Parent has.
  - e.g., Rec (Child) is allowed to access read-only variables (realm_id, ipa_bits) that Rd (Parent) has.
    - This rule can apply to any situation in which a child-parent relationship plays. (one more example is: Rec (Parent) - RecAux (Child))
  - As long as this rule is followed, Child can access those variables without acquiring the lock of Parent.

- Main question:  how to design and implement a secure interface for the above child-to-parent access rules?
  - <TODO> there will be multiple ways to do it..

nook1208:
When I searched, there is no way to define immutable field in struct.
I think that adding private `realm_id` and `ipa_bits` field in Rec and make only `init()` and getter function is the best like the following comment in [this reference](https://stackoverflow.com/questions/23743566/how-can-i-force-a-structs-field-to-always-be-immutable-in-rust/23748710#23748710)

> The best, as [Lily Ballard noted](https://stackoverflow.com/a/23748710/155423), is that you can declare your Shape field as private and make a getter method using impl A {...}.

let me make a PR for it :)

jinbpark:
`Rec and making only init() and getter function` --> I agree this is good enough.

My initial thought on how to implement this is as follows.
```
// define accessor
impl<'a, E> EntryGuard<'a, E> {
    // new function for child-parent access
    // NOTE:  ==> add a comment here about the two rules for child-parent access
    pub fn parent_content<T: Content>(&self) -> &T {
       // E:  the type of child (e.g., Rec),   T: the type of parent (e.g., Rd)
       // <TODO>  do type checking--> it looks like, either (1) or (2). ==> this is for Rule-1.
       //     (1) if E::type == Rec && T::type == Rd,  error otherwise  
[...]

PR #223, opened 2023-11-01
Add immutable rd fields to Rec

nook1208:
It's related with #222 issue.

This PR add immutable Rd fields (realmid, ipa_bits) to `Rec` struct.
    
Multiple RECs can be created per Realm.
So the mentioned fields of rec are always valid before destroying the current `Rec`

### DIFF
--- a/rmm/src/rmi/rec/mod.rs
+++ b/rmm/src/rmi/rec/mod.rs
@@ -84,7 +84,7 @@ impl Rec {
         self.runnable
     }
 
-    pub fn id(&self) -> usize {
+    pub fn vcpuid(&self) -> usize {
         self.vcpuid
     }
 
--- a/rmm/src/rmi/rec/mod.rs
+++ b/rmm/src/rmi/rec/mod.rs
@@ -3,6 +3,7 @@ pub mod mpidr;
 pub mod params;
 pub mod run;
 pub mod vtcr;
+use crate::rmi::error::Error;
 
 pub use self::handlers::set_event_handler;
 
@@ -17,6 +18,7 @@ pub enum RmmRecAttestState {
 }
 
 pub struct Rec {
+    init_done: bool,
     attest_state: RmmRecAttestState,
     attest_challenge: [u8; 64],
     /// PA of RD of Realm which owns this REC
@@ -26,6 +28,14 @@ pub struct Rec {
     ripas: Ripas,
     vtcr: u64,
     host_call_pending: bool,
+
+    /// Immutable Rd fields
+    ///
+    /// Multiple RECs can be created per Realm
+    /// so the following fields are always valid
+    /// before destroying the current Rec
+    realmid: usize,
+    ipa_bits: usize,
 }
 
 struct Ripas {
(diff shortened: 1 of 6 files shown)


rmm/src/rmi/rec/mod.rs
```diff
@@ -26,6 +28,14 @@ pub struct Rec {
     ripas: Ripas,
     vtcr: u64,
     host_call_pending: bool,
+
+    /// Immutable Rd fields
+    ///
+    /// Multiple RECs can be created per Realm
+    /// so the following fields are always valid
+    /// before destroying the current Rec
+    realmid: usize,
```
jinbpark, line 37: I think it's better not to have a copy of "realmid and ipa_bits" in Rec. It seems unnecessarily redundant.
I prefer the way mentioned in [this comment](https://github.com/Samsung/islet/issues/222#issuecomment-1788479862) in which we define a generic interface for child-to-parent access and use it for Rd-Rec case.

That's partly because there might be another child-to-parent access scenario (e.g., Rec-RecAux which we don't support yet). We can benefit from this generic interface for future cases.

rmm/src/rmi/rec/mod.rs
```diff
@@ -17,6 +18,7 @@ pub enum RmmRecAttestState {
 }
 
 pub struct Rec {
+    init_done: bool,
```
jinbpark, line 21: I see the intention of the use of `init_done` here. I like this check for safety. Here is my thought on this.

1. `init_done: Lock<bool>`  --> in terms of "struct", `init_done` has to be under the protection of locking (no matter what type of locking is) as init() can be called multiple times in parallel.

2. `init_done: bool`  --> but it does make sense particularly for "Rec" case because it is guaranteed "Rec" is accessed after acquiring granule_rec_lock.  So, I think we can leave it as is but it would be better to leave a comment on
why locking is not needed for this value. (e.g., // Safety: it is safe as long as "Rec" is accessed after granule lock)
I think this way is good enough to go but with a proper comment on safety.

3. A more common pattern to this problem in rust is to use [OnceCell](https://docs.rs/once_cell/latest/once_cell/).
```
struct Inner { 
   attest_state:  RmmRecAttestState,
   vtcr: u64,
   ....
}
struct Rec {
    inner: OnceCell<Inner>;  // OnceCell is used to guarantee "this inner" is initialized only once.
}

impl Rec {
    init() {
        // do initialize inner here.
    }
}

```

rmm/src/rmi/rec/mod.rs
nook1208: Oh, I thought that you agreed on this

> Rec and making only init() and getter function --> I agree this is good enough.

Lately, I realized a few worries about making a `no_lock` accessor.
Someone can make a mistake in accessing a mutable field of a parent by using the no-lock accessor.
So I prefer to restrict only a few immutable fields, only if there are many situations where we need to access that immutable field.

jinbpark: As we discussed offline, it's good enough to have a separate struct and put realmid/ipa_bits into that struct. (e.g., struct RealmInfo)

jinbpark:
As we discussed offline, could you leave a comment on locking in the order of "Rec-Rd"? (e.g., this potentially could cause a dead-lock, but it is safe at least for the Rd-Rec case)  Plus, a comment on an alternative approach to do the same job with making "state" atomic variables.

jinbpark:
It seems that REC_ENTER is the only place where a manual drop is used. How about using the following simple trick to eliminate that drop?

Before
```
let g_rd = get_granule_if!(rec.owner(), GranuleState::RD)?;
let rd = g_rd.content::<Rd>();
match rd.state() {
...
}
drop(g_rd);
```

after  (it doesn't have to be the following code. Just using one more block is enough)
```
if let Ok(g_rd) = get_granule_if!(rec.owner(), GranuleState::RD) {
    let rd = g_rd.content::<Rd>();
    match rd.state() { ... }
    // we don't need drop() calls because the lifetime of g_rd ends here.
}
```

nook1208:
> As we discussed offline, could you leave a comment on locking in the order of "Rec-Rd"? (e.g., this potentially could cause a dead-lock, but it is safe at least for the Rd-Rec case)

If we make it really safe, I think that it needs an `RmmRecState` field which is one of the attributes in `A2.3.2 REC attributes`, and checking it in REC_ENTER.

Let me make a new PR for it.

nook1208:
> Plus, a comment on an alternative approach to do the same job with making "state" atomic variables.

I've been thinking about using atomic read for `rd.state` without lock. 
It turns out that there is a corner case.
The state can be read with `Old` state while the state is changing `Old` to `New` with lock.

So I think that we should use lock as usual

nook1208:
I can't push my commit to this PR because permission was denied..
So I made a new PR #225 with the same content.
Please follow the #225 PR :(

PR #224, opened 2023-11-02
Refactoring related with attestation token

bitboom:
This PR has two purpose.

## 1. Remove duplication of CCA token implementation
- issue: CCA token is implemented both `sdk` and `cli`. 
- solution: Detach the implementation of cca_token form `cli` to `lib/cca_token`.
And use `lib/cca_token` in `sdk`.

## 2. `Report format` has differences between `aarch64` and `x86_64`
- issue: The report for `x86_64`  had `user_data` for simulation. This differences makes `decoding error` when it comes to run certifier demo. The client gets `attestation report` on aarch64 and the service decode it on x86_64.
- solution: Use same format on `aarch64` and `x86_64`


jinbpark (approved):
Good findings-:

hihi-wang (approved):

zpzigi754:
Just ask out of curiosity. 

Do we use `x86_64` for building sdk easier in the host environment? Or for supporting multiple targets?

zpzigi754 (commented):
Is including the binary file (`sdk/sdk-example-c`) intended?

bitboom:
> Just ask out of curiosity.
> 
> Do we use `x86_64` for building sdk easier in the host environment? Or for supporting multiple targets?

There are two reason.

1. Supports Certifier f/w. (Certifier requires simulation, named [islet-shim](https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/main/src/islet/islet_shim.cc), running on x86_64)
2. Builds sdk easier in the host environment as you said.

bitboom:
> Is including the binary file (`sdk/sdk-example-c`) intended?

It was my miss :( . I added it to `.gitignore`. Thanks!

zpzigi754 (approved):
I guess that it might be better to restore the original commit message (`sdk: Use lib/cca-token crate`) and push the last commit on top of it.

[current]
```
Detach the token module to lib/cca-token 
Add sdk-example-c binary to gitignore
```
[suggestion]
```
Detach the token module to lib/cca-token 
sdk: Use lib/cca-token crate
Add sdk-example-c binary to gitignore
```

bitboom:
> I guess that it might be better to restore the original commit message (`sdk: Use lib/cca-token crate`) and push the last commit on top of it.
> 
> [current]
> 
> ```
> Detach the token module to lib/cca-token 
> Add sdk-example-c binary to gitignore
> ```
> 
> [suggestion]
> 
> ```
> Detach the token module to lib/cca-token 
> sdk: Use lib/cca-token crate
> Add sdk-example-c binary to gitignore
> ```

I applied it. Thanks :)

PR #225, opened 2023-11-03
Add immutable rd fields to Rec

nook1208:
It's related #222 issue and it's from #223 because a push permission was changed..

This PR add immutable Rd fields (realmid, ipa_bits) to Rec struct.

Multiple RECs can be created per Realm.
So the mentioned fields of rec are always valid before destroying the current Rec

### DIFF
--- a/rmm/src/rmi/rec/handlers.rs
+++ b/rmm/src/rmi/rec/handlers.rs
@@ -58,12 +58,15 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         }
 
         for (idx, gpr) in params.gprs.iter().enumerate() {
-            if rmi.set_reg(rd.id(), rec.id(), idx, *gpr as usize).is_err() {
+            if rmi
+                .set_reg(rd.id(), rec.vcpuid(), idx, *gpr as usize)
+                .is_err()
+            {
                 return Err(Error::RmiErrorInput);
             }
         }
         if rmi
-            .set_reg(rd.id(), rec.id(), 31, params.pc as usize)
+            .set_reg(rd.id(), rec.vcpuid(), 31, params.pc as usize)
             .is_err()
         {
             return Err(Error::RmiErrorInput);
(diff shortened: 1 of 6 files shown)


rmm/src/rmi/rec/handlers.rs
```diff
@@ -99,7 +102,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             return Err(Error::RmiErrorRec);
         }
 
-        let g_rd = get_granule_if!(rec.owner(), GranuleState::RD)?;
+        let g_rd = get_granule_if!(rec.owner()?, GranuleState::RD)?;
```
jinbpark, line 105: I found `drop(g_rd);` still exists here. Could you eliminate this as I mentioned when this PR was in draft?

rmm/src/rmi/rec/mod.rs
```diff
@@ -16,31 +18,118 @@ pub enum RmmRecAttestState {
...
 
+struct Ripas {
+    start: u64,
+    end: u64,
+    addr: u64,
+    state: u8,
+}
+
+#[derive(Debug)]
+pub struct Inner {
```
bitboom, line 29: [minor] How about delete `pub` visibility?

bitboom (approved):

rmm/src/rmi/rec/handlers.rs
nook1208: Oh, I missed it, 
Fixed, thank you :)

rmm/src/rmi/rec/mod.rs
nook1208: Fixed, thank you !

rmm/src/rmi/rec/mod.rs
```diff
@@ -16,31 +18,118 @@ pub enum RmmRecAttestState {
...
+}
+
+/// Immutable Realm information in target Rd.
+///
+/// It can be written to only once because of OnceCell.
+/// Multiple RECs can be created per Realm
+/// so the fields in Inner are always valid
+/// before destroying the current Rec.
+struct ImmutRealmInfo {
+    inner: OnceCell<Inner>,
```
jinbpark, line 64: Good to see OnceCell here-:

jinbpark (approved):

PR #227, opened 2023-11-07
scripts: Apply realm_linux feature to rmm naturally

bitboom:
This PR enables `realm_linux` feature to rmm automatically when the realm is the linux.

## Before
```sh
./scripts/fvp-cca -nw=linux -rm=linux --realm-linux
```

## After
```sh
./scripts/fvp-cca -nw=linux -rm=linux
```

### DIFF
--- a/scripts/fvp-cca
+++ b/scripts/fvp-cca
@@ -204,7 +204,7 @@ def get_rmm_features(args):
 
         if args.stat == True:
             features += ["--features", "stat"]
-        if args.realm_linux == True:
+        if args.realm == "linux" or args.use_prebuilt == True:
             features += ["--features", "realm_linux"]
 
     if features:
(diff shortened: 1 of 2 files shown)


scripts/tests/realm-boot.sh
```diff
@@ -26,7 +26,7 @@ check_result()
 }
 
 tar -xf $ROOT/assets/prebuilt/out.tar.bz2 -C $ROOT
-$ROOT/scripts/fvp-cca -bo -rmm=islet --use-prebuilt --rmm-log-level=error --realm-linux
+$ROOT/scripts/fvp-cca -bo -rmm=islet --use-prebuilt --rmm-log-level=error --realm=linux
 $ROOT/scripts/fvp-cca -ro -nw=linux --realm=linux -rmm=islet --no-telnet &
```
zpzigi754, line 30: The above lines should be changed like the below  (only `ro` option should use `realm=linux`)
```
$ROOT/scripts/fvp-cca -bo -rmm=islet --use-prebuilt --rmm-log-level=error
$ROOT/scripts/fvp-cca -ro -nw=linux --realm=linux -rmm=islet --no-telnet &
```
It would boot realm-linux successfully.

scripts/fvp-cca
```diff
@@ -204,7 +204,7 @@ def get_rmm_features(args):
 
         if args.stat == True:
             features += ["--features", "stat"]
-        if args.realm_linux == True:
+        if args.realm == "linux":
```
zpzigi754, line 207: The above line can be changed to 
```
if args.realm == "linux" or args.use_prebuilt == True:
```

bitboom: Applied it !

scripts/tests/realm-boot.sh
bitboom: Thank you for the kind review 👍

zpzigi754 (approved):

bokdeuk-jeong (approved):

PR #228, opened 2023-11-07
Initial preparation for model checking

zpzigi754:
This PR prepares for basic environment setup for model checking. [Slightly modified kani](https://github.com/zpzigi754/kani/tree/use-aarch64) has been used to verify with `aarch64` target instead of `x86_64` target, yet using the unmodified one for normal usage would be fine. Verification harnesses will be added later.


bitboom (approved):
Really nice to see kani here!

nook1208 (approved):

hihi-wang:
Good to start with model checking with Kani!

PR #229, opened 2023-11-08
Add refactor commits for Rec and Realm Exit

nook1208:
There are refactor commits for Rec and Realm Exit.
The main commit is :

- refactor: Use owner instead of ImmutRealmInfo in Rec    

It uses Rec::owner as a reference of Rd to dereference it.
Note that only immutable fields of Rd must be dereferenced
by rec.owner by making getter method for the safety


rmm/src/rmi/rec/handlers.rs
```diff
@@ -96,13 +97,26 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
 
+        // Safety:
+        // Holding lock in the order of "Rec->Rd"(B) potentially could
+        // cause deadlock, because common locking order here is "Rd->Rec"(A):
+        // A: Rd.lock() held -> wait Rec.lock()
+        // B: Rec.lock() held -> wait Rd.lock()
+        //
+        // but the following condition can prevent the deadlock problem
+        // by blocking 'B' case only if the same Rec is already running.
+        if let RecState::Running = rec.get_state() {
```
jinbpark, line 115: This check can't guarantee no deadlock.
See the following situation, time goes from top to bottom.

```
<T1> Rec.lock()                        |
          ....                         |       <T2> Rd.lock()
          rec.get_state() check        | 
          (check passed)               |
<T1>  Rd.lock() --> wait here          |
                                       |  <T2> Rec.lock()   --> wait here. deadlock happened.
```

There are only two ways in theory to prevent this deadlock problem.
(1) make Rd's state atomic variable and do not grab Rd lock.
(2) just rely on the fact that there is no situation of the locking in the order of "RD-REC"  (true at least for Rd-Rec case)

If we choose (2), there is no need to have an additional check like `rec.get_state()`, if rec.state is only for deadlock checking.
Though, I still think (1) (using atomic variable) is a better way to go.

nook1208: I agreed that it looks difficult policy about lock as we discussed offline. 
So I removed the comment. thank you !

jinbpark:
As we discussed offline, I like the idea that enforces that more than two RECs are never going to be running at the same time. And I'm wondering what happens if we try to make that happen. (maybe, possible attack case??) That would be interesting.

rmm/src/rmi/realm/rd.rs
```diff
@@ -4,6 +4,7 @@ use crate::rmi::rtt::realm_par_size;
 use vmsa::guard::Content;
 
 // TODO: Integrate with our `struct Realm`
+#[derive(Debug)]
```
hihi-wang, line 7: Is is intentional?

nook1208: Yes. It needs to print the error log in REC_ENTER or can be used another place for debugging later:
```rust
listen!(mainloop, rmi::REC_ENTER, |arg, ret, rmm| {
...
if let RecState::Running = rec.get_state() {
            error!("Rec is already running: {:?}", rec); //print rec and the parent rd also
            return Err(Error::RmiErrorRec);
}
```

hihi-wang (approved):

rmm/src/rmi/rec/mod.rs
```diff
@@ -34,104 +35,35 @@ struct Ripas {
...
-pub struct Rec {
+pub struct Rec<'a> {
     attest_state: RmmRecAttestState,
     attest_challenge: [u8; 64],
     /// PA of RD of Realm which owns this REC
+    ///
+    /// Safety:
+    /// Only immutable fields of Rd must be dereferenced by owner
+    /// by making getter method for the safety
+    owner: OnceCell<&'a Rd>,
```
zpzigi754, line 46: Good to bind rd's lifetime with a reference!

bitboom (approved):

rmm/src/event/realmexit.rs
```diff
@@ -11,55 +11,56 @@ use crate::Monitor;
...
+pub enum RecExitReason {
     Sync(ExitSyncType),
     IRQ = 1,
     FIQ = 2,
     PSCI = 3,
-    RipasChange = 4,
-    HostCall = 5,
-    SError = 6,
-    Undefined = RMI_REC_EXIT_REASON_MASK, // fixed, 0b1111
+    SError = 4,
```
zpzigi754, line 20: I am personally against this change (and the commit) because of the previous comment: not syncing with the spec. I  think that `trap::handle_lower_exception()` is not the only place to use `RmiRecExitReason`. For example,  `RmiRecExitReason::RipasChange` can be used while handling rsi call.

rmm/src/rmi/rec/mod.rs
```diff
@@ -19,6 +19,13 @@ pub enum RmmRecAttestState {
     NoAttestInProgress,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub enum RecState {
+    Ready = 0,
+    Running = 1,
+}
```
zpzigi754, line 27: FYI, we already have `Ready` and `Running` in realm/vcpu.rs like the below. 
```
#[derive(Debug)]
pub enum State {
    Null,
    Ready,
    Running,
}
```

I think that using the existing one or removing the existing one would be necessary.

zpzigi754: I think that `vcpu::State::Running` and `RecState::Running` should mean the same state. Is there any case that they are in different states? (e.g., `vcpu::State` is `Running` while `RecState` is not `Running`)

nook1208: > I think that vcpu::State::Running and RecState::Running should mean the same state

I see. let me think about it :)

> Is there any case that they are in different states? (e.g., vcpu::State is Running while RecState is not Running)

Oh, I mean, just the same enum names 'State'. maybe we can use `alias`..?

zpzigi754: > Oh, I mean, just the same enum names 'State'. maybe we can use `alias`..?

I guess that it would be okay to have the same name, as long as they are in different namespaces. I also think that removing either `rmi/realm/rd.rs`'s `State` or `realm/mod.rs`'s `State` is necessary, as they are duplicate (not about this PR).

rmm/src/event/realmexit.rs
bokdeuk-jeong: The spec is about the interface between the RMM and the host or the RMM and a realm. As the current rmi exactly does that while using the information from the exception handler, we are in line with the spec.

On the other hand, RecExitReason is internally shared inside the RMM, i.e. not externally shared with the host or a realm via rmi/rsi. 
As the exception handler doesn't directly interact with external entities, it shouldn't be limited in its expression of information by the spec.

bokdeuk-jeong (approved):

rmm/src/rmi/rec/mod.rs
nook1208: Fixed, Thank you !

rmm/src/event/realmexit.rs
zpzigi754: I thought that the previous `RmiRecExitReason` would be eventually used to set `RecExit`'s `exit_reason` field as in the spec, but currently different but similar const values in `rmm/src/rmi/mod.rs` are used to set that field. The duplication between those const values and `RecExitReason` might need to be resolved later.

PR #234, opened 2023-11-14
Initial harness for verifying rmi features

zpzigi754:
This is an initial harness for verifying rmi features with model checking. Several data structures and functions were made unreachable to avoid a state explosion (will be handled more gracefully with stubbed functions and modular verification later). In the harness, input and output pair of RMI_FEATURES command is checked to conform with the spec as similarly done in TF-RMM's harness.

It found one correctness bug (#301).

## How to use? ##
```
(in islet/model-checking)

RMI_TARGET=rmi_features make verify

...
SUMMARY:
 ** 0 of 2626 failed (75 unreachable)

 ** 4 of 4 cover properties satisfied
```


hihi-wang:
Nice to have initial step for verification!

bokdeuk-jeong:
Could you explain what properties of an RMI do you intend to verify?

zpzigi754:
> Could you explain what properties of an RMI do you intend to verify?

Good question. This PR verifies low-level properties: free of the undesired behaviors including buffer overflow, invalid pointer dereferences, integer overflow, misaligned pointers, etc.

bokdeuk-jeong:
Do you plan to verify other properties of RMIs other than the properties that you mentioned above using klee or kani?
I was wondering if you also intend to define and verify properties that needs care in the RMM layer especially.

zpzigi754:
> Do you plan to verify other properties of RMIs other than the properties that you mentioned above using klee or kani? I was wondering if you also intend to define and verify properties that needs care in the RMM layer especially.

I will answer it in private.

jinbpark:
Just out of curiosity, how difficult is it to integrate KANI into our CI, in terms of engineering?

zpzigi754:
> Just out of curiosity, how difficult is it to integrate KANI into our CI, in terms of engineering?

Good question. I actually plan to intergrate kani and verification harnesses into our CI. It would not require much engineering effort. One thing to consider is that some verification harnesses would take a considerable time to check and having many harnesses might harm the usability in CI (e.g., spending hours to check the whole).

plat/fvp/src/mc/rmi_features.rs
```diff
@@ -0,0 +1,44 @@
...
+    let no_failures_pre = true;
+    let success_index_pre = index != 0;
+
+    // Execute command and read the result.
+    let out = Monitor::new().run(regs);
+    let result = out[0];
+    let value = out[1];
+
+    // Result assertion
+    let prop_result_ante = no_failures_pre;
```
bitboom, line 23: Can i ask what does ante mean?

plat/fvp/src/mc/rmi_features.rs
```diff
@@ -0,0 +1,44 @@
+use islet_rmm::monitor::Monitor;
+use islet_rmm::rmi;
+
+#[kani::proof]
+#[kani::unwind(4)]
```
bitboom, line 5: I am not familiar with formal verification, so I ask out of curiosity. Is there a particular reason for choosing 4?

bitboom (commented):
First of all, I think it is very meaningful that we can actually do Model Checking by introducing Kani to Islet.

However, it seems that the inevitable conditional compilation makes the code flow difficult. I expect this pattern to increase for the time being.

So here's a careful suggestion. Why don't we expose the functions or data structures required for model checking to proceed with model checking in an external crate? (ex. lib/model-checker) It would be appreciated if you could also add reasons why functions requires mocking for model checking. 

Again, thank you for your effort.

plat/fvp/src/mc/rmi_features.rs
```diff
@@ -0,0 +1,44 @@
...
+fn verify_features() {
+    // Initialize registers (symbolic input)
+    let regs: [usize; 8] = kani::any();
```
bokdeuk-jeong, line 8: kani::any() looks like generating a random value of a specific type. 
Could you explain more about how is a symbolic value in this kani different from any random concrete value? 
My understanding of symbolic execution is based on https://arxiv.org/pdf/1610.00502.
As in Fig.2, I thought the value representation of each variable is symbolic with constraints.

plat/fvp/src/mc/rmi_features.rs
zpzigi754: The variable's name comes from tf-rmm's harness and I think that it means [antecedent](https://en.wikipedia.org/wiki/Antecedent_(logic)).

plat/fvp/src/mc/rmi_features.rs
zpzigi754: In bounded model checking, `unwind` incidates the bound: the maximum number to iterate in loops. In general, choosing a small value helps to reduce the verification time while sacrificing the soundness. I chosed the value 4, because the vector encountered by the harness requires using the value larger than 3 in order not lose the soundness (e.g., the vector contains only 3 elements).

bitboom: Thanks for the explanation.

plat/fvp/src/mc/rmi_features.rs
zpzigi754: `kani::any()` generates a symbolic value of a specific type, not concrete value. In contrast to dynamic symbolic execution such as KLEE and DART where random concrete values are used for helping to solve the path constraints, no random concrete value is used for general symbolic execution as well as model checking including `kani`. Random concrete value can be thought of as an internal mechanism used in dynamic symbolic execution to help improve the efficiency in symbolic execution, whereas symbolic value can be thought of as a random value which is not concrete. 

If the value is generated with `kani::any()`, its representation becomes symbolic with constraints as in the mentioned Fig.2.

zpzigi754:
> However, it seems that the inevitable conditional compilation makes the code flow difficult. I expect this pattern to increase for the time being.
> 
> So here's a careful suggestion. Why don't we expose the functions or data structures required for model checking to proceed with model checking in an external crate? (ex. lib/model-checker)

Thank you for the valuable comment and suggestion! I also think that using conditional compilation might harm the code's readability. 

About exposing the functions or data structures in an external crate, the harness and related functions are already separated in `./plat/fvp/src/mc`. Surely, it can be moved to another place: I am not sure which place would be better though (maybe, `./islet/verification/model-checking` or `./islet/model-checking/`). 

Another decision to make would be where to put mock data structures and functions. I think that putting them near the original ones or not have their own trade-offs. If it is located far away from the original place (e.g., another crate), there could be higher chance that the mock structures and functions are deviated from the original ones as well as making them difficult to grasp the difference. In addition, redundant code would be increased, as using common code would become more restricted. 

Also, it might be against the general convention. For example, the below is a snippet of amazon's firecracker (https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/iovec.rs). It also put the mock structure near the original one.
```
#[cfg(kani)]
type IoVecVec = Vec<iovec>;
#[cfg(not(kani))]
type IoVecVec = SmallVec<[iovec; 4]>;
```
> It would be appreciated if you could also add reasons why functions requires mocking for model checking.

Regarding adding explanation about why mock is required, I will add the comments.

bitboom:
> About exposing the functions or data structures in an external crate, the harness and related functions are already separated in `./plat/fvp/src/mc`. Surely, it can be moved to another place: I am not sure which place would be better though (maybe, `./islet/verification/model-checking` or `./islet/model-checking/`).

Among them, I vote for `islet/model-checking`.

> Another decision to make would be where to put mock data structures and functions. I think that putting them near the original ones or not have their own trade-offs. 

I totally agree with you. Currently, it seems faster to develop and easier to compare together. Also, parts that will naturally may be merged. 

It might be good to separate it later when the verification code size grows larger and there are more branches.

zpzigi754:
> Among them, I vote for `islet/model-checking`.

I've moved the harness into `islet/model-checking` and added comments on the places where conditional build is used.

bitboom (approved):

PR #245, opened 2023-12-04
draft: Enable sdk with openssl-aarch64

bitboom:


zpzigi754:
> I put the whole openssl headers which are not modiled. Which means [camellia.h](https://github.com/openssl/openssl/blob/master/include/openssl/camellia.h) is not related `our secure os project`. Nevertheless, do you think you should erase it?

Oh, I didn't know that there is an encryption algorithm called [Camellia](https://en.wikipedia.org/wiki/Camellia_(cipher)) which has the same name of our former project. Not removing it would be correct!

zpzigi754 (approved):
Thank you for adding openssl on aarch64 target!

PR #249, opened 2023-12-13
sdk: Fit return's size of un/seal() properly

bitboom:
I found this issue while running the `certifier` example.
When a certifier's app calls unseal(), the return value should be fit the decrypted size without the plaintext's size.

### DIFF
--- a/sdk/src/sealing.rs
+++ b/sdk/src/sealing.rs
@@ -10,9 +10,10 @@ pub fn seal(plaintext: &[u8]) -> Result<Vec<u8>, Error> {
     let padding = Padding::PKCS1;
     let len = core::cmp::max(plaintext.len(), pri_key.size() as usize);
     let mut sealed = vec![0 as u8; len];
-    let _ = pri_key
+    let len = pri_key
         .public_encrypt(plaintext, &mut sealed, padding)
         .or(Err(Error::Sealing))?;
+    sealed.truncate(len);
     Ok(sealed)
 }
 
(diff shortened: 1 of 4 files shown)


jinbpark (approved):

nook1208 (approved):
Good find !

sdk/src/sealing.rs
```diff
@@ -10,19 +10,21 @@ pub fn seal(plaintext: &[u8]) -> Result<Vec<u8>, Error> {
     let padding = Padding::PKCS1;
     let len = core::cmp::max(plaintext.len(), pri_key.size() as usize);
     let mut sealed = vec![0 as u8; len];
-    let _ = pri_key
+    let len = pri_key
```
zpzigi754, line 13: I guess that the returned `len` could be compared to `plaintext.len()` in `assert_eq`? (the same comment can be applied to `unseal` as well).

zpzigi754 (approved):
Good to fix it!

hihi-wang (approved):

sdk/src/sealing.rs
bitboom: Thank you for the comment. Plaintext.len is not equal to sealed.len(encrypted_len).

In our test code, lengths are like:
 - plaintext: 9
 - sealed: 256
 - unsealed: 9

zpzigi754: Oh, I see. Thank you for the trial! (I've also confirmed that there is a case that the suggested comparison does not work where the key's length is larger than the plaintext's length).

PR #235, opened 2023-11-27
Remove rmi::Interface trait

zpzigi754:
This PR contains the refactoring commit related to the earlier discussion (#231). Previously, RMI interface trait was mainly used to sidestep the dependency issue between monitor and armv9a. Now that the dependency issue has been resolved by the integration (#175), the trait itslef doesn't have to be maintained.

[before]
```
rmm/src/realm/registry.rs contained most of the detailed RMI implementation with trait.
```

[after]
```
The below contains each module's corresponding RMI implementation or auxiliary functions without trait
[ 
  rmm/src/realm/vcpu.rs       { create_vcpu }
  rmm/src/realm/timer.rs      { send_(timer_)state_to_host }
  rmm/src/realm/context.rs    { set_reg, get_reg }
  rmm/src/rtt.rs              { (rtt_)create, (rtt_)destroy, ... }
  rmm/src/gic.rs              { send_(gic_)state_to_host, receive_(gic_)state_from_host }
  rmm/src/mmio.rs             { emulate_mmio }
  ... 
]

Note that the names inside the parentheses were removed to avoid duplicate words in namespace (e.g., rtt::rtt_create -> rtt:create).
```

PR #251, opened 2023-12-13
Remove rsi::Interface trait

zpzigi754:
This PR contains the refactoring commit related to the earlier discussion (#231) and PR (#235). Previously, RSI interface trait was mainly used to sidestep the dependency issue between monitor and armv9a. Now that the dependency issue has been resolved by the integration (#175), the trait itslef doesn't have to be maintained.

[before]
```
rmm/src/realm/registry.rs contained the detailed RSI implementation with trait.
```

[after]
```
The below contains each module's corresponding RSI implementation without trait.
{
  rmm/src/rsi/measurement.rs      { (measurement_)read, 
                                    (measurement_)extend }
  rmm/src/rsi/attestation/mod.rs  { get_(attestation_)token }
}

Note that the names inside the parentheses were removed to avoid duplicate words in namespace 
(e.g., measurement::measurement_read => measurement::read).
```

### DIFF
--- a/rmm/src/measurement/ctx.rs
+++ b/rmm/src/measurement/ctx.rs
@@ -4,36 +4,32 @@ use super::{
 };
 use crate::rmi::rec::params::Params as RecParams;
 use crate::{
-    event::RsiHandle,
     host::DataPage,
     rmi::realm::{params::Params as RealmParams, Rd},
-    rsi::{self, Interface},
+    rsi,
 };
 
 pub struct HashContext<'a> {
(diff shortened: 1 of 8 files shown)


jinbpark (approved):
Nice refactoring-:

bitboom (approved):

hihi-wang (approved):

hihi-wang:
Looks more clear :)

nook1208 (approved):

Issue #247, opened 2023-12-05
The list of code patterns where verification does not work well


zpzigi754:
The below will summarizes the code patterns which are not supported by formal verification tools. 

[KLEE]
```
lazy_static: causes a compilation error (== abort) in llvm IR
```

bokdeuk-jeong:
I understand why we need pr252. However, I still don't understand why 'fn name(self)' results aborts while processing Islet with klee. Could you explain more about this?

PR #252, opened 2023-12-15
klee: Fix self's usage where its reference is implied

zpzigi754:
This PR is a prerequisite for KLEE integration. Before this change, KLEE has aborted on compiling the previous code.

Relates to one of the cases in #247

### DIFF
--- a/rmm/src/realm/mm/stage2_tte.rs
+++ b/rmm/src/realm/mm/stage2_tte.rs
@@ -95,7 +95,7 @@ impl S2TTE {
         Ok((S2TTE::from(s2tte as usize), last_level))
     }
 
-    pub fn is_valid(self, level: usize, is_ns: bool) -> bool {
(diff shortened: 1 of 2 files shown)


hihi-wang:
Just for curious, why does methods takes ownership instead of reference `&self`? Most of does methods check for some status, with naming `is_XXX`, so I found no reason this to be `self`...

hihi-wang (approved):

nook1208 (approved):

bitboom (approved):

zpzigi754:
> Just for curious, why does methods takes ownership instead of reference `&self`? Most of does methods check for some status, with naming `is_XXX`, so I found no reason this to be `self`...

Good question. I think that `&self` should be used when borrow semantic is used while `self` be used when move semantic is used. I guess that the previous `self` was used mistakenly for ownership, or `self` might be allowed for borrow semantic as a syntactic sugar.

PR #253, opened 2023-12-18
Add a cross-platform e2ee example

bitboom:
This PR adds a cross-platform e2ee example.

The example consists of below components & scripts shortly:
For more information, please refer this [README](https://github.com/Samsung/islet/blob/certifier/examples/cross-platform-e2ee/README.md).

This PR requires https://github.com/Samsung/islet-asset/pull/11. 
*I will merge that before this PR is merged.*

## Cross-platform E2EE demo using Ceritifer
- Components
  - Certifier service (x64)
  - E2EE server app (x64)
  - E2EE client app (aarch64)
- Scripts
  - setup.sh: Setup prerequisites to build & run the demo
  - provisioning.sh: Provisioning stuffs to establish a secure channel
  - build-{service, server, client}.sh: Build each component
  - run-{service, server, client}.sh: Run each component

---

To make this PR concise, i will make following PRs for below:

- Add this example to CI (build process)
- Make PR on certifier ([islet-example-aarch64 branch](https://github.com/vmware-research/certifier-framework-for-confidential-computing/tree/bitboom/islet-example-aarch64) to main)
- Update our developer site

PR #254, opened 2023-12-27
Support to get measurements both x64 & arm64

bitboom:
This PR is for adding a comment, `How to get measurements`, to the cross-platform e2ee example.
_This PR is following PR of https://github.com/Samsung/islet/pull/253 especially reflecting [this comment](https://github.com/Samsung/islet/pull/253#discussion_r1433702390)._

## Added comment

You can get the server measurement with cli on the host machine:
```sh
$ cd $(islet)/cli
$ make x86_64
$ ./islet measur-read --index 0
Simulated attestation operation on x86_64.
"580bd77074f789f34841ea9920579ff29a59b9452b606f73811132b31c689da9"
```

You also can get the client measurement with cli on the realm:

```sh
$ $(islet)/scripts/fvp-cca -nw=linux -rm=linux --hes
$ telnet localhost 5000
$ ./launch-realm.net
# cd /shared
# insmod rsi.ko
# ./islet measur-read --index 0
[ 2533.544864] rsi: device rsi open
[ 2533.664676] rsi: ioctl: measurement_read: 0
[ 2533.831443] rsi: device rsi released
"8a1d5cd26a0ee477067d18e2ff051687d18e1450b513508ba98910fa262b1fa3"
```

## Known issue
When gets the measurement on the FVP the measurement has padded zero value.
"8a1d5cd26a0ee477067d18e2ff051687d18e1450b513508ba98910fa262b1fa300000000000000000000000000..."

I will check it out after minor refactoring job.


cli/src/main.rs
```diff
@@ -1,5 +1,14 @@
-mod checks;
-mod subcmds;
+cfg_if::cfg_if! {
+    if #[cfg(target_arch="aarch64")] {
+        mod checks;
+        mod subcmds_arm64;
+        use subcmds_arm64 as subcmds;
+    } else {
```
zpzigi754, line 6: How about using `else if #[cfg(target_arch="x86_64")]` explicitly? The else branch for unexpected targets then could be filled with `unreachable!()` macro. 

The same comment can be applied to other places too.

examples/cross-platform-e2ee/README.md
```diff
@@ -43,10 +43,37 @@ policies by editing the `provisioning.sh` script and
 specifying your own values for the following variables:
 *Those instructions are quoted from [well-descripted document](https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/main/sample_apps/simple_app_under_islet/instructions.md)*
 
+You can get the server measurement with cli on the host machine:
+```sh
+$ cd $(islet)/cli
+$ make x86_64
+$ ./islet measur-read --index 0
```
zpzigi754, line 50: I guess that `measurement-read` or `meas-read` might be better if the command name can be fixed (Reference: https://www.allacronyms.com/measurement/abbreviated).

zpzigi754 (approved):
Good to update the document and related setup for the measurement!

Is there any reason why CLI on x64 supports less commands compared to aarch64? 

I think that the `subcmds.rs` could be also conditionally compiled if many code are in common between `subcmds_arm64.rs` and `subcmds_x64.rs`.

jinbpark (approved):

hihi-wang (approved):

nook1208 (approved):

cli/src/main.rs
bitboom: I applied it to three cfg_if.

examples/cross-platform-e2ee/README.md
bitboom: I guess the origin author wants to fit subcmd's length as 5 like below:
- attest
- verify
- measur-*

But it's good to fix it as your suggestion.
Thank you for the comment and the link (it looks very useful) !

bitboom:
> Is there any reason why CLI on x64 supports less commands compared to aarch64?
> 
> I think that the `subcmds.rs` could be also conditionally compiled if many code are in common between `subcmds_arm64.rs` and `subcmds_x64.rs`.

## x64
x64 version runs like simulated without our RMM. It just **parses hard-coded report** to provide information.

```
developers <-> cli
```

## aarch64
aarch64 version runs with our whole components like below:

```
developers <-> cli <-> rsi.ko (kernel module) <-> RMM <-> HES
```

---

So, I think there are two options.

1. Apply conditional compilation to cli layer (AS-IS)
2. Apply conditional compilation to kernel module layer

If we choose option 2. We are able to common CLI code.
But we still have conditional compilation on kernel module layer.

PR #256, opened 2024-01-08
vmsa: Add MemAlloc trait for PageTable

nook1208:
This PR is based by the following requests:
1. Subtable's entry size don't need to be fixed by the first page table entry size 'N'
2. A new trait for memory alloc/dealloc is needed for PageTable structure for allowing to have different memory management methods per PageTable. (especially for stage2 page table)

### DIFF
--- a/lib/vmsa/src/page_table.rs
+++ b/lib/vmsa/src/page_table.rs
@@ -5,6 +5,8 @@ use super::error::Error;
 use super::guard::EntryGuard;
 use super::page::{Page, PageIter, PageSize};
 
+use alloc::alloc::Layout;
+
 use core::marker::PhantomData;
 
 // Safety/TODO:
@@ -76,11 +78,29 @@ pub trait Entry {
     fn points_to_table_or_page(&self) -> bool;
 }
 
+/// Safety: the caller must do proper error handling if it's failed to allocate memory
+pub trait MemAlloc {
+    unsafe fn alloc(layout: Layout) -> *mut u8 {
+        alloc::alloc::alloc(layout)
+    }
+
+    unsafe fn alloc_zeroed(layout: Layout) -> *mut u8 {
+        alloc::alloc::alloc_zeroed(layout)
+    }
+
+    unsafe fn dealloc(ptr: *mut u8, layout: Layout) {
+        alloc::alloc::dealloc(ptr, layout);
+    }
+}
+
 pub struct PageTable<A, L, E, const N: usize> {
     entries: [E; N],
     level: PhantomData<L>,
     address: PhantomData<A>,
 }
+
+impl<A: Address, L: Level, E: Entry, const N: usize> MemAlloc for PageTable<A, L, E, N> {}
+impl<A: Address, L: HasSubtable, E: Entry, const N: usize> MemAlloc for PageTable<A, L, E, N> {}
 
 pub trait PageTableMethods<A: Address, L, E: Entry, const N: usize> {
     fn new_with_base(base: usize) -> Result<*mut PageTable<A, L, E, N>, Error>;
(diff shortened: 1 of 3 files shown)


hihi-wang:
Correct me if I'm wrong with my understanding:

1. Memory alloc/dealloc is not allowed for only stage2 translation. (by spec)
2. Previously, checking logic was absent for this.
3. To add this, define new trait `MemAlloc`, apply dynamic dispatch.

nook1208:
> Correct me if I'm wrong with my understanding:
You're right :)

rmm/src/realm/mm/stage2_translation.rs
```diff
@@ -114,6 +115,28 @@ impl<'a> Stage2Translation<'a> {
     }
 }
 
+impl<'a> MemAlloc for Stage2Translation<'a> {
+    unsafe fn alloc(layout: Layout) -> *mut u8 {
+        error!("alloc for Stage2Translation is not allowed. {:?}", layout);
+        core::ptr::null_mut()
```
jinbpark, line 121: I think it would be better to have a comment with "Safety" tag here, as it doesn't use "Result" as a return value.
(I know that we already have proper error handling in lib/vmsa/src/page_table.rs.  But "safety" comment is still needed for future usage)
```
// Safety: the caller must do proper error handling with this null pointer.
core::ptr::null_mut()
```

This comment can be applied to other places.

nook1208: applied it, thanks :)

nook1208:
> So I wonder what the purpose of MemAlloc trait is (e.g., a wrapper for supporting Self::alloc).

In my opinion, we can't guarantee that there is no caller of alloc/dealloc forever. 
So in this case, restricting a memory allocation for stage 2 page table is helpful for defensive programming. 
And I think that it's more straightforward to figure out that it doesn't do any memory allocation by specifying MemAlloc implementation.

zpzigi754 (approved):

jinbpark (approved):

lib/vmsa/src/page_table.rs
```diff
@@ -342,12 +341,15 @@ where
...
                         .unwrap(),
                     )
-                } as *mut PageTable<A, L::NextLevel, E, N>;
+                }
+                    as *mut PageTable<A, L::NextLevel, E, { L::NextLevel::NUM_ENTRIES }>;
 
                 if subtable as usize != 0 {
-                    let subtable_ptr = subtable as *mut PageTable<A, L::NextLevel, E, N>;
+                    let subtable_ptr = subtable
+                        as *mut PageTable<A, L::NextLevel, E, { L::NextLevel::NUM_ENTRIES }>;
```
zpzigi754, line 368: [minor]
I guess that removing the curly braces outside `L::NextLevel::NUM_ENTRIES` might look more clear.

nook1208: Yes, I agree..! 
But I'm afraid that it's needed because of the following error : 
```bash
error[E0747]: type provided when a constant was expected
   --> lib/vmsa/src/page_table.rs:364:59
    |
364 |                     as *mut PageTable<A, L::NextLevel, E, L::NextLevel::NUM_ENTRIES>;
    |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: if this generic argument was intended as a const parameter, surround it with braces
    |
364 |                     as *mut PageTable<A, L::NextLevel, E, { L::NextLevel::NUM_ENTRIES }>;
    |                                                           +                           +

```

bitboom (approved):

hihi-wang:
LGTM :)

PR #275, opened 2024-02-19
Add an array-based granule status table

zpzigi754:
This PR adds an array-based granule status table. The intention is for reducing the overall complexity of page table-based one, helping to reason about the implementation. The imposed memory overhead is approximately 2 MB == 2 bytes (ENTRY_SIZE) * 0xfc000 (RMM_MAX_GRANULES). To select the previous implementation, `gst_page_table` feature can be
used.

## The change on the caller side
```
[before]
// Previously, the below match was used to expand the page table entry.
let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
     Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated)
     other => other,
}?;

[after]
// The match is no longer used, as the array will not be expanded.
let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
```

## Summary of Changes
- Remove `set_state_and_get_granule` macro which is not necessary in array-based GST
- Remove using `set_granule_raw()` function
- Remove using `set_granule` in GranuleStatusTable
- Remove using `set_granule_with_parent`
- Remove using `Inner` and `Rc` in Entry (two less layers)
- Remove using EntryGuard (one less layer)
- Change slightly the handling of `is_not_in_realm()`
- Replace using `vmsa::error::Error` with `crate::rmi::error::Error` (one less layer)
- The end value change of FVP_DRAM0_REGION and FVP_DRAM1_REGION (plus one from the original value)
- GranuleStatus's inner type change from u64 to u8


rmm/src/granule/array/entry.rs
```diff
@@ -0,0 +1,87 @@
...
+        unsafe { &*(addr as *const T) }
+    }
+
+    fn index(&self) -> usize {
+        let entry_size = core::mem::size_of::<Entry>();
+        let granule_size = core::mem::size_of::<Granule>();
+        //  XXX: is there a clever way of getting the Entry from Granule (e.g., container_of())?
+        //  [        Entry        ]
+        //  [  offset ] [ Granule ]
+        let granule_offset = entry_size - granule_size;
```
jinbpark, line 48: I like the technique that this `index()` does. (i.e., this technique helps to have only one field in `struct Entry()`, which looks so neat)
I agree with your comment "XXX", but I think it's enough to use the current technique in terms of security, as it doesn't use the "unsafe" keyword. (Though, if you figure out a better way, it would be great!)

rmm/src/granule/array/entry.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    pub fn set_state(&mut self, state: u8) -> Result<(), Error> {
+        if state == GranuleState::Delegated {
+            self.zeroize();
+        }
+        self.state = state;
+        Ok(())
+    }
+
+    pub fn content_mut<T: Content>(&mut self) -> &mut T {
```
jinbpark, line 32: I see that putting `content()`/`content_mut()`/`index()` into `Granule` can eliminate the need for `EntryGuard`.  I think this simplification (one less layer) can also be applied to the previous implementation (granule/page_table/*), which would be better.  I'll do this job in the future, as a refactoring task!

jinbpark:
I found that this PR (array-based GST) doesn't track references between granules. (e.g., Rec <-> Rd)

The page-table-based GST supports a child-parent relationship between two different granules (at the level of `struct Granule`), to track all references to a granule. It can prevent, for example, Rd from destroying when there is an associated Rec.

I think this reference counting doesn't have to be implemented via "Granule" (the page-table-based GST does) since it would involve more complexity in GST implementation, which is what you want to avoid.

Instead, you can use a simpler way, such as adding a field like `rec_count` into `struct Rd` to track references (this is not a general granule-level implementation, though).  I think this functionality (i.e., reference counting) doesn't have to be addressed in this PR.  But this should be addressed in a separate PR in the future-:

scripts/.config
```diff
@@ -16,7 +16,7 @@ bp.secure_memory=1
 bp.smsc_91c111.enabled=1
 bp.ve_sysregs.exit_on_shutdown=1
 cache_state_modelled=0
-cluster0.NUM_CORES=4
+cluster0.NUM_CORES=1
```
jinbpark, line 19: Is there any specific reason for this change?

zpzigi754:
> I found that this PR (array-based GST) doesn't track references between granules. (e.g., Rec <-> Rd)
> 
> The page-table-based GST supports a child-parent relationship between two different granules (at the level of `struct Granule`), to track all references to a granule. It can prevent, for example, Rd from destroying when there is an associated Rec.
> 
> I think this reference counting doesn't have to be implemented via "Granule" (the page-table-based GST does) since it would involve more complexity in GST implementation, which is what you want to avoid.
> 
> Instead, you can use a simpler way, such as adding a field like `rec_count` into `struct Rd` to track references (this is not a general granule-level implementation, though). I think this functionality (i.e., reference counting) doesn't have to be addressed in this PR. But this should be addressed in a separate PR in the future-:

Good to point it out! I also think reference-counting should be added for preventing the parent from being destroyed while it is in used. I will add it later :)

rmm/src/granule/array/entry.rs
zpzigi754: Right. I also think that we can remove or restrict the use of `EntryGuard` in the entire implementation, as `content()` and `content_mut()` will be used exclusively by granule. Please feel free to do it in the previous implementation :)

jinbpark (approved):

bitboom (approved):

rmm/src/granule/array/entry.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    pub fn content<T: Content>(&self) -> &T {
+        let addr = self.index_to_addr();
+        unsafe { &*(addr as *const T) }
+    }
+
+    fn index(&self) -> usize {
+        let entry_size = core::mem::size_of::<Entry>();
+        let granule_size = core::mem::size_of::<Granule>();
+        //  XXX: is there a clever way of getting the Entry from Granule (e.g., container_of())?
```
nook1208, line 45: I was also curious and looked it up, you can use [core::mem::offset_of](https://doc.rust-lang.org/nightly/core/mem/macro.offset_of.html) if you want :)

nook1208: Oh you mean that getting an entry from granule, sorry ;

nook1208 (approved):

rmm/src/granule/array/entry.rs
zpzigi754: Good to suggest using `core::more::offset_of`! In my previous trial however, it didn't work because `core::mem::offset_of` does not support the offset of private fields which `Mutex` and `UnsafeCell` contain.

PR #279, opened 2024-03-08
Use lazy_static for granule status table

zpzigi754:
This PR refactors granule status table's initialization code to use `lazy_static` macro, so that the accesses to it would not involve unsafe operations. As a result, 3 unsafe blocks have been removed. Also, some build errors of using the latest rustc (24-03-01) have been resolved.

```
(Unsafe-analysis results from cargo-geiger)

    x = safe code found in the crate
    y = total code found in the crate
    z = percentage of safe ratio as defined by x/y

   Functions         Expressions       Module
[before]
  122/132=92.42%  3814/4802=79.43%    islet_rmm 0.0.1
[after]
  121/131=92.37%  3811/4793=79.51%    islet_rmm 0.0.1
```

### DIFF
--- a/plat/fvp/src/entry.rs
+++ b/plat/fvp/src/entry.rs
@@ -2,6 +2,7 @@ use crate::allocator;
 use crate::log::LevelFilter;
 
 use armv9a::regs::*;
+use core::ptr::{addr_of, addr_of_mut};
 use islet_rmm::config::{NUM_OF_CPU, RMM_STACK_SIZE};
 use islet_rmm::io::stdout;
 use islet_rmm::logger;
@@ -76,12 +77,12 @@ unsafe fn init_mm() {
 unsafe fn setup() {
     static mut COLD_BOOT: bool = true;
 
-    if (&COLD_BOOT as *const bool).read_volatile() {
+    if (addr_of!(COLD_BOOT) as *const bool).read_volatile() {
         clear_bss();
         allocator::init();
         init_console();
         init_mm();
 
-        (&mut COLD_BOOT as *mut bool).write_volatile(false);
+        (addr_of_mut!(COLD_BOOT) as *mut bool).write_volatile(false);
     }
 }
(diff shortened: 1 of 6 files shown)


plat/fvp/src/entry.rs
```diff
@@ -76,12 +77,12 @@ unsafe fn init_mm() {
...
 
-    if (&COLD_BOOT as *const bool).read_volatile() {
+    if (addr_of!(COLD_BOOT) as *const bool).read_volatile() {
         clear_bss();
         allocator::init();
         init_console();
         init_mm();
 
-        (&mut COLD_BOOT as *mut bool).write_volatile(false);
+        (addr_of_mut!(COLD_BOOT) as *mut bool).write_volatile(false);
```
jinbpark, line 86: I didn't know `addr_of!()`.  It'd be better to use this macro from now on for getting pointers. Nice refactoring-:

jinbpark (approved):

bitboom (approved):

nook1208 (approved):

hihi-wang (approved):

hihi-wang:
Nice to remove `mut static`! 

Just bit curious here, is there any performance impact with this change? AFAIF, `lazy_static` executes code blocks when `GRANULE_STATUS_TABLE` is accessed for the first time only. So I guess this change may also enhance performance slightly, as we can remove call for `create_granule_status_table`

zpzigi754:
> Just bit curious here, is there any performance impact with this change? AFAIF, `lazy_static` executes code blocks when `GRANULE_STATUS_TABLE` is accessed for the first time only. So I guess this change may also enhance performance slightly, as we can remove call for `create_granule_status_table`

Good question! The effect of lazy evaluation in terms of performance seems controversial (reference: [lazy_evaluation_performance](https://en.wikipedia.org/wiki/Lazy_evaluation#Performance)). I guess that one slight advantage could have been obtained by the removal of the previous error checking routines (e.g., `if let Some(gst)`) which were repeatedly checked on runtime.

PR #285, opened 2024-03-14
rsi-el0: Remove the zero-padding in the display

zpzigi754:
This commit modifies the measurement length returned from linux-rsi module. RMM returns a zero-padded measurement value and linux-rsi module currently returns the maximum measurement length. Modifying the measurment length is necessary in order not to display the zero-padding in cli. The displayed format becomes consistent with the cross-platform-e2ee example's README.

```
(in realm's shell)

[before]
$ ./islet measurement-read --index 0
[  208.236303] rsi: device rsi open
[  208.269598] rsi: ioctl: measurement_read: 0
[  208.306435] rsi: device rsi released
"be33b6c8bf03e85fec35fdd3a3b856adbc61b525e901e59d701ca33a8c71408c0000000000000000000000000000000000000000000000000000000000000000"

[after]
$ ./islet measurement-read --index 0
[  206.805345] rsi: device rsi open
[  206.838067] rsi: ioctl: measurement_read: 0
[  206.917591] rsi: device rsi released
"be33b6c8bf03e85fec35fdd3a3b856adbc61b525e901e59d701ca33a8c71408c"
```

### DIFF
--- a/lib/rsi-el0/src/lib.rs
+++ b/lib/rsi-el0/src/lib.rs
@@ -50,6 +50,11 @@ pub fn measurement_read(index: u32) -> nix::Result<Vec<u8>> {
     let mut measure = [kernel::RsiMeasurement::new_empty(index)];
     let fd = Fd::wrap(nix::fcntl::open(DEV, FLAGS, MODE)?);
     kernel::measurement_read(fd.get(), &mut measure)?;
+    // TODO: remove the below condition after linux-rsi module returns
+    //       the correct measurement length
+    if measure[0].data[32..] == [0; 32] {
+        measure[0].data_len = 32;
+    }
     Ok(measure[0].data[..(measure[0].data_len as usize)].to_vec())
 }
 


jinbpark (approved):

nook1208 (approved):

bitboom (approved):

Havner:
I have a mixed feelings about this change. The length of the measurement is based on the hashing algorithm used and it might change. The RMM spec has 512 bits here for a reason and I don't feel that limiting the interface itself is the right solution. If you want to cut the measurement to the current length I feel that "post processing" in the user space (e.g. checking how many zeroes are at the end and cutting based on that) is the better and more generic solution than limiting the ioctl interface itself.

zpzigi754:
> I have a mixed feelings about this change. The length of the measurement is based on the hashing algorithm used and it might change. The RMM spec has 512 bits here for a reason and I don't feel that limiting the interface itself is the right solution. If you want to cut the measurement to the current length I feel that "post processing" in the user space (e.g. checking how many zeroes are at the end and cutting based on that) is the better and more generic solution than limiting the ioctl interface itself.

Good opinion! I also thought of processing the zero-padding either in the user space or kernel space. The reason I chose this direction is because `data_len` field in `struct rsi_measurement`, which is shared between user and kernel, is more natural to contain the currently-used size not the maximum size. Also, if the result of the hash calculation contain the zero values at the last x bytes coincidentally, post-processing the measurement value might not work correctly.

Havner:
To put it in other words you stance seems to be: we should do it (as if that is required somehow), we should just discuss how to do it best.
My opinion is: we shouldn't do it in a way that limits the standard, and if we can't do it that way (e.g. in userspace) we shouldn't do that at all.

Havner:
> Now, I understand what is your concern. If the hashing algorithm gets extended (a different hashing algorithm, like the algorithm of 48 bytes, is used by RMM), the kernel (rsi module) can return that value (48) by setting the value in the `data_len` field in `struct rsi_measurement`. In that case, the semantic of the `data_len` field will be consistent in the future. The change required then would be a simple switch statement for the different values of different algorithms.

1. You're only taking into consideration one specific RMM implementation. In case of the linux module it could be compiled once and used with different implementations as I wrote above (e.g. tf-rmm, islet. each with different hashing algorithm). In this case you cannot guarantee consistency in any way except doing what the standard does (full 512 bits).
2. In a place you want to do the switch (inside rsi.c) you don't even know the algorithm used so there is no data to make the switch from.

zpzigi754:
> > Now, I understand what is your concern. If the hashing algorithm gets extended (a different hashing algorithm, like the algorithm of 48 bytes, is used by RMM), the kernel (rsi module) can return that value (48) by setting 
> 1. You're only taking into consideration one specific RMM implementation. In case of the linux module it could be compiled once and used with different implementations as I wrote above (e.g. tf-rmm, islet. each with different hashing algorithm). In this case you cannot guarantee consistency in any way except doing what the standard does (full 512 bits).

That is not true. I agree that passing the full 512 bits from RMM to a realm OS is in the standard (spec). However, it is not the standard that passing that raw 512 bits from a realm OS to the requested app. The suggested direction in this PR can be actually applied to different RMM implementations regardless of islet or tf-rmm.

> 2. Not sure I understand that correctly, but it's the linux module that sets the `data_len`. And the linux module doesn't know the algorithm used (RSI measurement read doesn't contain it) so there is no data to feed the switch you mention.

According to the spec, there is a way for the linux module (a realm OS) to query and know the exact measurement length used. Please refer to 'RsiHashAlgorithm type' in the spec.

I agree when to pass the measurement length to user apps could be an interpretation issue. However, if we can free users from bothering to calculate some value without any compatability issue, why not hesitate using it?

hihi-wang (approved):

zpzigi754:
> How can you make it work (leaving point 2 for now) regardless of RMM implementation when you hard-code 32 bytes and then run that code on another implementation that uses SHA-512? You will cut the value in half.

I mean the switch statement based on the querying will be able to handle both of SHA-256 and SHA-512 regardless of RMM implementation. That's the reason I left a TODO comment. 
 
> Ok, they've added getting the Hash Algo in EAC5 (RSI_REALM_CONFIG), it wasn't there before. So if you get that information together with the measurement (in runtime) and use it than it's fine. But not hard coding it as it is now in this PR.

I didn't know that it is a new feature. It seems that `RsiHashAlgorithm type` has been introduced since EAC2. Then, there could be a slight backward-compatibility issue.

How about returning the current measurement size with the fixed value now and returning more diverse measurement sizes after EAC5 work is done?

Supporting EAC5 in islet is currently in progress and using a hard-coding seems inevitable either in user-space or kernel-space to resolve a discrepancy. Note that there is the discrepancy in the measurement format between cli execution on x64 and cli execution on aarch64.

Havner:
> I mean the switch statement based on the querying will be able to handle both of SHA-256 and SHA-512 regardless of RMM implementation. That's the reason I left a TODO comment.

Right, it will be able. It's not there at the moment in the state this PR is right now.

> I didn't know that it is a new feature. It seems that `RsiHashAlgorithm type` has been introduced since EAC2. Then, there could be a slight backward-compatibility issue.

Correct. What I should have said is that it wasn't there in BETA0. I haven't read any specs outside of BETA0/EAC5 so I don't know the state in between.

> How about returning the current measurement size with the fixed value now and returning more diverse measurement sizes after EAC5 work is done? Supporting EAC5 in islet is currently in progress and using a hard-coding seems inevitable either in user-space or kernel-space to resolve a discrepancy.

All I can say here is that we have to agree to disagree. For me the current state of the PR as it is now makes the situation worse, not better in any case. I would personally recommend to return to the situation the moment we can implement this properly (EAC5).

> Note that there is the discrepancy in the measurement format between cli execution on x64 and cli execution on aarch64.

Please elaborate, what discrepancy?
Also how can you get any measurements on x64? The whole CCA stack doesn't work on x64 as far as I know. Only the rsictl/islet tool does, but it's not utilizing any CCA related ioctls when run on x64.

Havner:
Another argument against this change as it is now (hard-coding measurement length) is that according to spec hash algorithm seems to be a runtime value (if I understand the spec properly), that is set with RMI_REALM_CREATE command (`RmiRealmParams`).

So it's not only that a compiled kernel will be run against different RMM implementations with different lengths. It can even be run against one implementation that will have a different measurement length for different realms. So hard-coding that on the client/realm side is even more incorrect.

zos:
By this logic, we shouldn’t allow the Host to create a Realm with any other
hash algorithm than sha-256, if we are treating this value as hardcoded as
default. Also in what you propose the hash_algo parameter will be useless,
when passed to RMI_REALM_CREATE.

Why not create a parameter to islet client which would override the length
of the measurement to be returned to user?

We are an open source project and I don’t understand why we are narrowing
down Islet implementation to only OUR specific use cases.

W dniu pt., 15.03.2024 o 16:01 Changho Choi ***@***.***>
napisał(a):

Havner:
> > Please elaborate, what discrepancy? Also how can you get any measurements on x64? The whole CCA stack doesn't work on x64 as far as I know. Only the rsictl/islet tool does, but it's not utilizing any CCA related ioctls when run on x64.
> 
> Please read and follow the instructions in https://github.com/islet-project/islet/blob/main/examples/cross-platform-e2ee/README.md.
> 
> The simulated measurement value on x64 can be obtained by
> 
> ```
> $ cd $(islet)/cli
> $ make x86_64
> $ ./islet measurement-read --index 0
> Simulated attestation operation on x86_64.
> "580bd77074f789f34841ea9920579ff29a59b9452b606f73811132b31c689da9"
> ```
> 
> It does not use ioctls, but use the same APIs in sdk. Also, it will display a different format for using the same APIs currently: one for 32 bytes and another for 64 bytes. That is the issue I received from another colleague. This symptom actually harms the user experience of islet and certifier framework.

My question was more out of curiosity. I don't know anything about that API but I don't see how that is relevant for this discussion as this discussion is exactly about the ioctl API which as you say yourself the X64 doesn't use (as it can't).

> My point is that the `data_len` field in `struct rsi_measurement` should be correctly used, otherwise that field is just useless and doesn't need to exist, because users would know in advance the maximum size to receive by the size of the byte array. 

I am the person that designed and implemented those ioctls. data_len is needed for one specific reason. Pointers in C don't know the length of the data the point to (contrary to Rust pointers). So every time you have a pointer you add a field that shows how much data is under that pointer. Simple as it is. There is no correct or incorrect way of usage here. It's a simple rule of thumb that data_len accompanies data pointer.

> Also, using the maximum value there could be totally misleading in terms of verbal meaning from the perspective of sdk. Those are the main targets of corrections I intended.

Fine. I get that. But by "correcting" something you created another issues I described above you don't relate to in this answer.

> Another point is that all the third party which islet depends on including tf-rmm, linux-cca, certifier framework, and kvmtool also currently use 32 bytes as the default measurement size in common use cases. Thus, I think that using a hard-coded value, for the time being, is totally fine during the progress of our project.

I'm sorry, but it's completely irrelevant what is above the ioctls here. It's irrelvant what and how they use the data. The ioctls need to be consistent with what is below, not above. If there is chance the data might be 64 bytes (and I proved beyond any doubt there is) the ioctls need to handle that. Your hard-coding 32 bytes value works against it.

> If you disagree with these advantages, namely better user experience and correct API usage, for the imaginary scenario, just leave it as it is.

Getting wrong data or cutting data in half and returning it only partially has no advantages. It's simply called a bug.

Havner:
> Another point is that all the third party which islet depends on (...) kvmtool also currently use 32 bytes as the default measurement size in common use cases.
> 
> (...) for the imaginary scenario

This is how imaginary is that scenario:

https://github.com/Havner/kvmtool-cca/blob/eac5/arm/aarch64/include/kvm/kvm-config-arch.h#L27

You can set the measurement algo in cmd line in runtime when running the realm with kvmtool. In runtime. While you propose to hardcode this in compile time.

Havner:
Also it seems that even Islet itself honors this parameter when doing measurements:

https://github.com/islet-project/islet/blob/main/rmm/src/measurement/ctx.rs#L20
https://github.com/islet-project/islet/blob/main/rmm/src/measurement/hash.rs#L60

zpzigi754:
@Havner, @zos, you are right. I've changed the commit so that the worried condition (incapability to support sha-512) would not happen.

PR #286, opened 2024-03-14
refactor: rsi: Fix to add padding instead of union

bitboom:
This PR refactors `rsi::HostCall`

- Deleted 4 unsafe blocks(1 method, 3 inner blocks)
- Added the width checking according to RMM spec

_Now `rsi::Hostcall` is also the target of Autopadding https://github.com/islet-project/islet/issues/276_


jinbpark (approved):

zpzigi754 (approved):
Good to remove unsafe blocks by the replacement of union type!

PR #287, opened 2024-03-20
Add auto-padding macro 

hihi-wang:
# What this pr for
- Introduce auto-padding macro, with automatic impl for `Default` trait as well.

# How to use
- Wrap with `pad_struct_and_impl_default` macro on type definition.

# Note
- Supports only integer primitives and arrays of it. In case of new type field is needed(which I believe there is no use of it though), you need to add type initialization matching on `type_check_and_init` macro. For instance, to introduce `f32`, add 

~~~
    (f32) => {
        0.0
    };
~~~

- Last arm of `type_check_and_init` matches with array type.
- Any naming suggestion is welcomed

### DIFF
--- a/lib/autopadding/Cargo.toml
+++ b/lib/autopadding/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "autopadding"
+version = "0.1.0"
+
+[dependencies]
+paste = "1.0.14"
+
+[features]
+checked = []
--- a/lib/autopadding/Cargo.toml
+++ b/lib/autopadding/Cargo.toml
@@ -1,6 +1,8 @@
 [package]
 name = "autopadding"
 version = "0.1.0"
+authors = ["Islet Contributors"]
+edition = "2021"
 
 [dependencies]
 paste = "1.0.14"
--- a/lib/autopadding/src/lib.rs
+++ b/lib/autopadding/src/lib.rs
@@ -1,11 +1,56 @@
 #![no_std]
 pub extern crate paste;
 
+// Unfortunately, [$t: ty; $size: expr] does not matches with array type...
+// instead, matches all integer primitves here.
 #[macro_export]
-macro_rules! pad_field {
+macro_rules! type_check_and_init {
+    (u8) => {
+        0
+    };
+    (i8) => {
+        0
+    };
+    (u16) => {
+        0
+    };
+    (i16) => {
+        0
+    };
+    (u32) => {
+        0
+    };
+    (i32) => {
+        0
+    };
+    (u64) => {
+        0
+    };
+    (i64) => {
+        0
+    };
+    (u128) => {
+        0
+    };
+    (i128) => {
+        0
+    };
+    (usize) => {
+        0
+    };
+    (isize) => {
+        0
+    };
+    ($t:ty) => {
+        [0; core::mem::size_of::<$t>()]
+    };
+}
+
+#[macro_export]
+macro_rules! pad_field_and_impl_default {
     // entry point.
     (@root $(#[$attr_struct:meta])* $vis:vis $name:ident { $($input:tt)* } ) => {
-        pad_field!(
+        pad_field_and_impl_default!(
             @munch (
                 $($input)*
             ) -> {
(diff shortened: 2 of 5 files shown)


lib/autopadding/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "autopadding"
+version = "0.1.0"
```
jinbpark, line 3: It would be better to add the following two lines as well,
as all the other packages under lib/ contains it too.

```
edition = "2021"
authors = ["Islet Contributors"]
```

jinbpark:
`pad_struct_and_impl_default` --> I like its name-:

rmm/src/rmi/realm/params.rs
```diff
@@ -5,46 +5,23 @@ use crate::measurement::Hashable;
 use crate::rmi::features;
 use crate::rmi::rtt::{RTT_PAGE_LEVEL, S2TTE_STRIDE};
 use crate::rmi::{HASH_ALGO_SHA256, HASH_ALGO_SHA512};
+use autopadding::*;
 
-const PADDING: [usize; 5] = [248, 767, 960, 6, 2020];
-
-#[repr(C)]
+pad_struct_and_impl_default!(
```
bokdeuk-jeong, line 10: I like your idea that macro also takes care of the default implementation. 👍

bokdeuk-jeong (approved):

jinbpark (approved):

lib/autopadding/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+                $vis struct $(#[$attr_struct])* $name
+            }
+        );
+    };
+
+    // TODO: Remove the zero-sized paddings added where no padding is required
+    (@guard ($current_offset:expr) -> {$(#[$attr:meta])* $vis:vis struct $name:ident $(($amount:expr, $vis_field:vis $id:ident: $ty:ty))*}) =>    {
+
+        $crate::paste::paste! {
+            #[repr(C)]
```
bitboom, line 66: I think this can be noted by developer with `attr_struct`. (not the forced way)

```rust
pad_struct_and_impl_default!(
   #[repr(C)]
    pub struct Params {
    }
}

File: rmm/src/rmi/realm/params.rs [hunk 4]
```diff
@@ -67,17 +44,17 @@ impl Hashable for Params {
     ) -> Result<(), crate::measurement::MeasurementError> {
         hasher.hash_fields_into(out, |alg| {
             alg.hash_u64(0); // features aren't used
-            alg.hash(self.padding0);
+            alg.hash(self._padfeatures_0);
```
bitboom, line 47: Could we use the origin name of the field? I think we should keep this for intuitive code.

```suggestion
            alg.hash(self.features_0);
```

bitboom (commented):
I really like auto-padding patterns :) This will make our code intuitive..

lib/autopadding/src/lib.rs
hihi-wang: I think this is needed, as rust compiler may reorganize order of fields if struct is not marked with `#[repr(C)]`

rmm/src/rmi/realm/params.rs
hihi-wang: I guess it's meaningless to hash padded fields here. Removed all padded fields here with new commit :)

lib/autopadding/src/lib.rs
bitboom: Oh i missed it. Thanks for the explanation!

bitboom (approved):

rmm/src/rmi/realm/params.rs
```diff
@@ -44,17 +44,12 @@ impl Hashable for Params {
     ) -> Result<(), crate::measurement::MeasurementError> {
         hasher.hash_fields_into(out, |alg| {
             alg.hash_u64(0); // features aren't used
-            alg.hash(self._padfeatures_0);
```
jinbpark: @hihi-wang 
I missed this last-minute change.

I think it'd be better to preserve the original hash calculation as it's meant to be used for attestation,
which means this run-time hash calculation must be paired with an offline tool that calculates measurement in advance. (the offline tool in ISLET is named `kvmtool-rim-measurer`).

So, it may cause a side effect, which is an attestation failure. (I'm not sure though)

PR #290, opened 2024-03-25
Draft safe-abstraction library to enhance safety of raw pointers

bitboom:
This PR introduces Safe Abstraction Crate.
- Following PRs:
  - [Fix structures, passed from realm, to use safe abstraction](https://github.com/islet-project/islet/pull/298)
  - [Fix structures, passed from host, to use safe abstraction](https://github.com/islet-project/islet/pull/297)
---

# Safe Abstraction Crate

The `safe_abstraction` crate is a library designed to facilitate safer abstraction over `unsafe` code.

Its primary goal is to enhance the safety of `unsafe` code by providing data structures and functions that minimize the need for direct `unsafe` code usage, and by offering traits for automating and explicitly marking parts of `unsafe` code that require developer intervention.

## Requirements
This crate is designed with the following requirements in mind to enhance safety and maintainability of Rust code, especially when dealing with `unsafe` code blocks. Here are the key requirements for using this crate: 

1. **Reduce the Absolute Amount of Unsafe Code**: The primary goal is to minimize the usage of `unsafe` code across the application. This reduction not only mitigates the potential for safety violations but also simplifies the codebase, making it easier to audit and maintain.

2. **Define Code-level Safety Rules for Unsafe Code**: Establish clear, enforceable safety rules that can be checked at the code level. These rules should guide the use of `unsafe` code and ensure that such code is used correctly and safely wherever necessary.

3. **Define Developer-guaranteed Safety Rules for Unsafe Code**: Beyond automated checks, there are safety guarantees that the developer must ensure when writing `unsafe` code. These rules require careful consideration and understanding of Rust's safety guarantees and how `unsafe` code might affect them.

4. **Provide an Interface for Safely Writing Unsafe Code When Requirements Are Met**: Once the safety rules defined in points 2 and 3 are satisfied, the crate should offer an interface that allows developers to write `unsafe` code in a manner that is considered safe. This interface acts as a bridge between the raw power of `unsafe` code and the safety guarantees Rust aims to uphold. 

5. **Ensure Compatibility with Unsafe Analysis Tools**: The implementation of the aforementioned interface and safety checks should not impede the effectiveness of unsafe analysis tools at the Mid-level Intermediate Representation (MIR) stage. These tools are crucial for optimizing Rust code and identifying potential safety violations. The crate's design and functionality should allow these tools to operate as intended, providing developers with insights into the safety and efficiency of their code. By adhering to these requirements, the crate aims to provide Rust developers with the tools they need to use `unsafe` code responsibly while maintaining the high safety standards that Rust is known for.

## Features

- **Encapsulation of Unsafe Code**: Offers a way to safely abstract `unsafe` operations, allowing for lower-level operations like memory access to be performed more safely.

- **Runtime Safety Checks**: Provides methods to perform crucial safety checks at runtime, such as verifying if a pointer is null and checking whether a pointer is properly aligned. These checks happen  when the methods are called during the execution of a program.

- **Compile-Time Type Safety Checks**: Enforces certain safety guarantees at compile time. For example, the use of Rust's type system can ensure that only pointers to types with known sizes are used, leveraging the `Sized` trait bound.

- **Developer-Driven Safety Assurance**: Introduces traits that allow developers to explicitly mark parts of `unsafe` code that still require manual safety guarantees, making it clear which parts of  the code need careful review.

## Key Concept

The core philosophy of this crate revolves around three data structures and one API that ensure safety when working with `unsafe` code. These data structures and API orchestrate a comprehensive safety assurance process, allowing for safer abstractions and interactions with potentially unsafe operations. Here's an overview of how these components interact:

- **`SafetyChecked` Trait**: Defines rules that can be checked at the code level. These checks aim to verify the basic safety requirements directly through code analysis.

- **`SafetyAssured` Trait**: Establishes rules that must be assured by the developer. These are higher-level safety guarantees
that cannot be fully automated and require developer intervention to ensure compliance.

- **`SafetyAssumed` Struct**: Provides safe abstraction, enabling the writing of code solely with safe constructs once the safety rules have been satisfied. This ensures that interactions with the underlying unsafe operations are safely abstracted,
maintaining the safety guarantees of Rust.

- **`assume_safe` API**: Acts as a gateway, verifying that all defined rules are met. If the verification process is successful,
it returns an instance of `SafetyAssumed`. The use of "Assume" signifies that complementary verification via formal methods or other MIR analysis tools might be necessary for comprehensive safety assurance.

```
+-----------------+   +-----------------+
|  SafetyChecked  |   |  SafetyAssured  |
+-----------------+   +-----------------+
            |                 |
            |                 |
            V                 v
+---------------------------------------+
|           assume_safe API             |
+---------------------------------------+
                   |
                   |
                   v
          +-----------------+
          |  SafetyAssumed  |
          +-----------------+
```

By implementing the `SafetyChecked` and `SafetyAssured` traits, developers can confidently utilize the `SafetyAssumed` structure to safely access instances that adhere to these rules, allowing for the safe encapsulation of `unsafe` code while maintaining compatibility with MIR analysis tools.

### Without Safe Abstraction
```rust
// This approach works, but consider the potential side effects
// if the implementation of the mock function is not verifiable,
// or if the developer uses `unsafe` solely for functionality
// without due consideration  for Memory Safety.
//
// Imagine the consequences of neglecting Memory Safety in pursuit of mere operation.
let addr = mock_get_addr_of_instance_from_external();

unsafe {
    let raw_ptr = &mut *(addr as *mut MyStruct);
    raw_ptr.set(0xABC);
    raw_ptr.print();
}
```
### With Safe Abstraction
```rust
// We can apply Safe Abstraction for accessing instances
// that have been checked and assured by three traits.
// This approach encapsulates unsafe code
// but still allows for analysis of the behavior at the MIR stage.
//
// Additioanlly, in client crates, the `#![forbid(unsafe_code)]` attribute can be used
// to prohibit the use of unsafe code.
let addr = mock_get_addr_of_instance_from_external();

let safety_assumed = assume_safe::<MyStruct>(addr).expect("Memory Safety Violation!");
safety_assumed.mut_with(|my_struct: &mut MyStruct| {
    my_struct.set(0xDEF);
    my_struct.print();
});
```

## Safety Rules about Raw Pointers
1. **Compile-time Size Determination**: The size of the structure the raw pointer points to must be determined at compile time, ensuring that memory operations are safe and predictable.

2. **Non-null**: The raw pointer must not be null, preventing dereference of null pointers

3. **Proper Alignment**: The address contained in the raw pointer must be properly aligned, ensuring that the memory access respects the hardware's alignment requirements.

4. **Initialization**: The instance pointed to by the raw pointer must be guaranteed to be properly initialized, avoiding undefined behavior due to uninitialized memory access.

5. **Ownership Rules Compliance**: The instance pointed to by the raw pointer must comply with Rust's ownership rules, preventing data races and ensuring safe memory access.

## Core logic
```rust
 pub fn assume_safe<T: SafetyChecked + SafetyAssured>(addr: usize) -> Option<SafetyAssumed> {
     let ptr = addr as *const T;
     // Safety: This cast from a raw pointer to a reference is considered safe
     //         because it is used solely for the purpose of verifying alignment and range,
     //         without actually dereferencing the pointer.
     let ref_ = unsafe { &*(ptr) };
     let checked = ref_.is_not_null() && ref_.is_aligned() && ref_.has_permission();
     let assured = ref_.initialized() && ref_.lifetime() && ref_.ownership();

     match checked && assured {
         true => Some(SafetyAssumed { addr }),
         false => None,
     }
 }
```

## TODO
- Check [Stack Borrow](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md)


lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+        /// to strengthen security and ensure that all
+        /// safety guarantees are thoroughly verified.
+        pub fn mut_with<T, F, R>(&self, mut f: F) -> R
+        where
+            T: SafetyChecked + SafetyAssured,
+            F: FnMut(&mut T) -> R,
+        {
+            unsafe {
+                let obj = T::as_mut(self.addr);
+                f(obj)
```
jinbpark, line 290: This is one of the key features of this PR, calling a closure in unsafe block for MIRI to be able to track into the body of the function.  Nice idea-:

rmm/src/rsi/mod.rs
```diff
@@ -72,21 +74,25 @@ pub fn do_host_call(
...
         .ok_or(Error::RmiErrorInput)?;
 
+    let safety_assumed = assume_safe::<HostCall>(pa.into()).ok_or(Error::RmiErrorInput)?;
+    let imm = safety_assumed.with(|host_call: &HostCall| host_call.imm());
+
     unsafe {
-        let host_call = HostCall::parse_mut(pa.into());
         if rec.host_call_pending() {
             for i in 0..HOST_CALL_NR_GPRS {
                 let val = run.entry_gpr(i)?;
```
jinbpark, line 83: After this safe abstraction is applied to "run", we can eliminate this big unsafe block. Am I correct??
e.g., `safety_assumed.mut_with(|run: &mut Run| run.entry_gpr(i))?;`

bitboom: > After this safe abstraction is applied to "run", we can eliminate this big unsafe block. Am I correct?? e.g., `safety_assumed.mut_with(|run: &mut Run| run.entry_gpr(i))?;`

Yes, that is the direction I am going for.
Although it is a bit different, I want to get rid of the big unsafe block even if additional checks or constraints are added.

rmm/src/rsi/hostcall.rs
```diff
@@ -48,3 +44,44 @@ impl core::fmt::Debug for HostCall {
             .finish()
     }
 }
+
```
jinbpark, line 47: In general, what PointerGuard/PointerMutGuard/EntryGuard do can be replaced with a set of traits introduced in this PR.

But currently, what PointerGuard wants to do is different from safe_abstraction: 
```
impl<'a, T: HostAccessor> PointerGuard<'a, T> {
    fn validate(&self) -> bool {
        let obj = unsafe { &*self.inner.ptr };   // 1. unsafe abstraction
        obj.validate()  // 2. per-struct based validation (it has nothing to with unsafe blocks. it means validation checks)
    }
}
```

1. unsafe abstraction --> it can be replaced by safe_abstraction.
2. validate() --> but it cannot be replaced by safe_abstraction, since safe_abstraction is designed for unsafe blocks.

It would be nice to implement another abstraction layer for validation checks having nothing to do with unsafe blocks,   when we try to replace PointerGuard/PointerMutGuard/EntryGuard with safe_abstraction.

jinbpark: Oh, I think this validation check can be performed within `initialized()` function. That seems what `initialized()` is designed for, according to your comment below -:
```
 /// For example, it should check if memory allocations have been made
 /// and if they are filled with appropriate values,
```

bitboom: Thank you for the detailed explanation. As you mentioned, while substituting `unsafe abstraction`, it seems necessary to also incorporate additional methods that fit the context, just like the original code.

bitboom:
@zpzigi754 Thank you for mentioning "how closures can hinder formal verification" by the offline feedback.

I researched further into this issue and have opted to use a combination of closures and macros in our implementation.
Considering the significance of this topic, I've documented my findings and approach [here](https://github.com/islet-project/islet/blob/raw-ptr-safety/lib/safe-abstraction/API_DESIGN.md).

If your formal verification work is not hindered by the use of closures right now,
would it be acceptable to proceed with the current closure method
while planning to implement the macro method at a later time?

This is to prevent the PR from getting too large.

lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        /// # Caution
+        /// It's important to remember that while this function aims
+        /// to provide a safer interface for interacting with `unsafe` code,
+        /// the inherent risks associated with `unsafe` code cannot be entirely eliminated.
+        /// Developers are encouraged to use `unsafe` analysis tools
+        /// to strengthen security and ensure that all
+        /// safety guarantees are thoroughly verified.
+        pub fn with<T, F, R>(&self, f: F) -> R
+        where
+            T: SafetyChecked + SafetyAssured,
```
zpzigi754, line 273: Really nice to use a trait bound!

rmm/src/rsi/hostcall.rs
```diff
@@ -48,3 +44,44 @@ impl core::fmt::Debug for HostCall {
...
+impl safe_abstraction::raw_ptr::RawPtr for HostCall {}
+
+impl safe_abstraction::raw_ptr::SafetyChecked for HostCall {
+    fn has_permission(&self) -> bool {
+        use safe_abstraction::raw_ptr::RawPtr;
+        let align_down = self.addr() & !(GRANULE_SIZE - 1);
+        get_granule_if!(align_down, GranuleState::Data).is_ok()
+    }
```
zpzigi754, line 55: This seems to be related to a granule invariant, and might be a bit different to the `read/write` permission in the description.

lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        fn initialized(&self) -> bool;
+
+        /// Validates the lifetime of the instance to ensure it is still valid.
+        ///
+        /// This method should confirm that the instance's lifetime
+        /// is still active and it hasn't been dropped or invalidated.
+        /// It's used to guard against dangling references or accessing freed memory,
+        /// which can lead to undefined behavior.
+        /// Ensuring a valid lifetime helps in maintaining memory safety across the application.
+        fn lifetime(&self) -> bool;
```
zpzigi754, line 185: I guess that checking this would be better to be done by `stacked borrow` instead of developer's manual marking.

zpzigi754:
> If your formal verification work is not hindered by the use of closures right now, would it be acceptable to proceed with the current closure method while planning to implement the macro method at a later time?

Thank you for the consideration. It would be okay to proceed with the current closure usage. I will find a way for the verification work (e.g., conditional build).

lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// Returns `Some(SafetyAssumed)` if all safety checks are satisfied,
+    /// or `None` if any of the checks fail, indicating that it is not safe to proceed.
+    pub fn assume_safe<T: SafetyChecked + SafetyAssured>(addr: usize) -> Option<SafetyAssumed> {
+        let ptr = addr as *const T;
+        // Safety: This cast from a raw pointer to a reference is considered safe
+        //         because it is used solely for the purpose of verifying alignment and range,
+        //         without actually dereferencing the pointer.
+        let ref_ = unsafe { &*(ptr) };
+        let checked = ref_.is_not_null() && ref_.is_aligned() && ref_.has_permission();
+        let assured = ref_.initialized() && ref_.lifetime() && ref_.ownership();
```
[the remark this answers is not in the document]
bitboom: I am drafting this crate as a general-purpose library for both enhancing unsafe code and supporting formal verification. When formal verification is complete, we can ensure that the trait always returns true, as shown below.

Compiler may optimize below code.

```rust
imp HostCall for SafetyChecked {
    fn is_not_null(&self) -> bool {
       // This is verified by XXX.
       true
    }
}

File: rmm/src/rsi/hostcall.rs (hunk 5 above)
Comment (bitboom) [reply rmm/src/rsi/hostcall.rs:55]: I apologize for the confusion. The "such as" keyword was missing in the README file. I will update it accordingly.
The original intention was like below as shown in lib.rs:

```rust
//! - `SafetyChecked`: Implements checks to ensure that the raw pointer is not null,  
//! properly aligned, and has the appropriate permissions for the intended memory operation.  

...  

4. **Appropriate Permissions**: The raw pointer must have the appropriate permissions  
//! for the operation, such as read, write, or execute permissions.  
```

lib/safe-abstraction/src/lib.rs
bitboom: This comment seems to be very helpful in developing that part. 👍 
Could I take note of that part separately and deal with it as the main topic later?

lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        pub fn with<T, F, R>(&self, f: F) -> R
+        where
+            T: SafetyChecked + SafetyAssured,
+            F: Fn(&T) -> R,
+        {
+            unsafe {
+                let obj = T::as_ref(self.addr);
+                f(obj)
+            }
+        }
```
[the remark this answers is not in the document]
bitboom: Oh.. I hadn't considered that aspect. If the developer is malicious, he doesn't need to do that because he just write the unsafe code. However, If the developer intend to be safe, this part should be consider any side effects or misuse like returning `&mut T`.

But not returning value makes code less intuitive. I'll leave it as is for now and take a closer look to see if there are any specific misuse issues.

lib/safe-abstraction/API_DESIGN.md
```diff
@@ -0,0 +1,123 @@
...
+
+2. **Performance Considerations**:
+The impact of using closures for abstraction
+on runtime performance should be evaluated.
+It’s crucial to ensure that the introduction of
+closures does not introduce significant overhead.
+
+3. **Formal Verification**:
+Using closures to abstract `unsafe` code can make formal verification
+more challenging is due to the anonymous nature of closures and their capturing behavior.
```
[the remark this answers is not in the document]
bitboom: Done :)

lib/safe-abstraction/src/lib.rs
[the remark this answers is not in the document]
bitboom: I noted it Closures Considerations section in `API_DESIGN.md` like below:

4. **Restricting Return Type**: Without restricting return type, considering that we need to address potential issues from the perspectives of API Misuse and Breaking Encapsulation.

lib/safe-abstraction/src/lib.rs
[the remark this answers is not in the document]
bitboom: I noted it as Implementation Guide in README .

--- 

## Implementation Guide
When implementing the SafetyChecked and SafetyAssured traits, we strongly recommend utilizing MIR analysis tools or formal verification tools to ensure the absence of safety issues within these methods.

Once the safety of these methods is guaranteed by such rigorous analysis, it's advised to make these methods always return true.

This approach not only solidifies the safety guarantees of your code but also aids in performance optimization. When the compiler recognizes that these methods consistently return true, it can perform optimizations by eliminating unnecessary checks or branches, thereby enhancing the runtime efficiency of your application.

Therefore, after thorough verification, set the body of these trait methods to unconditionally return true. This not only assures the safety of your code but also enables the compiler to optimize for better performance, ensuring a high level of efficiency without compromising safety.

lib/safe-abstraction/src/lib.rs
zpzigi754: Sure :)

lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        /// # Safety
+        ///
+        /// When calling this method, you have to ensure that all of the following is true:
+        ///
+        /// * The pointer must point to an initialized instance of `T`.
+        ///
+        /// * You must enforce Rust's aliasing rules
+        unsafe fn as_mut<'a, T: RawPtr>(addr: usize) -> &'a mut T {
+            &mut *(addr as *mut T)
+        }
```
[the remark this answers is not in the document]
bitboom: Thanks for noting it. That's the reason that these functions expose `unsafe`.

rmm/src/rsi/hostcall.rs
```diff
@@ -48,3 +44,44 @@ impl core::fmt::Debug for HostCall {
...
+}
+
+impl safe_abstraction::raw_ptr::SafetyAssured for HostCall {
+    fn initialized(&self) -> bool {
+        // This instance is initialized
```
hihi-wang, line 60: How about marking those comments with some keywords like `Guarantee` or others? [Just like `Safety` is used for unsafe](https://std-dev-guide.rust-lang.org/policy/safety-comments.html)

bitboom: It looks much better 👍 
I fixed it like below. Thanks a lot :)

```rust
fn initialized(&self) -> bool {
  // The initialization of this memory is guaranteed according to the RMM Specification A2.2.4 Granule Wiping.
  // ...
}

fn lifetime(&self) -> bool {
  // The instance's lifetime is guaranteed while being processed by the RMM.
  // ..
}

fn ownership(&self) -> bool {
  // The instance's ownership is guaranteed while being processed by the RMM.
  // ..
}

File: lib/safe-abstraction/src/lib.rs [hunk 12]
```diff
@@ -0,0 +1,315 @@
...
+        /// This method should check that the ownership
+        /// and borrowing rules of Rust are properly observed.
+        /// For instance, it could validate
+        /// that mutable references to the data structure are uniquely held,
+        /// or that ownership transfers (such as passing a value to another thread or function)
+        /// have been performed safely,
+        /// without violating Rust's guarantees about memory safety and concurrency.
+        /// Observing ownership rules is fundamental
+        /// in preventing data races and ensuring thread safety.
+        fn ownership(&self) -> bool;
```
jinbpark, line 198: For further clarification, I actually tried to come up with scenarios when "ownerhip should be false".

- For example, in the DATA_CREATE_UNKNOWN RMI, "ownership should be false" when dealing with the given raw pointer (e.g., page address), since RMM doesn't have to read/write a realm's data page.  The only thing RMM needs to do for this RMI is to map the page into RTT without data initialization. (or should this be done by has_permission()?)

Anyway, I think it would be helpful to think about when "ownership should be false" happens for further use cases.

bitboom: I agree with adding scenarios when the function should return false. It will very helpful to developers.

So i added like below:
```rust
        /// A return value of `true` signifies that the instance fully complies with all
        /// required permissions, including granule rules. Conversely, `false` must be returned
        /// in the following circumstances:
        /// - The instance violates the granule rules specified in the RMM Spec.
        /// - Any other detected permission violations that could potentially lead to unsafe
        ///   memory access patterns or data integrity issues.
        fn has_permission(&self) -> bool;

        /// However, it is crucial to return `false` under the following conditions:
        /// - The instance is derived from a raw pointer, and operates in a multi-core or multi-threaded
        ///   context without appropriate synchronization mechanisms (e.g., mutexes). This situation
        ///   significantly increases the risk of unsafe access patterns, such as data races or
        ///   simultaneous mutable aliasing, that violate Rust’s guarantees on memory safety.
        /// - Any other detected violation of ownership rules, such as incorrect lifecycle management
        ///   leading to potential dangling references or use-after-free scenarios.
        fn verify_ownership(&self) -> bool;
```

bitboom:
Changed: 
- Unify ownership() & lifetime() to verify_ownership()
- Add meet_extra_constraints() for post validation

zpzigi754:
How about removing `has_permission()`? It currently checks a granule's state and I think that checking granule's state is better to be done outside `has_permission`, because checking it is the responsibility of the spec (e.g., checking `Run` data structure's state is done by `rec enter` in spec). If the checking is done inside `has_permission()` again, it would either become duplicate or deviate from the spec in terms of the checking place.

bitboom:
> How about removing `has_permission()`? It currently checks a granule's state and I think that checking granule's state is better to be done outside `has_permission`, because checking it is the responsibility of the spec (e.g., checking `Run` data structure's state is done by `rec enter` in spec). If the checking is done inside `has_permission()` again, it would either become duplicate or deviate from the spec in terms of the checking place.

I agree with you. I removed it to handle outside. Thanks for the feedback!

zpzigi754 (approved):

lib/safe-abstraction/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The raw address of the target instance to be safely accessed.
+    ///
+    /// # Returns
+    ///
+    /// Returns `Some(SafetyAssumed)` if all safety checks are satisfied,
+    /// or `None` if any of the checks fail, indicating that it is not safe to proceed.
+    pub fn assume_safe<T: SafetyChecked + SafetyAssured>(addr: usize) -> Option<SafetyAssumed> {
```
nook1208, line 181: This function looks intuitive 👍 

One thing came into my mind that if 'assume_safe' is failed, the failure reason should be sent to the caller in some way.
Then, finally developers can do something to fix it.

bitboom: Thank you for the suggestion! I will add `Error Code` and change it to return `Result<SafetyAssumed, Error>`.
Could I make it after two followed PR merged?

nook1208: of course you can ! :)

jinbpark (approved):

PR #293, opened 2024-03-28
lib: autopadding: Fix to use array's len

bitboom:
Using core::mem::size_of couldn't get proper length when the element's type of array is not u8. (ex. [u64; 32])

### DIFF
--- a/lib/autopadding/src/lib.rs
+++ b/lib/autopadding/src/lib.rs
@@ -42,7 +42,7 @@ macro_rules! type_check_and_init {
         0
     };
     ($t:ty) => {
-        [0; core::mem::size_of::<$t>()]
+        [0; <$t>::LEN]
     };
 }
 
@@ -129,4 +129,12 @@ macro_rules! pad_struct_and_impl_default {
     ) => {
         $crate::pad_field_and_impl_default!(@root $(#[$attr])* $vis $name { $($fields)* } );
     };
+}
+
+pub trait ArrayLength {
+    const LEN: usize;
+}
+
+impl<T, const LENGTH: usize> ArrayLength for [T; LENGTH] {
+    const LEN: usize = LENGTH;
 }


hihi-wang:
Oh I assumed arrays are all bytes stream... didn't realized that there is other use cases. Thanks for let me know.

lib/autopadding/src/lib.rs
```diff
@@ -130,3 +130,11 @@ macro_rules! pad_struct_and_impl_default {
         $crate::pad_field_and_impl_default!(@root $(#[$attr])* $vis $name { $($fields)* } );
     };
 }
+
+pub trait ArrayLength {
+    const LEN: usize;
+}
+
+impl<T, const LENGTH: usize> ArrayLength for [T; LENGTH] {
+    const LEN: usize = LENGTH;
```
hihi-wang, line 139: Shouldn't this to be `LENGTH * core::mem::size_of<T>()`?

lib/autopadding/src/lib.rs
```diff
@@ -42,7 +42,7 @@ macro_rules! type_check_and_init {
         0
     };
     ($t:ty) => {
-        [0; core::mem::size_of::<$t>()]
+        [0; <$t>::LEN]
```
zpzigi754, line 45: Is this case exclusively used by an array type? If then, couldn't we use <$t>.len() instead?

lib/autopadding/src/lib.rs
bitboom: `LENGTH * core::mem::size_of<T>()` is the size of array. We need the length of elements to initialize array.

```rust
Self {
  // $id: type_check_and_init!($ty)
  arr: [0; 32]
}
```

lib/autopadding/src/lib.rs
bitboom: > Is this case exclusively used by an array type? 

Yes, right.

> If then, couldn't we use <$t>.len() instead?

`$t` is the type like `[u8; 32]`. So we couldn't call `len()`.

lib/autopadding/src/lib.rs
hihi-wang: Oh my bad :)

jinbpark (approved):

hihi-wang (approved):

PR #295, opened 2024-04-01
rmi: Apply autopadding to RmiRecRun

bitboom:
This PR eliminates all `union fields within the RMM, along with the associated unsafe code.
It becomes evident that a substantial amount of unsafe code was stemming from the use of union fields.


hihi-wang:
Just bit curious here, is `union` was used just for fixing data type size?

~~~
#[repr(C)]
union Flags {
    val: u64,
    reserved: [u8; 0x200],
}
~~~

For instance, `union` here was for fixing `Flags` to has `0x200` size, so that 0x200-64 would be filled with padding?

hihi-wang (approved):

bitboom:
> Just bit curious here, before this commit, is `union` was used just for fixing data type size?
> 
> ```
> #[repr(C)]
> union Flags {
>     val: u64,
>     reserved: [u8; 0x200],
> }
> ```
> 
> For instance, `union` here was for fixing `Flags` to has `0x200` size, so that 0x200-64 would be filled with padding?

Exactly! According to rmm spec B3.4.12 RmiRecEntry type, the offset of `gprs` is 0x200.

```rust
struct EntryInner {
    // 0x0
    flags: Flags, 
    // 0x200
    gprs: GPRs,
...
}
```

jinbpark (approved):

zpzigi754 (approved):
Good to remove many unsafe usages by not using union!

PR #301, opened 2024-04-08
Fix return values in rmi features

zpzigi754:
This PR fixes one correctness bug, found by [model checking](https://en.wikipedia.org/wiki/Model_checking) ([relevant PR](https://github.com/islet-project/islet/pull/234)). It also explicitly assigns a value of zero to X1 for better understanding of RMI features' behavior.

## Related spec

[B3.3.4.2 in beta0 and B4.3.4.2 in eac5]
```
The RMI_FEATURES command does not have any failure conditions.
```

[B3.3.4.3 in beta0 and B4.3.4.3 in eac5]
```
pre: index != 0
post: X1 == Zero()
```

### DIFF
--- a/rmm/src/rmi/features.rs
+++ b/rmm/src/rmi/features.rs
@@ -1,7 +1,6 @@
 use crate::event::Mainloop;
 use crate::listen;
 use crate::rmi;
-use crate::rmi::error::Error;
 
 extern crate alloc;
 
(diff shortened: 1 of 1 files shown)


bitboom (approved):
Really nice to apply the model checking 👍

nook1208 (approved):
Good find!

jinbpark (approved):

PR #303, opened 2024-04-14
ci: Add cargo vet to ensure that 3rd-party deps

bitboom:
This PR applies `cargo-vet` to CI

ref) https://mozilla.github.io/cargo-vet/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,12 @@ jobs:
     - name: Check pylint
       run: ./scripts/pylint.sh
 
+    - name: Check third-party Rust dependencies
+      run: |
+        cargo install --locked cargo-vet
+        cargo vet init
+        cargo vet
+
     - uses: actions-rs/audit-check@v1
       with:
         token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/ci.yml
```diff
@@ -143,10 +143,17 @@ jobs:
     - name: Check pylint
       run: ./scripts/pylint.sh
 
-    - uses: actions-rs/audit-check@v1
+    - name: Audit Cargo.lock against the advisory DB
```
zpzigi754, line 146: How about not using the name like the previoiusly or removing `Cargo.lock`?

Also, I think that putting this workflow at the end (e.g., after cargo-vet) would be better.

.github/workflows/ci.yml
```diff
@@ -143,10 +143,17 @@ jobs:
...
+      uses: actions-rs/audit-check@v1
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
 
+    - name: Ensure that third-party Rust dependencies
```
zpzigi754, line 151: How about changing the name as `Audit third-party rust dependencies` or `Check third-party rust dependencies`?

.github/workflows/ci.yml
bitboom: Applied to the former.

.github/workflows/ci.yml
bitboom: Fixed it to the latter.

.github/workflows/ci.yml
```diff
@@ -143,10 +143,17 @@ jobs:
...
+      run: |
+        cargo install --locked cargo-vet
```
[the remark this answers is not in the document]
bitboom: I referred [the installation's command](https://mozilla.github.io/cargo-vet/install.html).

zpzigi754 (approved):

PR #305, opened 2024-04-15
ci: Do not exit failure even if a keyword is not found

zpzigi754:
This PR temporarily fixes the below ci issue related to `Run unsafe analyzer` workflow. The keyword which the script is looking for seems to be missing for some reason.

```
Run ./scripts/unsafe-analyzer
...
[!] Running cargo geiger...
[!] Failed to run: cargo geiger --output-format Ratio @ /root/islet/plat/fvp
Failed to match (ignoring source) package: libc 0.2.153 (registry+https://github.com/rust-lang/crates.io-index)
Failed to match (ignoring source) package: libc 0.2.153 (registry+https://github.com/rust-lang/crates.io-index)
Failed to match (ignoring source) package: libc 0.2.153 (registry+https://github.com/rust-lang/crates.io-index)
...
```

### DIFF
--- a/scripts/unsafe-analyzer
+++ b/scripts/unsafe-analyzer
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 import os
 import subprocess
-import sys
+#import sys
 
 ROOT = os.path.realpath(os.path.join(os.path.dirname(__file__), ".."))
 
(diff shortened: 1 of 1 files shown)


bitboom (commented):
Could you please temporarily disable this CI job?

It appears that this issue is causing the stdout to increase for some reason. I will fix it soon.

zpzigi754:
> Could you please temporarily disable this CI job?
> 
> It appears that this issue is causing the stdout to increase for some reason. I will fix it soon.

How about using the current fix? 
It does not disable the workflow and allows to use the `unsafe-analyzer` with no stdout issue. 

If there is another issue in the current fix, I will disable the job as suggested.

bitboom (approved):
I am fine with the current fix too :)

PR #307, opened 2024-04-19
Do not miss content wiping in granule undelegate

zpzigi754:
This PR fixes a bug in `granule_undelegate`, found by model checking. Note that the previous page table-based GST didn't contain this bug.

[B.3.3.6.3 Success conditions in beta0 (B.4.3.6.3 in eac5)]
```
ID                 Condition
...
gran_content       Contents of target Granule are wiped.
```

### DIFF
--- a/rmm/src/granule/array/entry.rs
+++ b/rmm/src/granule/array/entry.rs
@@ -22,7 +22,10 @@ impl Granule {
     }
 
     pub fn set_state(&mut self, state: u8) -> Result<(), Error> {
-        if state == GranuleState::Delegated {
+        let prev = self.state;
+        if (prev == GranuleState::Delegated && state == GranuleState::Undelegated)
+            || (state == GranuleState::Delegated)
+        {
             self.zeroize();
         }
         self.state = state;


rmm/src/granule/array/entry.rs
```diff
@@ -22,7 +22,10 @@ impl Granule {
     }
 
     pub fn set_state(&mut self, state: u8) -> Result<(), Error> {
-        if state == GranuleState::Delegated {
+        let prev = self.state;
```
bokdeuk-jeong, line 25: I seems that existing implementation is already conforming the spec.

A2.2.4 Granule wiping (from eac5)

- **When the state of a Granule has transitioned from P to DELEGATED and then to any other state, any content associated with P has been wiped.**
- Any sequence of Granule state transitions which passes through the DELEGATED state causes the Granule contents to be wiped. This is necessary to ensure that information does not leak from one Realm to another, or from a Realm to the Host. Note that no agent can observe the contents of a Granule while its state is DELEGATED.
- Wiping is an operation which changes the observable value of a memory location from X to Y, such that the value X cannot be determined from the value Y.
- Wiping of a memory location does not reveal, directly or indirectly, any confidential Realm data.
- Wiping is not guaranteed to be implemented as zero filling.
- Realm software should not assume that the initial contents of uninitialized memory (that is, Realm IPA space which is backed by DATA Granules created using RMI_DATA_CREATE_UNKNOWN) are zero.

In my understanding, UNDELEGATE always has to go though DELEGATE meaning that when a granule is requested to be undelegated, it is already wiped. 

Could you explain more when current implementation doesn't follow the spec?

zpzigi754: Good comment! The bug will only be triggered with the condition when a delegated granule contains stale contents and this actually depends on the implementation.

Given our implementation, suppose that `RMI_DATA_CREATE` is called with some invalid input which successfully copies the contents to the delegated page (stale contents), yet fails to set that granule as data granule (e.g., returning an error). In this case, the delegated page would have contents which should be wiped out. 

In addition to this specific place, there could be other similar places where stale contents have not been handled properly in the delegated state. In this case, the added patch may help to wipe out the contents conservatively.

bokdeuk-jeong (approved):

bitboom (approved):

jinbpark (approved):

nook1208 (approved):

PR #308, opened 2024-04-21
ci: Fix the error related with installing packages

bitboom:
This PR fixes [the CI error](https://github.com/islet-project/islet/actions/runs/8776046268/job/24079464476).

- Apply `sudo apt update` with `--fix-missing` which addressed issues caused by missing packages sources
- Fix to use `apt-get` which provides a stable CLI interface for automated environments

### DIFF
--- a/scripts/deps/pkgs.sh
+++ b/scripts/deps/pkgs.sh
@@ -2,9 +2,11 @@
 
 set -e
 
-apt update && apt -y install sudo
+apt-get update && apt -y install sudo
 
-sudo apt install -y -qq --no-install-recommends \
+sudo apt-get update
+
+sudo apt-get install -y -qq --no-install-recommends --fix-missing \
 	gcc-multilib g++-multilib cmake libssl-dev \
 	binutils python3-pip \
 	device-tree-compiler xterm fakeroot mtools fdisk cpio \


nook1208 (approved):
good to fix !

scripts/deps/pkgs.sh
```diff
@@ -2,9 +2,11 @@
 
 set -e
 
-apt update && apt -y install sudo
+apt-get update && apt -y install sudo
 
-sudo apt install -y -qq --no-install-recommends \
+sudo apt-get update
```
zpzigi754, line 7: I guess that the later `apt-get update` can be removed, as they are duplicate.

hihi-wang (approved):

scripts/deps/pkgs.sh
bitboom: > I guess that the one `apt-get update` can be removed, as they are duplicat

I remember this line was added to install `sudo`. Do you want to try without `apt-get update`?

zpzigi754: Right. The first `apt-get update` would be necessary for the installing `sudo`. I mean the second `sudo apt-get update` can be removed if there is no additional purpose.

bitboom: `apt-get update` makes below errors:

```
W: Problem unlinking the file /var/cache/apt/pkgcache.bin - RemoveCaches (13: Permission denied)
W: Problem unlinking the file /var/cache/apt/srcpkgcache.bin - RemoveCaches (13: Permission denied)
```

Also only adding `--fix-missing` without `sudo apt-get update` couldn't resolve this. Please refer [here](https://github.com/islet-project/islet/actions/runs/8776299138)

zpzigi754: I see. Thank you for the answer!

zpzigi754 (approved):

PR #311, opened 2024-04-24
safe-abstraction: Add Error Code

bitboom:
This PR adds `Error` to `safe-abstraction`. 

- Reflected [review](https://github.com/islet-project/islet/pull/290#discussion_r1552693060)
- Followed PR: [Fix structures, passed from host, to use safe abstraction](https://github.com/islet-project/islet/pull/297)

### DIFF
--- a/rmm/src/host/mod.rs
+++ b/rmm/src/host/mod.rs
@@ -14,7 +14,14 @@ pub fn copy_from<T: SafetyChecked + SafetyAssured + Copy>(addr: usize) -> Option
     PageTable::get_ref().map(addr, false);
     let ret = assume_safe::<T>(addr).map(|safety_assumed| *safety_assumed);
     PageTable::get_ref().unmap(addr);
-    ret
+
+    match ret {
+        Ok(obj) => Some(obj),
+        Err(err) => {
+            error!("Failed to convert a raw pointer to the struct. {:?}", err);
+            None
+        }
+    }
 }
 
 pub fn copy_to<T: SafetyChecked + SafetyAssured + Copy>(src: &T, dst: usize) -> Option<()> {
@@ -25,7 +32,14 @@ pub fn copy_to<T: SafetyChecked + SafetyAssured + Copy>(src: &T, dst: usize) -> 
     PageTable::get_ref().map(dst, false);
     let ret = assume_safe::<T>(dst).map(|mut safety_assumed| *safety_assumed = *src);
     PageTable::get_ref().unmap(dst);
-    ret
+
+    match ret {
+        Ok(_) => Some(()),
+        Err(err) => {
+            error!("Failed to convert a raw pointer to the struct. {:?}", err);
+            None
+        }
+    }
 }
 
 /// DataPage is used to convey realm data from host to realm.
--- a/rmm/src/rmi/error.rs
+++ b/rmm/src/rmi/error.rs
@@ -1,5 +1,7 @@
 use crate::{measurement::MeasurementError, rsi};
 
+use safe_abstraction::raw_ptr;
+
 // B3.4.1 RmiCommandReturnCode type
 // Default index is 0
 #[derive(Debug)]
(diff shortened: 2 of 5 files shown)


rmm/src/host/mod.rs
```diff
@@ -1,31 +1,32 @@
...
     if !validate_addr(dst) || get_granule_if!(dst, GranuleState::Undelegated).is_err() {
-        return None;
+        return Err(Error::RmiErrorInput);
     }
 
     PageTable::get_ref().map(dst, false);
     let ret = assume_safe::<T>(dst).map(|mut safety_assumed| *safety_assumed = *src);
     PageTable::get_ref().unmap(dst);
-    ret
+    Ok(ret?)
```
hihi-wang, line 29: I guess caller still cannot distinguish `SafetyCheckFailed` and `AssuranceCheckFailed`, as both will converted to `RmiErrorInput` by `From` impl

rmm/src/rmi/realm/mod.rs
```diff
@@ -53,7 +53,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let rd_obj = rd_granule.content_mut::<Rd>();
         rmm.page_table.map(rd, true);
 
-        let params = host::copy_from::<Params>(params_ptr).ok_or(Error::RmiErrorInput)?;
+        let params = host::copy_from::<Params>(params_ptr)?;
```
zpzigi754, line 56: I think that the previous one might be better to tell the error type, as it is mentioned explicitly. 

The comment can be applied to other similar places.

rmm/src/host/mod.rs
bitboom: I added it to log message when converting it. Thanks!

rmm/src/rmi/realm/mod.rs
bitboom: I restored it to `host::copy_*` functions. In other cases, called `assume_safe` directly, i uses `?` operation to print error message.

rmm/src/host/mod.rs
hihi-wang: hmm... I think logging while on converting Error type is not idiomatic, but unfortunately, it seems inevitable if we cannot add error type on `rmi::error::Error`...

rmm/src/rmi/realm/mod.rs
zpzigi754: Adding a log in `? operation` might help to determine the error by looking at the log, but what I meant was that knowing it directly by reading the code, just the line causing a specific error, might be better.

rmm/src/host/mod.rs
bitboom: As you mentioned, we inevitably have to return the RMI error codes defined in the spec.

nook1208 (approved):

rmm/src/rmi/error.rs
```diff
@@ -59,3 +61,10 @@ impl From<rsi::error::Error> for Error {
         }
     }
 }
+
+impl From<raw_ptr::Error> for Error {
+    fn from(error: raw_ptr::Error) -> Self {
+        error!("Failed to convert a raw pointer to the struct. {:?}", error);
+        Error::RmiErrorInput
+    }
+}
```
zpzigi754, line 70: How about removing this conversion?

bitboom: Removing this part would negate the purpose of this PR. The PR logs potential error information when casting a raw pointer to a specific structure, in a situation where we must return RmiErrorInput. To log individually, we would have to log errors for about 30 different structure conversion statements.

If you think additional error information is not needed, it might be better to close this PR.

zpzigi754: I see. Then, it might be better to keep it. BTW, I'm actually working on removing our internal errors (e.g., NotExistRealm) not defined in the spec. So, the two directions are somewhat conflicting. Anyway, thank you for the detailed explanation!

hihi-wang (approved):

rmm/src/rmi/error.rs
bitboom: > BTW, I'm actually working on removing our internal errors (e.g., NotExistRealm) not defined in the spec.

Oh, I got your point now. Thanks for the explanation. 👍 

I also believe that unnecessary errors that are not even used as logging information should be eliminated. However, at the same time, it seems that the method of outputting necessary information for error handling of our internal logic, including errors of third-party libraries, should be applied together.

One of them is to print error logging when converting an error code as the current PR does. This is not an idle method, but I believe it will be used for debugging purposes before a better method comes out.

For now, why don't we merge this PR and discuss if there is any better way?

zpzigi754: Sure. Merging it would be fine :)

zpzigi754 (approved):

PR #312, opened 2024-04-24
Integrate realm and rd

zpzigi754:
This PR contains the refactoring commits related to the earlier discussion (#230). Realm and Rd (Realm Descriptor) have almost the same purpose. It aims to keep only Rd which is defined by spec.

## Summary of changes
```
-pub struct Realm {
-   id: usize,
-   pub vmid: u16,
-   pub state: State,
-   pub vcpus: Vec<Arc<Mutex<VCPU>>>,
-   pub page_table: Arc<Mutex<Box<dyn IPATranslation>>>,
-   pub measurements: [Measurement; MEASUREMENTS_SLOT_NR],
-}

pub struct Rd {
-   realm_id: usize,
+   vmid: u16,
    ...
+   s2_table: Arc<Mutex<Box<dyn IPATranslation>>>,
    hash_algo: u8,
+   pub measurements: [Measurement; MEASUREMENTS_SLOT_NR],
+   pub vcpus: Vec<Arc<Mutex<VCPU>>>,
}
```
## Notes
- `vmid` is integrated with `id` so that only `vmid`, passed by the Host, is used
- `VMID_SET` is used instead of the previous `RMS` (Realm Map Struct?)
- `trait Context` and redundant `enum State` have been removed

## Changes in the caller side
```
[before]
   let (s2tte, last_level) = get_realm(realm_id)
          .ok_or(Error::RmiErrorOthers(NotExistRealm))?
          .lock()
          .page_table
          .lock()
          .ipa_to_pte(GuestPhysAddr::from(ipa), level)
          .ok_or(error_code)?;

[after]
   let (s2tte, last_level) = rd
          .s2_table()
          .lock()
          .ipa_to_pte(GuestPhysAddr::from(ipa), level)
          .ok_or(error_code)?;
```

```
[before]
   let measurements = crate::realm::registry::get_realm(realmid)
        .ok_or(Error::RmiErrorOthers(NotExistRealm))?
        .lock()
        .measurements;

[after]
   let measurements = rd.measurements;
```

## A caution
- Two data aborts are newly generated during realm creation for some reasons which will be analyzed and fixed later

### DIFF
--- a/rmm/src/realm/vcpu.rs
+++ b/rmm/src/realm/vcpu.rs
@@ -1,9 +1,6 @@
-use super::Realm;
-
 use crate::gic;
 use crate::realm::context::Context;
-use crate::realm::registry::get_realm;
-use crate::realm::registry::RMS;
+use crate::realm::registry::VMID_SET;
 use crate::realm::timer;
 use crate::rmi::error::Error;
 use crate::rmi::realm::Rd;
@@ -20,15 +17,13 @@ pub struct VCPU {
     pub context: Context,
     pub state: State,
     pub pcpu: Option<usize>,
-    pub realm: Arc<Mutex<Realm>>, // Realm struct the VCPU belongs to
     me: Weak<Mutex<Self>>,
 }
 
 impl VCPU {
-    pub fn new(realm: Arc<Mutex<Realm>>) -> Arc<Mutex<Self>> {
+    pub fn new() -> Arc<Mutex<Self>> {
         Arc::<Mutex<Self>>::new_cyclic(|me| {
             Mutex::new(Self {
-                realm,
                 me: me.clone(),
                 state: State::Ready,
                 context: Context::new(),
@@ -58,7 +53,8 @@ impl VCPU {
     }
 
     pub fn is_realm_dead(&self) -> bool {
-        Arc::strong_count(&self.realm) == 0
+        // XXX: is this function necessary?
+        false
     }
 }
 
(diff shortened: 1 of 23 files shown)


jinbpark:
This refactoring is what I have been wanting to see! Thanks for bringing this up :)

bokdeuk-jeong (approved):
It is great to have a streamlined realm management. 
Do you think it is better to leave the definition of realm, Rd, inside rmi? I think it is more appropriate to place it inside realm.

zpzigi754:
> Do you think it is better to leave the definition of realm, Rd, inside rmi? I think it is more appropriate to place it inside realm.

Good question! I also think that most of the data structures including `realm` and `Rd` are better to be placed outside the current rmi module.

bitboom (approved):
Looks nice! Do you have plan to integrate vcpu and rec too?

zpzigi754:
> Do you have plan to integrate vcpu and rec too?

Partly yes, but not in the near time. I've refactored `Realm` in order to help verify `Realm`-related RMIs. So, the integration of `vcpu` and `rec` would be required before verifying `Rec`-related RMIs. The refactoring task itself could be done by another.

rmm/src/realm/vcpu.rs
```diff
@@ -87,31 +71,33 @@ pub enum State {
...
 }
 
 pub fn remove(id: usize) -> Result<(), Error> {
-    RMS.lock().1.remove(&id).ok_or(Error::RmiErrorInput)?;
+    VMID_SET
+        .lock()
+        .remove(&id)
+        .then_some(())
+        .ok_or(Error::RmiErrorInput)?;
     Ok(())
```
bitboom, line 102: [minor] I guess it's enough with `ok_or`

```suggestion
        .ok_or(Error::RmiErrorInput)
```

zpzigi754: Thank you for the suggestion! I've reflected the suggestion :)

PR #314, opened 2024-04-29
stat: Fix to use atomic operations

bitboom:
This PR aims to eliminate `unsafe code` associated with Mutable Static Variables (MSVs).

- I have investigated 10 variables used as MSV. 
  - VMID_SET, RMM_PAGE_TABLE, LOGGER, COLLECTED_STAT_CNT, STATS
  - RMM_SHARED_BUFFER_LOCK, REALM_ATTEST_KEY, PLAT_TOKEN, GRANULE_STATUS_TABLE, STDOUT
- All accesses to MSVs have been made safe with this PR

The strategy for accessing MSVs is outlined below.
1. Immutable Variables: Directly declare static variables for immutable data, requiring no synchronization.
2. Primitive Types: Use `core::sync::atomic` for primitive types, enabling atomic operations that ensure thread safety.
3. Dynamic Initialized Types: Implement `lazy_static` for types requiring runtime initialization, ensuring thread-safe setup.
4. Synchronization Mechanisms: Choose between `SpinLock` and `Mutex` based on your system requirements

### DIFF
--- a/rmm/src/stat.rs
+++ b/rmm/src/stat.rs
@@ -1,6 +1,8 @@
 use crate::allocator;
 use crate::rmi;
 use crate::rsi;
+
+use core::sync::atomic::{AtomicUsize, Ordering};
 use spin::mutex::Mutex;
 
 //NOTE: RMI, RSI_CMD_MAX are should be updated whenever there is a new command
@@ -18,7 +20,7 @@ const RSI_CMD_CNT: usize = RSI_CMD_MAX - RSI_CMD_MIN + 1;
 const MAX_CMD_CNT: usize = max(RMI_CMD_CNT, RSI_CMD_CNT);
 const MAX_KIND: usize = Kind::Undefined as usize;
 
-static mut COLLECTED_STAT_CNT: u64 = 0;
+static COLLECTED_STAT_CNT: AtomicUsize = AtomicUsize::new(0);
 
 const fn max(a: usize, b: usize) -> usize {
     if a >= b {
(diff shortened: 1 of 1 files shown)


rmm/src/stat.rs
```diff
@@ -18,7 +20,7 @@ const RSI_CMD_CNT: usize = RSI_CMD_MAX - RSI_CMD_MIN + 1;
 const MAX_CMD_CNT: usize = max(RMI_CMD_CNT, RSI_CMD_CNT);
 const MAX_KIND: usize = Kind::Undefined as usize;
 
-static mut COLLECTED_STAT_CNT: u64 = 0;
+static COLLECTED_STAT_CNT: AtomicUsize = AtomicUsize::new(0);
```
hihi-wang, line 23: Is `AtomicU64` is not supported on our target?

bitboom: Our target supports `AtomicU64` as well. Since `u64` and `usize` are the same size on our main target and we use `usize` in other variables, like `MAX_CMD_CNT: usize`, i switched to using usize for consistency.

rmm/src/stat.rs
```diff
@@ -95,12 +97,10 @@ impl Stats {
             return Err(Error::IntegerOverflow);
         }
 
-        unsafe {
-            COLLECTED_STAT_CNT = COLLECTED_STAT_CNT.wrapping_add(1);
-            if COLLECTED_STAT_CNT % 10 == 0 {
-                self.print();
-            }
-        };
+        COLLECTED_STAT_CNT.fetch_add(1, Ordering::SeqCst);
```
jinbpark, line 100: I think `COLLECTED_STAT_CNT`  should be declared as a member field of `struct Stats`.

`STATS` is already protected by Mutex as below.
```
lazy_static! {
    pub static ref STATS: Mutex<Stats> = Mutex::new(Stats::new());
}
```

And, `COLLECTED_STAT_CNT`  is currently used only in `STATS` so is not needed to be declared as a separate global variable.

@nook1208 :  Could you confirm this?

nook1208: Oh, good point!
I agree with you. I'll include it to the Stats structure 
Thank you !

nook1208: @bitboom 
I raise #318 PR to fix it. 
I think we can remove `COLLECTED_STAT_CNT` 
Is it okay with you :) ?

bitboom: @nook1208 Absolutely! I think it's the better solution. I will close this PR. Thanks both @jinbpark and @nook1208  :)

bitboom:
Replaced with https://github.com/islet-project/islet/pull/314

PR #315, opened 2024-05-02
Add Cargo.lock for dependency consistency

bitboom:
This commit adds the Cargo.lock file to ensure that the exact versions of dependencies are locked down across all environments.

This is crucial for maintaining consistent behavior of our application in development, testing, and production, reducing the `works on my machine` issues by aligning the dependency versions used by all developers and in all deployments.

ref) https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html


zpzigi754:
After reading the reference, I am still not sure whether adding `Cargo.lock` is a desirable policy in our case, as previously it was recommended not to add `Cargo.lock` for the libraries (maybe the Cargo's general recommendation can change again). 

The change would be summarized as the below. I think that testing with newer dependencies is a good feature which is currently done by CI (and can be reproduced locally with `cargo update`), but can be gone with the change.
 
[before this PR]
- CI will check with the newest versions in the dependencies
- `Cargo.lock` doesn't have to be maintained

[after this PR]
- CI will check with the same old versions in the dependencies
- `Cargo.lock` needs to be maintained which can be another burden

If what we want is fixing the versions of certain third-party crates, another option we can use would be modifying `Cargo.toml`.

bitboom:
> After reading the reference, I am still not sure whether adding `Cargo.lock` is a desirable policy in our case, as previously it was recommended not to add `Cargo.lock` for the libraries (maybe the Cargo's general recommendation can change again).
> 
> The change would be summarized as the below. I think that testing with newer dependencies is a good feature which is currently done by CI (and can be reproduced locally with `cargo update`), but can be gone with the change.
> 
> [before this PR]
> 
> * CI will check with the newest versions in the dependencies
> * `Cargo.lock` doesn't have to be maintained
> 
> [after this PR]
> 
> * CI will check with the same old versions in the dependencies
> * `Cargo.lock` needs to be maintained which can be another burden
> 
> If what we want is fixing the versions of certain third-party crates, another option we can use would be modifying `Cargo.toml`.

We released the SDK working on the certifier as a tag [certifier-v1.0-beta](https://github.com/islet-project/islet/releases/tag/certifier-v1.0-beta).
However, our code has not changed, due to an update from a third party (zeroize), the previously released version did not work. 
- issue : https://github.com/ccc-certifier-framework/certifier-framework-for-confidential-computing/issues/244
- summary: error: package `zeroize v1.8.0` cannot be built because it requires rustc 1.72 or newer

Either way, we have to stop this.
Modifying Cargo.toml is good for fixing the version we need, but it cannot solve this problem.

zpzigi754 (approved):
If `Cargo.lock` would help to pinpoint the issue and others, adopting it could be beneficial.

PR #317, opened 2024-05-03
Add harnesses for verifying rmi granule delegate and undelegate

zpzigi754:
This PR adds two verification harnesses for model checking: RMI_GRANULE_DELEGATE and RMI_GRANULE_UNDELEGATE.

It found one correctness bug (https://github.com/islet-project/islet/pull/307). 

## How to use?
```
(in islet/model-checking)

$ RMI_TARGET=rmi_granule_(un)delegate make verify 

...
SUMMARY:
 ** 0 of 3055 failed (81 unreachable)

 ** 14 of 14 cover properties satisfied
```

## Summary of changes
- Add a ghost gpt field in granule entry for tracking Granule Protection Table (GPT)'s change
- Add `is_valid` predicate for ensuring Granule Status Table (GST)'s valid status
- Put common helper functions in `common` module (`pre_/post_granule_state()`, `pre_/post_granule_gpt()`, etc)
- Add host memory and EL3 gpt change modeling code to help harnesses interact with their environment

## About host memory modeling
This models the Host memory as a pre-allocated memory region which can avoid a false positive related to invalid memory accesses in model checking. Also, instead of using the same starting address (e.g., 0x8000_0000), it uses a mock region filled with non-deterministic contents. This approach helps to address an issue related to the backend CBMC's pointer encoding, as 0x8000_0000 cannot be distinguished from null pointer in CBMC.

## Invariants with `is_valid()` predicate
The added `is_valid()` predicate can be used to ensure that all granule entries in GST are in valid status conforminig to the spec A2.2.1.

```
A2.2.1 Granule attributes in beta0-eac4

If the state of a Granule is not UNDELEGATED then the PAS of the Granule is REALM.
If the state of a Granule is UNDELEGATED then the PAS of the Granule is not REALM.
```

### DIFF
--- a/model-checking/src/rmi_granule_delegate.rs
+++ b/model-checking/src/rmi_granule_delegate.rs
@@ -1,0 +1,119 @@
+use crate::common::addr_is_granule_aligned;
+use crate::common::{post_granule_gpt, post_granule_state, pre_granule_gpt, pre_granule_state};
+use crate::get_granule;
+use islet_rmm::granule::entry::GranuleGpt;
+use islet_rmm::granule::validate_addr;
+use islet_rmm::granule::GranuleState;
+use islet_rmm::monitor::Monitor;
+use islet_rmm::rmi;
+use islet_rmm::rmi::error::Error;
+
+#[kani::proof]
+#[kani::unwind(9)]
+fn verify_granule_delegate() {
+    // Initialize registers (symbolic input)
+    let regs: [usize; 8] = kani::any();
+    kani::assume(regs[0] == rmi::GRANULE_DELEGATE);
+    // TODO: check the below again
+    let addr = regs[1];
+
+    // Pre-conditions
+    let failure_gran_align_pre = !addr_is_granule_aligned(addr);
+    let failure_gran_bound_pre = !validate_addr(addr);
+    let failure_gran_state_pre = pre_granule_state(addr) != GranuleState::Undelegated;
+    let failure_gran_gpt_pre = pre_granule_gpt(addr) != GranuleGpt::GPT_NS;
+    let no_failures_pre = !failure_gran_align_pre
+        && !failure_gran_bound_pre
+        && !failure_gran_state_pre
+        && !failure_gran_gpt_pre;
+
+    // Execute command and read the result.
+    let out = Monitor::new().run(regs);
+    let result = out[0];
+
+    // Post conditions have been moved to be inside if statements
+    // to constrain the conditions in order not to touch the panic in unwrap();
+
+    // Failure condition assertions
+    let prop_failure_gran_align_ante = failure_gran_align_pre;
+
+    kani::cover!();
+    if prop_failure_gran_align_ante {
+        let failure_gran_align_post = result == Error::RmiErrorInput.into();
+        let prop_failure_gran_align_cons = failure_gran_align_post;
+
+        kani::cover!();
+        assert!(prop_failure_gran_align_cons);
+    }
+
+    let prop_failure_gran_bound_ante = !failure_gran_align_pre && failure_gran_bound_pre;
+
+    kani::cover!();
+    if prop_failure_gran_bound_ante {
+        let failure_gran_bound_post = result == Error::RmiErrorInput.into();
+        let prop_failure_gran_bound_cons = failure_gran_bound_post;
+
+        kani::cover!();
+        assert!(prop_failure_gran_bound_cons);
+    }
+
+    let prop_failure_gran_state_ante =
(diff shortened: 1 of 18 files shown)


jinbpark:
How about putting `$ RMI_TARGET=rmi_granule_(un)delegate make verify` into README.md (in root dir).  It would be nice to show people what we're doing for verification in the first place.  (+ updating our github pages later?)

rmm/src/granule/array/mod.rs
```diff
@@ -9,18 +9,27 @@ pub const GRANULE_SHIFT: usize = 12;
...
 pub(super) const FVP_DRAM1_REGION: core::ops::Range<usize> = core::ops::Range {
     start: 0x8_8000_0000,
     end: 0x8_8000_0000 + 0x8000_0000,
 };
 
+#[cfg(not(kani))]
 pub(super) const FVP_DRAM1_IDX: usize =
     (FVP_DRAM0_REGION.end - FVP_DRAM0_REGION.start) / GRANULE_SIZE;
 
+#[cfg(kani)]
```
jinbpark, line 27: I can see why this cfg(kani) is needed for GRANULE_MEM_SIZE and GRANULE_REGION. (i.e., why it's not using FVP_DRAM_REGION.  You mentioned it at one of our weekly meetings)  

How about adding some comments on why? It would help not just us but visitors to grasp our verifications better.

model-checking/src/rmi_granule_delegate.rs
```diff
@@ -0,0 +1,119 @@
...
+use islet_rmm::granule::entry::GranuleGpt;
+use islet_rmm::granule::validate_addr;
+use islet_rmm::granule::GranuleState;
+use islet_rmm::monitor::Monitor;
+use islet_rmm::rmi;
+use islet_rmm::rmi::error::Error;
+
+#[kani::proof]
+#[kani::unwind(9)]
+fn verify_granule_delegate() {
```
jinbpark, line 13: One question:
Will granule transitions other than delegate/undelegate (e.g., delegate to rd, delegate to rec, ...) be verified by data structure verification, not kani, right?

rmm/src/granule/array/entry.rs
```diff
@@ -130,12 +130,27 @@ impl Granule {
...
+    #[cfg(not(kani))]
     fn zeroize(&mut self) {
         let addr = self.index_to_addr();
         unsafe {
             core::ptr::write_bytes(addr as *mut u8, 0x0, GRANULE_SIZE);
         }
     }
+    #[cfg(kani)]
+    // DIFF: assertion is added to reduce the proof burden
+    //       `write_bytes()` uses a small count value
```
nook1208, line 142: The size of zeroizing affects a lot for model checking ? I'm just curious :)

rmm/src/granule/array/mod.rs
```diff
@@ -34,7 +43,23 @@ pub fn validate_addr(addr: usize) -> bool {
...
+        // if the address is out of range.
+        warn!("address need to be aligned 0x{:X}", addr);
+        return false;
+    }
+    let g_start = GRANULE_REGION.as_ptr() as usize;
+    let g_end = g_start + GRANULE_MEM_SIZE;
+    if !(addr >= g_start && addr < g_end) {
+        return false;
+    }
+    true
```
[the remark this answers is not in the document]
zpzigi754: Thank you for the suggestion! I've changed the code as suggested :)

rmm/src/granule/array/mod.rs
zpzigi754: Good suggestion! I've added the comment in the source code in addition to the previous git commit message.

model-checking/src/rmi_granule_delegate.rs
zpzigi754: Good question! I think that it could be handled in kani as well. For example, `delegation to Rd` and `delegation to Rec` would be handled in rmi_create_realm and rmi_create_rec harness respectively.

jinbpark (approved):

rmm/src/granule/array/entry.rs
nook1208: I see, thank you for the information !

nook1208 (approved):

zpzigi754:
> How about putting `$ RMI_TARGET=rmi_granule_(un)delegate make verify` into README.md (in root dir). It would be nice to show people what we're doing for verification in the first place, especially given that only one month left to CC summit. (+ updating our github pages later?)

Thank you for the suggestion! I've slightly updated README.md and added a page about verification.

PR #328, opened 2024-05-29
ci: Enhance tests with inotifywait

bitboom:
This PR improves the ACS test and realm booting test.

- Using `inotifywait`, it terminates when a file is not changed for a specific amount of time.
- If the exit condition matches the last change, it exits immediately.
- The time required for `realm-booting job of CI` has been reduced by half. (about 15m -> 7m)


jinbpark (approved):

nook1208 (approved):

zpzigi754:
Good to reduce the workflow time for realm-linux booting with inotifywait!

PR #331, opened 2024-05-31
Fix to use aarch64-cpu

bitboom:
This PR updates to use `aarch64-cpu` as disscused in https://github.com/islet-project/islet/discussions/326.

- Appoximately 100 `unsafe` keyword related to system registers has been removed.
- [Added registers](https://github.com/islet-project/aarch64-cpu/tree/armv9) not supported by `aarch64-cpu` to the forked repo.
- Forked aarch64-cpu would be contributed to the upstream

*p.s. This PR does not completely replace `armv9a`. It has been applied to parts where `unsafe` could be removed.*

## Before this PR
```sh
Unsafe Keywords:
 - Total           | 184
 - Functions       | 29
 - Blocks          | 152
 - Impls           | 3
```

## After this PR
```sh
Unsafe Keywords:
 - Total           | 86
 - Functions       | 29
 - Blocks          | 54
 - Impls           | 3
```


hihi-wang (approved):

rmm/src/rmi/rec/exit.rs
```diff
@@ -114,7 +116,8 @@ fn handle_data_abort(
     run.set_exit_reason(rmi::EXIT_SYNC);
     run.set_hpfar(hpfar_el2);
 
-    let fault_ipa = ((HPFAR_EL2::FIPA & hpfar_el2) << 8) as usize;
+    let fault_ipa = hpfar_el2 & (HPFAR_EL2::FIPA.mask << HPFAR_EL2::FIPA.shift);
+    let fault_ipa = (fault_ipa << 8) as usize;
```
zpzigi754, line 120: In my opinion, the previous syntax looks more succint as it does not have `mask` and `shift`. This comment can be applied to other similar places, too.

zpzigi754 (approved):
Good to switch to aarch64-cpu!

rmm/src/rmi/rec/exit.rs
bitboom: I totally agree with you.

But mask's definition is different between `armv9a` and `aarch64-cpu(tock-register)`.
In the case of HPAR_EL2::FIPA (OFFSET 4, NUMBIT 40), armv9a::HPAR_EL2::FIPA of armv9a itself means mask with shift applied (ff..ff0).

And aarch64-cpu::HPAR_EL2::FIPA.mask is It means ff..ff without shift applied.

rmm/src/lib.rs
```diff
@@ -78,25 +78,26 @@ pub unsafe fn start(cpu_id: usize) {
...
-            | HCR_EL2::TERR
-            | HCR_EL2::TLOR
-            | HCR_EL2::RW
-            | HCR_EL2::TSW
-            | HCR_EL2::TACR
-            | HCR_EL2::TIDCP
-            | HCR_EL2::TSC
-            | HCR_EL2::TID3
-            | (HCR_EL2::BSU & 0b01)
-    //        | HCR_EL2::TWI
```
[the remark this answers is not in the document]
bitboom: I restored it!

rmm/src/rmi/rec/exit.rs
zpzigi754: Right. It seems that using the previous syntax is not allowed without the aarhc64-cpu's modification.

PR #332, opened 2024-06-05
safe-abstraction: Apply to granule-locked structs

bitboom:
This PR applies safe abstraction to structures managed with granule locks.

1. The affected structures are RD, RTT and REC.
2. `content_mut` has been modified to manage `&mut` in accordance with Rust's syntax.
3. Four `unsafe` keyword have been removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1226,6 +1226,7 @@ name = "vmsa"
 version = "0.0.1"
 dependencies = [
  "armv9a",
+ "safe_abstraction",
  "spinning_top",
 ]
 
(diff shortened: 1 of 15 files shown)


jinbpark (approved):

bokdeuk-jeong:
@bitboom , @jinbpark 
It's a long overdue question, how can we relate it to the intern's task?

jinbpark:
What the intern is working on is a dead-lock detection. It would look like a spinlock library (which can replace `spinning_top::Spinlock`) having the ability to detect a dead-lock situation.

PR #334, opened 2024-06-10
Avoid double copy on RMI_DATA_CREATE 

bitboom:
## Background
To mitigate concurrency issues with memory regions shared
between the host and RMM, the `host::copy_from` operation was
used to copy data from the host.

## Problem
However, for the RMI_DATA_CREATE case, this mitigation
let to double copy costs.

## This PR
To address this, a new function `host::direct_copy_from` has been
added to perform the copy in a single step.


bitboom:
> What this PR is doing has to do with our off-line discussion today? a new yet efficient method to enforce one-time copy using rust syntax.

In our offline discussion, we also talked about taking ownership of individual fields in the structure when reading them individually. Additional considerations are required to support these matters.

rmm/src/rmi/rtt.rs
```diff
@@ -206,17 +206,14 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // `page_table` is currently not reachable in model checking harnesses
         rmm.page_table.map(target_pa, true);
 
-        // read src page
-        let src_page = host::copy_from::<DataPage>(src_pa).ok_or(Error::RmiErrorInput)?;
+        // copy src to target
+        host::direct_copy_from::<DataPage>(src_pa, &mut target_page).ok_or(Error::RmiErrorInput)?;
```
nook1208, line 210: Is it different to use host::copy_to ? (maybe the 'src' is not included in rmm page table yet ?)

bitboom: Right! Also, the type of destination is different. The function name has been changed to a clearer one. (copy_to_ptr & copy_to_obj)

nook1208 (approved):

rmm/src/host.rs
```diff
@@ -33,7 +33,30 @@ pub fn copy_from<T: SafetyChecked + SafetyAssured + Copy>(addr: usize) -> Option
     }
 }
 
-pub fn copy_to<T: SafetyChecked + SafetyAssured + Copy>(src: &T, dst: usize) -> Option<()> {
+pub fn copy_to_obj<T: SafetyChecked + SafetyAssured + Copy>(src: usize, dst: &mut T) -> Option<()> {
```
zpzigi754, line 36: Good to remove the duplicate copy! One minor comment is that passing `src` might be against the convention where only dst is only passed like
```
pub unsafe fn copy_to(self, dest: *mut T, count: usize)
```
Reference: https://doc.rust-lang.org/std/primitive.pointer.html#method.copy_to

zpzigi754: Oh, it seems that @nook1208 has already left the similar comment.

bitboom: Thanks for the suggestion. Is it okay with `copy_to_obj`?

zpzigi754: I think that removing `src` might be better with `self`, but leaving it as it is would be also okay.

bitboom: Oh I understand your words now. To do so, I need to implement a method in `AssumeSafe`. I will consider this later. Thanks!

PR #344, opened 2024-07-04
Safe instance creation at given memory address

bitboom:
This PR presents a quick draft for safely creating an instance at a given memory address.
As discussed in https://github.com/islet-project/islet/pull/342, utilizing the idea proposed by @L0czek, this approach involves parsing the given address as uninitialized memory, then copying that instance to that address.

For REC and RD, the copy cost was not considered significant since the operation is called only once.

### Concept
```rust
// Create Instance
let addr = GIVEN_ADDRESS;
let value = T;
unsafe {
  let src: core::mem::MaybeUninit<T> = core::mem::MaybeUninit::new(value);
  let src = &src as *const core::mem::MaybeUninit<T>;
  let dst = addr as *mut core::mem::MaybeUninit<T>;
  core::ptr::copy_nonoverlapping(src, dst, 1);
}

// Access Instance
let ptr = addr as *mut core::mem::MaybeUninit<T>;
let value = (*ptr).assume_init_ref();
```

### API changed 
```rust
- let mut rec = rec_granule.content_mut::<Rec<'_>>()?;
+ let mut rec = rec_granule.new_uninit_with::<Rec<'_>>(Rec::new())?;
```

### DIFF
--- a/lib/safe-abstraction/src/lib.rs
+++ b/lib/safe-abstraction/src/lib.rs
@@ -218,8 +218,30 @@ pub mod raw_ptr {
 
         Ok(SafetyAssumed {
             addr,
+            assume_init: false,
             _phantom: core::marker::PhantomData,
         })
+    }
+
+    pub fn assume_safe_uninit_with<T: SafetyChecked + SafetyAssured>(
+        addr: usize,
+        value: T,
+    ) -> Result<SafetyAssumed<T>, Error> {
+        let ptr = addr as *const T;
+        // Safety: This cast from a raw pointer to a reference is considered safe
+        //         because it is used solely for the purpose of verifying alignment and range,
+        //         without actually dereferencing the pointer.
+        let ref_ = unsafe { &*(ptr) };
+
+        if !ref_.is_not_null() || !ref_.is_aligned() {
+            return Err(Error::SafetyCheckFailed);
+        }
+
+        if !ref_.verify_ownership() {
+            return Err(Error::AssuranceCheckFailed);
+        }
+
+        Ok(SafetyAssumed::new_maybe_uninit_with(addr, value))
     }
 
     /// Represents a target instance that has passed all necessary safety checks.
@@ -239,7 +261,28 @@ pub mod raw_ptr {
     ///                and ensure that the Rust compiler accounts for `T` in its type checking.
     pub struct SafetyAssumed<T: SafetyChecked + SafetyAssured> {
         addr: usize,
+        assume_init: bool,
         _phantom: core::marker::PhantomData<T>,
+    }
+
+    impl<T> SafetyAssumed<T>
+    where
+        T: SafetyChecked + SafetyAssured,
+    {
+        pub fn new_maybe_uninit_with(addr: usize, value: T) -> Self {
+            unsafe {
+                let src: core::mem::MaybeUninit<T> = core::mem::MaybeUninit::new(value);
+                let src = &src as *const core::mem::MaybeUninit<T>;
+                let dst = addr as *mut core::mem::MaybeUninit<T>;
+                core::ptr::copy_nonoverlapping(src, dst, 1);
+            }
+
+            Self {
+                addr,
+                assume_init: true,
+                _phantom: core::marker::PhantomData,
+            }
+        }
     }
 
     impl<T> AsRef<T> for SafetyAssumed<T>
@@ -253,7 +296,14 @@ pub mod raw_ptr {
         /// are in place. Mutable access is granted under the presumption of exclusive ownership
         /// and proper synchronization when accessed in multi-threaded contexts.
         fn as_ref(&self) -> &T {
-            unsafe { T::as_ref(self.addr) }
+            unsafe {
+                if self.assume_init {
+                    let ptr = self.addr as *mut core::mem::MaybeUninit<T>;
+                    (*ptr).assume_init_ref()
+                } else {
+                    T::as_ref(self.addr)
+                }
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


lib/safe-abstraction/src/lib.rs
```diff
@@ -253,7 +296,14 @@ pub mod raw_ptr {
         /// are in place. Mutable access is granted under the presumption of exclusive ownership
         /// and proper synchronization when accessed in multi-threaded contexts.
         fn as_ref(&self) -> &T {
-            unsafe { T::as_ref(self.addr) }
+            unsafe {
+                if self.assume_init {
+                    let ptr = self.addr as *mut core::mem::MaybeUninit<T>;
+                    (*ptr).assume_init_ref()
+                } else {
+                    T::as_ref(self.addr)
```
bokdeuk-jeong, line 304: Is it okay to return the reference while its assume_init is false?

rmm/src/rmi/rec/handlers.rs
```diff
@@ -70,8 +70,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         #[cfg(not(kani))]
         // `page_table` is currently not reachable in model checking harnesses
         rmm.page_table.map(rec, true);
-        let mut rec = rec_granule.content_mut::<Rec<'_>>()?;
-
+        let mut rec = rec_granule.new_uninit_with::<Rec<'_>>(Rec::new())?;
```
bokdeuk-jeong, line 73: Could you please explain what happens if we keep this line with the existing implmentation?
Even with previous implementation, all access to rec is made after rec.init() call. So, I don't see why we need this change.

lib/safe-abstraction/src/lib.rs
```diff
@@ -239,9 +261,30 @@ pub mod raw_ptr {
...
+    impl<T> SafetyAssumed<T>
+    where
+        T: SafetyChecked + SafetyAssured,
+    {
+        pub fn new_uninit_with(addr: usize, value: T) -> Self {
+            unsafe {
+                let src: core::mem::MaybeUninit<T> = core::mem::MaybeUninit::new(value);
+                let src = &src as *const core::mem::MaybeUninit<T>;
+                let dst = addr as *mut core::mem::MaybeUninit<T>;
+                core::ptr::copy_nonoverlapping(src, dst, 1);
```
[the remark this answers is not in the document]
bitboom: The current use case involves copying `src` within RMM memory to `dst` outside RMM memory. However, to defensive code, it might be prudent to either either check if the memory regions overlap or use `copy` instead of `copy_nonoverlapping`.

lib/safe-abstraction/src/lib.rs
bitboom: If `assume_init` returns false, it offers the same security level as the existing API (using `content_mut`). I plan to explore ways to improve this by integrating [provenance](https://doc.rust-lang.org/std/primitive.pointer.html#method.expose_provenance) concept.

rmm/src/rmi/rec/handlers.rs
bitboom: The most important question! Previously, before calling `rec.init()`, the compiler could not determine whether the `rec` memory was initialized, leading to writing directly to it. Currently, the `rec` memory is first initialized with a `MaybeUninit` instance before invoking `init()`."

bokdeuk-jeong: I get this part, "the compiler could not determine whether the rec memory was initialized,", but not "leading to writing directly to it".  Could you elaborate on that?

bitboom: > "leading to writing directly to it". Could you elaborate on that?

I mean this part is `rec.init()`. 

## Previous
```rust
// Compiler couldn't know whether the memory was initialized
let mut rec = rec_granule.content_mut::<Rec<'_>>()?; 
// Write REC directly
rec.init();
```

## Current
```rust
// Let compiler know the memory was initialized by MaybeUninit<Rec>
let mut rec = rec_granule.new_uninit_with::<Rec<'_>>(Rec::new())?;
// Write REC directly
rec.init();
```

rmm/src/rec.rs
```diff
@@ -83,10 +106,8 @@ impl Rec<'_> {
         self.set_ripas(0, 0, 0, 0);
         self.set_runnable(flags);
         self.set_state(State::Ready);
-        self.context = Context::new();
         self.context.sys_regs.vttbr = vttbr;
         self.context.sys_regs.vmpidr = vmpidr;
-        self.attest_state = RmmRecAttestState::NoAttestInProgress;
```
nook1208, line 89: If we remove these two lines because they will be initialized with a constructor function (Rec::new()), I think that the other fields(ripas, State) which is not related with inputs of this init() similarly shouldn't be touched too.

rmm/src/rec.rs
```diff
@@ -57,6 +57,29 @@ pub struct Rec<'a> {
 }
 
 impl Rec<'_> {
+    pub fn new() -> Self {
```
nook1208, line 60: Good to add this constructor!

nook1208 (approved):

rmm/src/rec.rs
bitboom: Sounds reasonable :) I fixed it. Thanks!

lib/safe-abstraction/src/lib.rs
```diff
@@ -239,9 +261,30 @@ pub mod raw_ptr {
...
         addr: usize,
+        assume_init: bool,
         _phantom: core::marker::PhantomData<T>,
     }
 
+    impl<T> SafetyAssumed<T>
+    where
+        T: SafetyChecked + SafetyAssured,
+    {
+        pub fn new_uninit_with(addr: usize, value: T) -> Self {
```
hihi-wang, line 272: minor) how about naming this as `new_maybe_uninit_with`? Though it's too long, it makes more clear intention

hihi-wang (approved):

lib/safe-abstraction/src/lib.rs
bitboom: I reflected it :)

PR #346, opened 2024-07-11
Spinlock Library with potential Deadlock Detection Feature

Goutam-115:
Spinlock library for ISLET platform, that provides potential deadlock detection feature for the developers. This library uses conditional compilation feature flags for different uses. Normally it behaves as a locking library for mutual exclusion, and also provides API for potential deadlock detection. This library uses a strong algorithm for deadlock detection to reduce false positive cases during analysis. Both real time deadlock and potential deadlock situation can be detected by this library. ReadMe file is added in the library for the developers explaining the usage of the library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -460,6 +460,7 @@ dependencies = [
  "linked_list_allocator",
  "log",
  "uart",
+ "spinlock",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


jinbpark:
@Goutam-115, could you turn this PR to a draft PR?

zpzigi754:
FYI, we use [spinning_top](https://github.com/rust-osdev/spinning_top)'s spinlock. Thus, I guess that modifying the existing one might help to remove the duplicate functionality regarding locking mechanism.

PR #348, opened 2024-07-15
vmsa: introduce new_in() with an allocator as a parameter

bokdeuk-jeong:
The ulimate goal is to eliminate redundant functions for the realm s2 table management which are scattered across the board in rtt.rs, stage2_translation.rs, stage2_tte.rs, page.rs and lib/vmsa.
Prior to that, this PR does followings.
1. Create a page table with an allocator that the function caller provides.
2. Return as a reference and reduce unsafe usages.
3. Simplify mutiple look-alike functions into one.

### DIFF
--- a/lib/vmsa/src/page_table.rs
+++ b/lib/vmsa/src/page_table.rs
@@ -36,38 +36,17 @@ pub trait Entry {
     fn address(&self, level: usize) -> Option<PhysAddr>;
 
     fn set(&mut self, addr: PhysAddr, flags: u64, is_raw: bool) -> Result<(), Error>;
-    fn set_with_page_table_flags(&mut self, addr: PhysAddr) -> Result<(), Error>;
+    fn point_to_subtable(&mut self, index: usize, addr: PhysAddr) -> Result<(), Error>;
 
     // returns EntryGuard which allows accessing what's inside Entry while holding a proper lock.
     fn lock(&self) -> Result<Option<EntryGuard<'_, Self::Inner>>, Error> {
         Err(Error::MmUnimplemented)
     }
 
-    // do memory allocation using closure,
-    // this is useful if a page table wants entry-level locking,
-    // as validity_check and set_* function must be done under entry-level locking.
-    // Safety: this  implementation doesn't guarantee entry-level locking.
-    //         there is a race window between `is_valid()` and `set_with_page_table_flags()`.
-    //         this is only safe if with a big lock. If you want entry-level locking, do override this function properly.
-    fn set_with_page_table_flags_via_alloc<T: FnMut() -> usize>(
-        &mut self,
-        _index: usize,
-        mut alloc: T,
-    ) -> Result<(), Error> {
-        if !self.is_valid() {
-            let table = alloc();
-            if table == 0 {
-                Err(Error::MmAllocFail)
-            } else {
-                self.set_with_page_table_flags(PhysAddr::from(table))
-            }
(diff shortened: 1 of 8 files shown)


rmm/src/realm/mm/stage2_translation.rs
```diff
@@ -55,14 +55,18 @@ macro_rules! init_table {
             RootTable<$level, $pages>,
             entry::Entry,
             { <RootTable<$level, $pages> as Level>::NUM_ENTRIES },
-        >::new_with_base($base)
+        >::new_init_in(&RttAllocator { base: $base }, |entries| {
```
jinbpark, line 58: Minor: is there any reason it uses `new_init_in()` instead of `new_in()`?

jinbpark (approved):

lib/vmsa/src/page_table.rs
```diff
@@ -158,47 +139,55 @@ pub trait PageTableMethods<A: Address, L, E: Entry, const N: usize> {
...
         if table as usize == 0 {
             return Err(Error::MmAllocFail);
         }
-
-        let table = table as *mut PageTable<A, L, E, N>;
         unsafe {
-            let arr: [E; N] = core::array::from_fn(|_| E::new());
-            (*table).entries = arr;
+            for entry in (*table).entries.iter_mut() {
+                // In general .write is required to avoid attempting to destruct
```
hihi-wang, line 153: How about adding an example link that using `write` pattern on Rust std?

rmm/src/realm/mm/stage2_translation.rs
bokdeuk-jeong: This will be changed into using new_in() in the end. I just wanted to test and show how it can be used by applying new_init_in(). new_init_in() is intened to replace the exising usage of RttPage struct.

lib/vmsa/src/page_table.rs
bokdeuk-jeong: I tried, but couldn't. If you can appy  write(), could you commit your code to this branch?

hihi-wang: Oh, what I meant was to adding an example link on the comment side, not adjusting that pattern on our code

bokdeuk-jeong: I took a reference from allocator-api2-0.2.16/src/stable/boxed.rs where my cargo stores crates locally, but don't know how to find the corresponding URL.

```
 905     /// Manually create a `Box` from scratch by using the global allocator:
 906     /// ```
 907     /// use std::alloc::{alloc, Layout};
 908     ///
 909     /// unsafe {
 910     ///     let ptr = alloc(Layout::new::<i32>()) as *mut i32;
 911     ///     // In general .write is required to avoid attempting to destruct
 912     ///     // the (uninitialized) previous contents of `ptr`, though for this
 913     ///     // simple example `*ptr = 5` would have worked as well.
 914     ///     ptr.write(5);
 915     ///     let x = Box::from_raw(ptr);
 916     /// }
 917     /// ```

```

hihi-wang: You may use this(github link that noted on crates.io) : https://github.com/zakarumych/allocator-api2/blob/main/src/stable/boxed.rs#L905

bitboom (approved):

hihi-wang (approved):

nook1208 (approved):

PR #349, opened 2024-07-15
Pagetable with no function overriding for adopting stable rust

zpzigi754:
The purpose of this change is removing `function overriding` ([specialization feature](https://rust-lang.github.io/rfcs/1210-impl-specialization.html)) in the page table implementation which prevents us from using a `stable rust`. 

This PR is just for the reference; it is not aimed to be merged, because it is not the final form of page table I have in mind. 

[before]
```
pub struct PageTable<A, L, E, const N: usize> {
     entries: [E; N],
     level: PhantomData<L>,
     address: PhantomData<A>,
}
```

[after]
```
pub struct PageTable<A, E, const N: usize> {
     entries: [E; N],
     address: PhantomData<A>,
 }
```


bokdeuk-jeong:
I fully understand your motivation. However, level managment seems to be more of PageTable's role rather than Entry's.
Could you come up with another approach?

jinbpark:
Should we look at how other Rust-based OS projects deal with page table management, in the context of 1) using a minimal number of unsafe keyword, 2) using stable rust as much as possible?  That might help to some extent. (Also, I'm curious about how other people do this stuff)

bokdeuk-jeong:
Our code is inspired by https://os.phil-opp.com/page-tables/.
With the same motivation behind your question, the projects that I have investigated thus far had simpler configurations like with a fixed root level and a predetermined number of root pages.

PR #353, opened 2024-07-18
Changes to RTT process code (rtt.rs)

bokdeuk-jeong:
1. Move rtt.rs to realm/mm
2. Unify the page table management by removing RttPage and using PageTable to initialize delegated pages for RTT purpose.
3. Make dependencies uni-directional.
       e.g.) Move RTT_ constants in rmi/rtt.rs to ream/mm/rtt.rs and let rmi/rtt.rs dependent on realm/mm/rtt.rs,
       not the other way round.
4. Eliminate realm_linux feature in rmm build.
5. Use S2TTE as Inner of Entry and remove RawPTE.

TODOs after this PR is merged: add tbl operations along with page table upates.


scripts/fvp-cca
```diff
@@ -234,8 +234,6 @@ def get_rmm_features(args):
 
         if args.stat == True:
             features += ["--features", "stat"]
-        if args.realm == "linux" or args.realm == "aosp" or args.use_prebuilt == True:
-            features += ["--features", "realm_linux"]
```
jinbpark, line 238: Is it necessary to remove these lines  in this PR?

bokdeuk-jeong: Yes. We don't need this feature configuration anymore. The feature was used to statically differentiate how to set up a page table for each case such as linux or other (e.g., acs) realms. 
We are now able to set up the exact page table as requested during the realm creation RMI.

jinbpark (approved):

scripts/fvp-cca
zpzigi754: I forgot to remove it: the feature became obsolete after eac5 update. Thank you for the removal!

rmm/src/realm/mm/entry.rs
```diff
@@ -10,21 +10,27 @@ use vmsa::RawGPA;
 use armv9a::bits_in_reg;
 
 #[derive(Clone, Copy)]
-pub struct Entry(RawPTE);
```
zpzigi754, line 13: The previous RawPTE was the entry for the stage 1 translation while `S2TTE` is the entry for the stage 2 translation. I think that distiniguishing the types for the different entries would be better, as they have non-overlapped functionalities. Also, using `S2TTE` for the stage 1 translation would be confusing.

bokdeuk-jeong: PageTable in realm/mm is for state 2 translation only. RawPTE was meant for state2 only. S2TTE and RawPTE is rather a  duplication.
lib/vmsa is agnostic in terms of stage 1 and 2.

hihi-wang (approved):

nook1208 (approved):

rmm/src/realm/mm/entry.rs
zpzigi754: Uoops, my fault. I thought that it was the `PTDesc` in `mm/page_table/entry.rs`. Sorry for the confusion!

rmm/src/realm/mm/rtt.rs
```diff
@@ -1,60 +1,59 @@
...
-        2 => GRANULE_SIZE << S2TTE_STRIDE,
-        1 => GRANULE_SIZE << (S2TTE_STRIDE * 2),
-        0 => GRANULE_SIZE << (S2TTE_STRIDE * 3),
-        _ => unreachable!(),
-    }
+fn create_pgtbl_at(rtt_addr: usize, flags: u64, mut pa: usize, map_size: usize) {
+    let alloc = RttAllocator { base: rtt_addr };
+    let mut new_s2tte = pa as u64 | flags;
+
+    let _ = PageTable::<
```
bitboom, line 32: (minor) How about to handle the error like `unwrap()` or `expect("Failed to create the page table.")`?

bitboom (approved):

rmm/src/realm/mm/rtt.rs
```diff
@@ -1,60 +1,59 @@
...
+        GuestPhysAddr,
+        table_level::L3Table, //Table Level is not meaninful here
+        entry::Entry,
+        { table_level::L3Table::NUM_ENTRIES },
+    >::new_init_in(&alloc, |entries| {
+        for e in entries.iter_mut() {
+            let _ = (*e).set(PhysAddr::from(pa), new_s2tte, true);
```
zpzigi754, line 39: Not about this PR, but I think that removing the `is_raw` field (the value `true` here) might be better for the perception by integrating all `Entry``s `set()` method to be called as `is_raw` (not passing `MAP_EXTRA_FLAG` in default).

rmm/src/realm/mm/rtt.rs
```diff
@@ -1,60 +1,59 @@
...
-        0 => GRANULE_SIZE << (S2TTE_STRIDE * 3),
-        _ => unreachable!(),
-    }
+fn create_pgtbl_at(rtt_addr: usize, flags: u64, mut pa: usize, map_size: usize) {
+    let alloc = RttAllocator { base: rtt_addr };
+    let mut new_s2tte = pa as u64 | flags;
+
+    let _ = PageTable::<
+        GuestPhysAddr,
+        table_level::L3Table, //Table Level is not meaninful here
```
zpzigi754, line 34: As mentioned, the table level might be misleading, as it could be `L1Table` or `L2Table` unlike the passed value set as `L3Table`. I think that using `create_pgtbl_at()` itself is good for the readability, but I am not sure using `new_init_in()` in this function is desirable, as it makes the interface heavier and not correct in terms of the table level.

rmm/src/realm/mm/stage2_tte.rs
```diff
@@ -24,84 +21,48 @@ pub mod invalid_ripas {
...
-    ADDR_L1_PAGE[47 - 30], // XXX: check this again
-    ADDR_L2_PAGE[47 - 21], // XXX: check this again
-    ADDR_L3_PAGE[47 - 12], // XXX: check this again
-    ADDR_FULL[55 - 12],
+    CONT[52 - 52],
+    // https://armv8-ref.codingbelief.com/en/chapter_d4/d43_1_vmsav8-64_translation_table_descriptor_formats.html
+    ADDR_BLK_L0[47 - 39],      // block descriptor; level 0 w/o concatenation
+    ADDR_BLK_L1[47 - 30],      // block descriptor; level 1
+    ADDR_BLK_L2[47 - 21],      // block descriptor; level 2
+    ADDR_BLK_L3[47 - 12],      // page descriptor; level 3
```
zpzigi754, line 53: Is there any reaon to use `ADDR_BLK_L3` while L3 is a page descriptor? (for consistency?)

rmm/src/realm/mm/rtt.rs
bokdeuk-jeong: I do agree with you and had been hesitant with using L3Table, but ended up with deciding the code less complicated and leaving a comment instead.

rmm/src/realm/mm/stage2_tte.rs
bokdeuk-jeong: for consistency. :) 
Do you prefer to using PAGE instead?

zpzigi754: Either would be okay!

zpzigi754 (approved):

Issue #350, opened 2024-07-15
realm measurement using veraison


darshanp4:
hello ,

I am trying to attestation using veraison with islet from given steps at : [veraison example run](https://github.com/islet-project/islet/blob/ed25d9d6c89b5719268be97ce770ec8e1bfbaadd/examples/veraison/RUN.md)

while doing attest using [./bin/rsictl attest -o token.bin](https://github.com/islet-project/islet/blob/ed25d9d6c89b5719268be97ce770ec8e1bfbaadd/examples/veraison/RUN.md#:~:text=shared%20%23%20./bin/rsictl%20attest%20%2Do%20token.bin)

facing below error :

```
rsi: device rsi open
rsi: ioctl: unknown ioctl cmd: 3494148290
rsi: device rsi released
Error: EINVAL
```

dmesg :
[ 0.000000] RME: Using RSI version 1.0
[ 0.345241] squashfs: version 4.0 (2009/01/31) Phillip Lougher
[ 1.911383] Block layer SCSI generic (bsg) driver version 0.4 loaded (major 249)
[ 12.149515] VFIO - User Level meta-driver version: 0.3
[ 12.516895] registered taskstats version 1
[ 747.676207] rsi: loading out-of-tree module taints kernel.
[ 747.691658] rsi: Initializing
[ 747.699933] rsi: Chardev registered with major 245
[ 747.712592] rsi: Device created on /dev/rsi
[ 747.723240] rsi: Is realm: true
[ 747.731454] rsi: RSI version, ret: RSI_SUCCESS, lower: 1.0, higher: 1.0
[ 747.747861] rsi: RSI config, ret: RSI_SUCCESS, ipa_width_in_bits: 33
[ 884.258661] rsi: device rsi open
[ 884.267063] rsi: ioctl: unknown ioctl cmd
[ 884.281800] rsi: device rsi released

and while doing realm measurement , I am not able find the below files mentioned at [realm measurement](https://github.com/islet-project/islet/blob/ed25d9d6c89b5719268be97ce770ec8e1bfbaadd/examples/veraison/RUN.md#:~:text=out/shared%2C-,Image.realm%20initramfs%2Drealm.cpio.gz%20realm.sh,-)

> Image.realm & realm.s

can someone help with it?

p-sawicki2:
[...]

PR #354, opened 2024-07-18
Revert "Use DUCT-TAPE to fix irq test"

L0czek:
This reverts commit 2e0f9ffa672f526c91b5a6128cb70df082f33fd4.

It appears that some (maybe Islet) is not handling some IRQ properly. As a result this patch broke reading the attestation token even though the cca-rmm-acs tests were passing. It it most likely due to the fact that the test suite was a simple bare metal binary that doesn't utilize much of the emulated hardware. Reverting this patch doesn't remove any functionality. In the future, the IRQ handling should be investigated.

### DIFF
--- a/lib/armv9a/src/lib.rs
+++ b/lib/armv9a/src/lib.rs
@@ -11,16 +11,3 @@ pub use regs::*;
 pub const fn bits_in_reg(mask: u64, val: u64) -> u64 {
     (val << (mask.trailing_zeros())) & mask
 }
-
-pub fn is_irq_pending() -> bool {
-    let val: u64;
-
-    unsafe {
-        core::arch::asm!(
-            "mrs {}, ISR_EL1",
-            out(reg) val
-        )
-    }
-
-    val != 0
-}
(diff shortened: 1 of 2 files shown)


Havner:
I've verified that the #350 is fixed with that commit. From my understanding reverting https://github.com/islet-project/islet/commit/2e0f9ffa672f526c91b5a6128cb70df082f33fd4 means that one of ACS tests will start to fail again, but it's of a less importance that being able to fetch the token and perform attestation.

@L0czek please make a note here what github issues is tracking this stuff.

L0czek:
> @L0czek please make a note here what github issues is tracking this stuff.

This issue is tracked in https://github.com/islet-project/islet/issues/352 at 
> ### The attestation irq exit test
> The attestation_rec_exit_irq attempts to test whether reading the attestation token from rmm can be interrupted by an IRQ. Since, the IRQ is a lower level exception, it cannot interrupt the control flow of Islet. As a result, it needs to be actively checked during RSI handling. [...]

PR #358, opened 2024-08-09
    realm/mm: RMI_RTT_FOLD support

bokdeuk-jeong:
1. Implement RTT_FOLD RMI.    
      - Added an RMI handler for RTT_FOLD.
      - Provide an iterator for the entry enumeration in a table at a specific level in PageTable and use the iterator when rtt.rs needs to access the whole entries at a time like rtt-fold.
    3. Fix errors in rtt found during the fold test.
      - Fix errors in subtable access in PageTable
      - Fix implementations that doesn't comply with RMM spec.
     4. Tlb invalidation for state2 translation table
      - added two versions of range invalidation, just for case that being unable to use tlb-rmi support.
     5. Misc
        - Move realm address space related checks which depends on ipa_bits to Rd.
        - Remove is_raw() param of set() in Entry
       
```
REGRESSION REPORT:
==================
   TOTAL TESTS     : 100
   TOTAL PASSED    : 92
   TOTAL FAILED    : 2
   TOTAL SKIPPED   : 5
   TOTAL SIM ERROR : 1

******* END OF ACS *******

```

[ Todo's but not in this scope ]
1. Implement more to pass acs-tests all.
```
Suite=gic : gic_timer_rel1_trig
Suite=pmu_debug : pmu_overflow
Suite=memory_management : mm_hipas_assigned_ripas_empty_da_ia
Suite=memory_management : mm_hipas_unassigned_ripas_empty_da_ia
Suite=gic : gic_timer_rel1_trig
Suite=pmu_debug : pmu_overflow
Suite=memory_management : mm_hipas_assigned_ripas_empty_da_ia
Suite=memory_management : mm_hipas_unassigned_ripas_empty_da_ia
```

3. Thoroughly review each rmi whether it complies with RMI spec.
4. Utilize the table entry enumerator more for other functions in rtt.

### DIFF
--- a/rmm/.cargo/config.toml
+++ b/rmm/.cargo/config.toml
@@ -1,3 +1,6 @@
 [build]
 target = "aarch64-unknown-none-softfloat"
-rustflags = ["-C", "target-feature=+ecv"]
+rustflags = [
+  "-C", "target-feature=+ecv",
+  "-C", "target-feature=+tlb-rmi"
+]
--- a/rmm/src/realm/mm/rtt.rs
+++ b/rmm/src/realm/mm/rtt.rs
@@ -241,9 +241,6 @@ pub fn get_ripas(rd: &Rd, ipa: usize, level: usize) -> Result<u64, Error> {
     Ok(s2tte.get_ripas())
 }
 
-// FIXME: cca-rmm-acs and linux realm is using old version of entry state.
-// i.e. no differentiation between (un)assigned and (un)assigned_ns
-// while RMM spec defines each individually.
 pub fn read_entry(rd: &Rd, ipa: usize, level: usize) -> Result<[usize; 4], Error> {
     let (s2tte, last_level) = S2TTE::get_s2tte(rd, ipa, level, Error::RmiErrorRtt(0))?;
 
--- a/rmm/src/realm/mm/rtt.rs
+++ b/rmm/src/realm/mm/rtt.rs
@@ -55,7 +55,6 @@ fn create_pgtbl_at(
 
 pub fn create(rd: &Rd, rtt_addr: usize, ipa: usize, level: usize) -> Result<(), Error> {
     let mut invalidate = Tlbi::NONE;
-    let mut rtt_granule = get_granule_if!(rtt_addr, GranuleState::Delegated)?;
 
     let (parent_s2tte, last_level) = S2TTE::get_s2tte(rd, ipa, level - 1, Error::RmiErrorInput)?;
 
@@ -116,8 +115,6 @@ pub fn create(rd: &Rd, rtt_addr: usize, ipa: usize, level: usize) -> Result<(), 
         panic!("Unexpected s2tte value:{:X}", parent_s2tte.get());
     }
 
-    set_granule(&mut rtt_granule, GranuleState::RTT)?;
-
     let parent_s2tte = rtt_addr as u64 | bits_in_reg(S2TTE::DESC_TYPE, desc_type::L012_TABLE);
     rd.s2_table().lock().ipa_to_pte_set(
         GuestPhysAddr::from(ipa),
@@ -141,8 +138,8 @@ pub fn destroy<F: FnMut(usize)>(
     let invalidate;
     let (parent_s2tte, last_level) = S2TTE::get_s2tte(rd, ipa, level - 1, Error::RmiErrorRtt(0))?;
 
-    if (last_level != level - 1) || !parent_s2tte.is_table(level - 1) {
-        let top_ipa = skip_non_live_entries(rd, ipa, level)?;
+    if (last_level != level - 1) || !parent_s2tte.is_table(last_level) {
+        let top_ipa = skip_non_live_entries(rd, ipa, last_level)?;
         f(top_ipa);
         return Err(Error::RmiErrorRtt(last_level));
     }
@@ -177,7 +174,7 @@ pub fn destroy<F: FnMut(usize)>(
 
     set_granule(&mut g_rtt, GranuleState::Delegated)?;
 
-    let top_ipa = skip_non_live_entries(rd, ipa, level)?;
+    let top_ipa = skip_non_live_entries(rd, ipa, last_level)?;
     Ok((rtt_addr, top_ipa))
 }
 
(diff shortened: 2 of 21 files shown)


rmm/src/realm/mm/rtt.rs
```diff
@@ -236,6 +241,9 @@ pub fn get_ripas(rd: &Rd, ipa: usize, level: usize) -> Result<u64, Error> {
     Ok(s2tte.get_ripas())
 }
 
+// FIXME: cca-rmm-acs and linux realm is using old version of entry state.
+// i.e. no differentiation between (un)assigned and (un)assigned_ns
+// while RMM spec defines each individually.
```
zpzigi754, line 246: About this comment, tf-rmm as well as islet currently interpret the spec with the following table where unassigned and unassigned_ns is distinguished with `NS` bit. 

```
// ======================================================================================
// s2tte type           level DESC_TYPE[1:0] HIPAS[4:2]    RIPAS[6:5]   NS  OA alignment
// ======================================================================================
// unassigned_empty     any   invalid[0]     unassigned[0] empty[0]     0   n/a
// --------------------------------------------------------------------------------------
// unassigned_ram       any   invalid[0]     unassigned[0] ram[1]       0   n/a
// --------------------------------------------------------------------------------------
// unassigned_destroyed any   invalid[0]     unassigned[0] destroyed[2] 0   n/a
// --------------------------------------------------------------------------------------
// assigned_empty       2,3   invalid[0]     assigned[1]   empty[0]     0   to level
// --------------------------------------------------------------------------------------
// assigned_ram         3     page[3]        n/a           n/a          0   to level
//                      2     block[1]       n/a           n/a          0   to level
// --------------------------------------------------------------------------------------
// assigned_destroyed   any   invalid[0]     assigned[1]   destroyed[2] 0   n/a
// ======================================================================================
// unassigned_ns        any   invalid[0]     unassigned[0] n/a          1   n/a
// --------------------------------------------------------------------------------------
// assigned_ns          3     page[3]        n/a           n/a          1   to level
//                      2     block[1]       n/a           n/a          1   to level
// ======================================================================================
// table              <=2     table[3]       n/a           n/a          n/a to 4K
// ======================================================================================
```

zpzigi754 (approved):
Really good to support fold and pass many additional acs tests!

rmm/src/realm/mm/stage2_tte.rs
```diff
@@ -12,7 +16,6 @@ pub const INVALID_UNPROTECTED: u64 = 0x0;
 pub mod invalid_hipas {
     pub const UNASSIGNED: u64 = 0b00;
     pub const ASSIGNED: u64 = 0b01;
-    pub const DESTROYED: u64 = 0b10;
```
zpzigi754, line 15: This field, an obsolete one after the spec change, is what I was about to remove, too. Good to fix it!

rmm/src/realm/mm/stage2_translation.rs
```diff
@@ -108,27 +128,121 @@ impl<'a> Stage2Translation<'a> {
...
-                ipa = bits_in_reg(TLBI_OP::NS, tlbi_ns::IPAS_S)
-                    | bits_in_reg(TLBI_OP::TTL, 0b0100 | _level)
+                ipa = bits_in_reg(TLBI_OP::TTL, 0b0100 | level as u64)
                     | bits_in_reg(TLBI_OP::IPA, ipa);
-                // corresponds to __kvm_tlb_flush_vmid_ipa()
                 asm!(
-                    "dsb ishst",
-                    "tlbi ipas2e1is, {}",
-                    "isb",
+                    "tlbi IPAS2E1IS, {}",
```
zpzigi754, line 144: I guess that the previous one (`tlbi ipas2e1is, {}`) and the new one (`tlbi IPAS2E1IS, {}`) is the same in terms of the functionality (the difference seems small letter vs uppercase). Is is correct?

rmm/.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
 [build]
 target = "aarch64-unknown-none-softfloat"
-rustflags = ["-C", "target-feature=+ecv"]
+rustflags = [
+  "-C", "target-feature=+ecv",
+  "-C", "target-feature=+tlb-rmi"
```
zpzigi754, line 5: Is this target-feature (tlb-rmi) used for supporting `tlbi RIPAS2LE1` instruction?

rmm/src/realm/mm/stage2_translation.rs
bokdeuk-jeong: Yes. the number 1 and the letter 'l' are confusing. I prefer using captial to prevent mistakes.

rmm/.cargo/config.toml
bokdeuk-jeong: Yes.

rmm/src/realm/mm/rtt.rs
bokdeuk-jeong: I was referring to the return value definintion, not the interpertation of entry, which is used for the return value with X2 for the RTT_READ_ENTRY rmi call.
However, I rushed myself yesterday and missed the return type in the spec docuemnt.
B4.4.24 RmiRttEntryState type
I only defines three types, ASSIGNED, UNASSIGNED, TABLE.
I will update the commit accordingly.

bokdeuk-jeong: However, I still don't get that why it returns this simplied form while defines (UN)ASSIGNED and (UN)ASSIGNED_NS sperarately.  There's no value in the output that includes ns, non-ns flags. Isn't is?

zpzigi754: Right. The spec treats `RmmRttEntryState` (abstract type)  and `RmiRttEntryState` (concrete type) differently where `RttEntryState function` handles the conversion of the former to the latter. I guess that the output does not distiniguish `ASSIGNED (UNASSIGNED)` from `ASSIGNED_NS (UNASSIGNED_NS)`, because the Host can tell the difference of them using the value of IPA (e.g., if the IPA points to an unprotected IPA, then `ASSIGNED` would be thought as `ASSSIGNED_NS`.)

bitboom (approved):

bitboom:
FYI, the CI failure of `rust-unsafe-analysis` is fixed at https://github.com/islet-project/islet/pull/360.

nook1208 (approved):

hihi-wang (approved):

PR #362, opened 2024-09-19
Implement VHUK_A and VHUK_M derivation in islet-hes

zos:
This is the first step of the sealing key derivation process, which produces Virtual Hardware Unique Keys (VHUKs) used as an IKM (Input Key Material) of the realm sealing key derivation process.


zos:
wrong branch

PR #367, opened 2024-09-27
App provisioning framework

L0czek:
# App provisioning

This PR adds app provisioning framework to islet project which allows for installing applications inside realm using OCI containers. The framework components handle everything from downloading, verifying, installing and running the applications. It provides encrypted, persistent storage that each application can use to store confidential data. The realm also checks the authenticity of each application by verifying the application signature against a chain of provided certificates that begin at the embedded CA. 

The framework consists of several main components:
* The [realm image](https://github.com/islet-project/realm-manager/tree/main/realm) which implements a special realm with application manager which is responsible for installing and running applications.
* The [warden daemon](https://github.com/islet-project/realm-manager/tree/main/warden) which is responsible for creating disks, network interfaces and finally running kvmtool
* The [application registry server](https://github.com/islet-project/image-registry/tree/main/ir-server) which is an OCI container repository. It utilizes the RA-TLS protocol to ensure that the realm is trustworthy before it allows the application to be installed.
* The [realm metadata tool](https://github.com/islet-project/realm-metadata-tool) which is used to create the realm metadata block so that the sealing keys can be based on vendors public keys.
* The [RIM measuring tool](https://github.com/islet-project/assets/tree/eac5/3rd-kvmtool-realm-metadata-rim-measurer) is a modified kvmtool which is capable of calculating realm's RIM without launching it.
* This PR which adds key derivation and realm metadata block functionality to enable sealing keys for encryption in realms.

## Running the demo

The provided demo shows how an example application can be installed in a realm in islet. The detailed instruction goes over:
* setting up islet,
* preparing the special realm image,
* creating an example application image, 
* signing the application,
* setting up the application registry,
* running islet and installing the application.

The detailed instruction to reproduce the app provisioning setup can be found [here](https://github.com/islet-project/islet/blob/app-provisioning/examples/app-provisioning/README.md).

## What works now
* Installing and running applications
* Applications are preserved across reboots
* Applications are authenticated based on ecdsa signatures
* Sealing keys can be measurements dependent, if the realm metadata block was not provided to allow for running any realms securely.
* In case the metadata block is present, the sealing keys are based on realm vendor public key to allow for easy application update (the encryption keys doesn't change between application and realms versions)

### RIM calculation related
Unfortunately, due to the design of how the RIM is calculated for realms, specifying a single RIM value for the application provisioning realm is not trivial. It is caused by measuring the **D**evice **T**ree **B**lob which changes based on the kvmtool arguments. Basically, the realm RIM is dependent on such things as:
* virtual network card(s),
* virtio block devices,
* other virtual hardware,
* the order of the hardware in the DeviceTree.

This is especially troublesome as currently the warden daemon provides one disk image for each application. As a result, the kvmtool creates a separate virtio block device for each disk, leading to a change in the **DTB**. This should be investigated in the future to either stabilize the **DTB** or provide a rim measuring software. An example of such application is the [modified kvmtool](https://github.com/islet-project/assets/tree/eac5/3rd-kvmtool-realm-metadata-rim-measurer). Therefore, it will be further investigated in the future.

### Something wrong with Islet rmm
While running the provisioning setup, we started encountering a truly random errors from various layers of the stack that could not be reproduced in tf-rmm. In detail, we encountered the following errors:
* TLS decrypt error during application installation,
* application image hash mismatch,
* some "inode already exists" error from the Linux kernel,
* disk image corruption (the file system wasn't preserved across realm reboots),
* TLS handshake error.

Since we couldn't reproduce them while running tf-rmm there might be some memory issue introduced by islet rmm. We first encountered this issue when we tried to test the application installation using a statically linked rust "hello world" application. It weights about 1MB, due to static linking and debug build. Every time we tried to install it while running islet, the installation process would fail with TLS decrypt error. We then switch to tf-rmm to check if the issue still persisted. To our surprise, it did not. In the next step, we created a lighter "hello world" application that weight only 1KB and managed to get it installed under Islet. This let us conclude that Islet introduces some memory instability that leads to arbitrary corruption when large amounts of data are processed. As of now, we weren't able to pinpoint the issue. You can try to reproduce this issue by trying under islet to:
*  install the big application (`example_app` from the instruction) 
*  install the lighter one (the `light_app` from the instruction). 

Naturally, this issue must be investigated, as this makes running applications under Islet unstable. Apart from that, https://github.com/islet-project/islet/issues/361 also suggest an issue with realm's memory management.


p-sawicki2 (commented):
LGTM

.gitmodules
```diff
@@ -4,23 +4,23 @@
...
+	branch = eac5/3rd-nw-linux-realm-metadata
 [submodule "third-party/optee-build"]
 	path = third-party/optee-build
 	url = https://github.com/islet-project/assets
 	branch = 3rd-optee-build
 [submodule "third-party/realm-linux"]
 	path = third-party/realm-linux
 	url = https://github.com/islet-project/assets
-	branch = eac5/3rd-realm-linux
+	branch = devel/p.sawicki2/eac5-3rd-realm-linux-revert-rsi-workaround
```
bokdeuk-jeong, line 15: Please do not point any devel or personal branches.
Why don't you directly update eac5/3rd-realm-linux?

.gitmodules
```diff
@@ -4,23 +4,23 @@
...
 [submodule "third-party/tf-a-tests"]
 	path = third-party/tf-a-tests
 	url = https://git.trustedfirmware.org/TF-A/tf-a-tests.git
 	branch = master
 [submodule "third-party/tf-rmm"]
 	path = third-party/tf-rmm
-	url = https://git.trustedfirmware.org/TF-RMM/tf-rmm.git
-	branch = rmm-spec-v1.0-eac5
+	url = git@github.com:p-sawicki2/tf-rmm.git
+	branch = devel/p.sawicki2/eac5-realm-metadata-sealing
```
bokdeuk-jeong, line 23: here too

.gitmodules
```diff
@@ -32,7 +32,7 @@
 [submodule "third-party/kvmtool"]
 	path = third-party/kvmtool
 	url = https://github.com/islet-project/assets
-	branch = eac5/3rd-kvmtool
+	branch = eac5/3rd-kvmtool-realm-metadata
```
bokdeuk-jeong, line 35: Why do we need another branch for eac5?
Couldn't we just update and use the existing eac5/3rd-kvmtool?

.gitmodules
```diff
@@ -61,7 +61,7 @@
 [submodule "third-party/tf-a-rss"]
 	path = third-party/tf-a-rss
 	url = https://github.com/islet-project/arm-trusted-firmware
-	branch = eac5-rss
+	branch = eac5-rss-realm-metadata
```
bokdeuk-jeong, line 64: this too.

examples/app-provisioning/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Provision the Application/Realm
+
+### Obtain the RIM
+
+To obtain the RIM we can use the lkvm-rim-measurer tool.
+Firstly, clone the lkvm-rim-measurer tool repository.
+
+    cd $ROOT
+    git clone --single-branch -b eac5/3rd-kvmtool-realm-metadata-rim-measurer git@github.com:islet-project/assets.git  rim-measurer
```
bokdeuk-jeong, line 56: We will have kvmtool under islet/third-party. Why we need to clone this again?

examples/app-provisioning/README.md
```diff
@@ -0,0 +1,272 @@
...
+## Prepare the provisioning framework
+
+    cd $ROOT
+    git clone https://github.com/islet-project/realm-manager.git
+
+### Build realm image
+
+    cd $ROOT/realm-manager/realm
+    make deps
+    make compile-image
```
bokdeuk-jeong, line 31: We already have linux-ream under islet/third-party. This is duplicate work.
Why don't you make it built from islet?
By putting your own config under islet/third-party/linux-realm/ and modifying fvp-cca and related scripts to use linux config for app-provisioning, we can avoid building another realm-linux at realm-manager.

examples/app-provisioning/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+#### Copy resulting kernel and initramfs images to islet shared dir
+
+    cp out/Image $ROOT/islet/out/shared/nasz-realm
+    cp out/initramfs.cpio.gz $ROOT/islet/out/shared/initramfs.cpio.gz
+
+### Build warden daemon
+
+    cd $ROOT/realm-manager/warden
```
bokdeuk-jeong, line 40: Could you add a single line introducing what 'warden' is?

bokdeuk-jeong (commented):
I nicely describes instructions to execute the app-provisioning example, but it looks hard to understand what it is doing. It could be even nicer if explanations about what each instructions is doning is added.

examples/app-provisioning/README.md
```diff
@@ -0,0 +1,272 @@
...
+    sudo iptables -t nat -A POSTROUTING -j MASQUERADE -s 192.168.10.0/24
+    sudo ip addr del 193.168.10.15/24 dev ARM${USER}
+    sudo ip addr add 192.168.10.15/24 dev ARM${USER}
+
+## Setup network in normal world linux (paste this in telnet 5000 console)
+
+Replace `106.10.9.180` with some DNS server that works for you.
+
+    ip link set eth0 up
+    ip addr add 192.168.10.40/24 dev eth0
```
bokdeuk-jeong, line 238: IP addresses used here for both the host in FVP and the guest conflict the configuration in fvp-cca for linux-net. These should be consistent with the configuration in fvp-cca which has already existed.

.gitmodules
p-sawicki2: It will be removed.

.gitmodules
p-sawicki2: > Why do we need another branch for eac5?

The eac5/3rd-kvmtool branch keeps the original kvmtool source code from the [Arm repository](https://gitlab.arm.com/linux-arm/kvmtool-cca) that I didn't want to modify during the development of the realm-metadata feture, as the main Islet branch points to it.

> Couldn't we just update and use the existing eac5/3rd-kvmtool?

It should be possible, as the realm-metadata feature is optional.

.gitmodules
bokdeuk-jeong: Should this PR be chaged to draft? If not, these should be removed, before getting merged.

.gitmodules
p-sawicki2: > Please do not point any devel or personal branches.

This also will be changed.

> Why don't you directly update eac5/3rd-realm-linux?

It's because this change will cause issues with rsi.ko on the main Islet branch.

.gitmodules
bokdeuk-jeong: If we had to keep the original kvmtool, it should directly point to the original repo instead of locally keeping it inside Islet. 
Unless it affects the rest of Islet that don't use app-provisiong(container image), please feel free to change eac5/3rd-kvmtool.

examples/app-provisioning/README.md
L0czek: Sure, would you like to just build the kernel with special config? This would leave us with the need to build the initramfs from realm-manager repo.

examples/app-provisioning/README.md
L0czek: Sure, could we also change the ip range utilized by fvp-cca script. Currently, it uses a public ip subnet `193.168.10.1/24`. Is there a reason for this or can we change it to a private one?

.gitmodules
p-sawicki2: We can totally remove these changes from the Islet repo, or create a dedicated tf-rmm repository within the Islet organization and put them there. The question is, are you interested of having tf-rmm that introduce realm metadata and sealing key derivation mechanism? As you know Islet RMM already implements all these features, so the only reason why we should keep them in Islet organization is that they might be useful during investigation of the #368.

examples/app-provisioning/README.md
p-sawicki2: The reason is that it is not referenced by any submodule in the `islet/third-party`. Anyway, the changes from `eac5/3rd-kvmtool-realm-metadata-rim-measurer` will be moved to the `eac5/3rd-kvmtool-rim-measurer` branch and an appropriate submodule from `islet/third-party` will point that new branch.

examples/app-provisioning/README.md
jinbpark: @L0czek No special reason for that address (193.168.x.x).  You can change it to 192.168.x.x.

jinbpark:
Could we get a more developer-friendly way (not islet developers, those who want to build apps but are not knowledgeable about CCA internals) to compile, install, and run a new application?? (it might be good for testing purpose in the beginning of app development)

For example, from the develop perspective,
1. Develop and compile a new application (call this binary `light_app`)
2. ./scripts/fvp-cca ..... --test_app=/path/to/light_app
3. `fvp-cca` takes all the complexities of installing and running this app. (e.g., signing, running a host daemon, ...)

I don't think it needs to be done in this PR. But this option would be nice for external developers.
(+ Once we get this PR merged, we'd better to migrate our examples (confidential-ml, e2ee) into this new provisioning scheme)

jinbpark:
Do you have anything, in your mind, having to do with handling multiple applications?
I found in host_daemon_details document that it seems one realm has multiple applications provisioned by the realm daemon running inside the realm. 

Do you assume that one device has only one realm daemon (i.e., one realm VM) and all applications should be provisioned into this realm? Or is it flexible?? (It would be flexible I guess.  Each application might take up different realm VM)

One more thing I'm wondering is,  is there any isolation mechanism between applications and the realm daemon?
I think the daemon is a typical linux process so the isolation between them is the same as what Linux offers to applications. Am I correct?

Would it be beneficial to employ a stronger isolation? e.g., the realm daemon is isolated from other applications and even underlying linux kernel, meaning that the realm daemon is secure even the kernel is compromised. 
A new CCA feature, named CCA planes, can be used to achieve this isolation.

rmm/src/rmi/realm/mod.rs
```diff
@@ -147,6 +169,46 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
 
         Ok(())
     });
+
+    listen!(
+        mainloop,
+        rmi::ISLET_REALM_SET_METADATA,
```
jinbpark, line 175: Do we need some additional validation checks for this RMI?

For example, REALM_SET_METADATA is only allowed to a realm whose state is in REALM_NEW, which means this should be called before the realm execution. If this is mainly for realm authentication, it should be done this way IMO. (but hard to come up with what actually goes wrong if without such validation checks.)

p-sawicki2: @Havner Could you please add an additional check whether the RD is in a NEW state? (similarly as here https://github.com/p-sawicki2/tf-rmm/blob/294ee8d4ab8fcad46228ea57761f1806a2eff52b/runtime/rmi/metadata.c#L73C6-L73C25)

Havner: Done

L0czek:
@jinbpark 
> Could we get a more developer-friendly way (not islet developers, those who want to build apps but are not knowledgeable about CCA internals) to compile, install, and run a new application?? (it might be good for testing purpose in the beginning of app development)
> [...]
> I don't think it needs to be done in this PR. But this option would be nice for external developers.

Sure, we should redesign the build system to automate stuff. I just think this is a task for another PR as it would most likely require reworking the build system. We should focus now on reviewing and merging this PR and then later discuss how we want to keep various compenents and then remake to build system.

> Do you have anything, in your mind, having to do with handling multiple applications?
I found in host_daemon_details document that it seems one realm has multiple applications provisioned by the realm daemon running inside the realm.
> Do you assume that one device has only one realm daemon (i.e., one realm VM) and all applications should be provisioned into this realm? Or is it flexible?? (It would be flexible I guess. Each application might take up different realm VM)

Current design assumes that you can have multiple realms on a signel device and a single realm can run multiple applications. Naturally, we are actively researching the security concerns resulting from running multiple applications insiade a realm. For now we came to the conclusion with @p-sawicki2 that enforcing that all aplications in a signel realm must be from a single vendor is enough. This would eliminate potential security issues arising from adding an application to an existing realm that could have access to existing confidential data of other applications. We assume that applications from a single vendor will not try to hack themselves.

> One more thing I'm wondering is, is there any isolation mechanism between applications and the realm daemon?
I think the daemon is a typical linux process so the isolation between them is the same as what Linux offers to applications. Am I correct?
> Would it be beneficial to employ a stronger isolation? e.g., the realm daemon is isolated from other applications and even underlying linux kernel, meaning that the realm daemon is secure even the kernel is compromised.
> A new CCA feature, named CCA planes, can be used to achieve this isolation.

Currently, there is no real isolation between applications and the realm daemon. It was a deliberate design choice due to two main reasons:
* We wanted to provide a setup that will be as minimal as possible. Thus, adding any isloation would complicate the design and potentially affect application compatibility and performance.
* The realms already provide isloation as they are separate vms and each application or use case on the host side should use its own realm.
* You can always provision an OCI container with containerd and for ex. docker-compose and just start whatever application you want with isolation based on namespaces.
* After the application were installed (before the are started) the realm daemon doesn't really keep any secrets. For example, the AES encryption keys are kept in kernel keyring and thus cannot be read by any user application (there is a special key type for an "use only and no reading" key).

In conclusion, we tried to make this design as simple as possible and not including application isolation was a conciousness decision. Although, nothing is really stopping you from provisioning an OCI image which implements namespaces. In the future, we might look at the CCA planes feature but for now I think that namespaces are sufficient when isolation is needed.

examples/app-provisioning/README.md
L0czek: I changes the ip addresses to `192.168.10.1/24` and removed now unnecessary steps from the instruction. Could you please check if it works also on your PC?

jinbpark: @L0czek I've confirmed that it works on my PC. (e.g., pinging from realm to host)

.gitmodules
p-sawicki2: Done

.gitmodules
p-sawicki2: Done

rmm/src/rsi/mod.rs
```diff
@@ -358,4 +362,34 @@ pub fn set_event_handler(rsi: &mut RsiHandle) {
 
     listen!(rsi, IPA_STATE_GET, get_ripas_state);
     listen!(rsi, IPA_STATE_SET, set_ripas_state);
+
+    listen!(rsi, ISLET_REALM_SEALING_KEY, |_arg, ret, _rmm, rec, _| {
```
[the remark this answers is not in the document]
p-sawicki2: Done

.gitmodules
p-sawicki2: Done

examples/app-provisioning/README.md
p-sawicki2: Done

.gitmodules
p-sawicki2: @bokdeuk-jeong What is your proposal regarding the TF-RMM?

jinbpark (approved):

.gitmodules
bokdeuk-jeong: @p-sawicki2 
Sorry for the delayed response.
"The question is, are you interested of having tf-rmm that introduces realm metadata and sealing key derivation mechanisms?" Yes!

p-sawicki2: done

bokdeuk-jeong (approved):

PR #366, opened 2024-09-27
Fix spec violations in realm create

zpzigi754:
This fixes rtt_state in B4.3.9.2 and rtt_state in B4.3.9.3, found during the process of verification. The state should be checked and updated by the number of `rtt_num_start`. This PR also removes hundreds of annoying data aborts: I've handled it with a rather radical approach (mapping without unmapping), because using unmapping would affect too many interfaces. I will fix the below undestroyed issue and use the unmapping in another PR. 

```
B4.3.9.2 Failure conditions
rtt_state pre: !RttsStateEqual(
                    params.rtt_base, params.rtt_num_start, DELEGATED)
          post: ResultEqual(result, RMI_ERROR_INPUT)

B4.3.9.3 Success conditions
rtt_state RttsStateEqual(
              Realm(rd).rtt_base, Realm(rd).rtt_num_start, RTT)
```

```
[before]
[TRACE]islet_rmm::rmi::realm::params -- Params { flags: lpa2: 0 sve: 0 pmu: 0, s2sz: 34, sve_v1: 0, num_bps: 0, num_wps: 0, p
[DEBUG]islet_rmm::exception::trap::syndrome -- Data Abort without a change in Exception level
[DEBUG]islet_rmm::exception::trap -- Data Abort (higher), far:88500000
[DEBUG]islet_rmm::exception::trap -- translation, level:3, esr:96000047
[DEBUG]islet_rmm::exception::trap::syndrome -- Data Abort without a change in Exception level
[DEBUG]islet_rmm::exception::trap -- Data Abort (higher), far:88501000
[DEBUG]islet_rmm::exception::trap -- translation, level:3, esr:96000047
[DEBUG]islet_rmm::exception::trap::syndrome -- Data Abort without a change in Exception level
[DEBUG]islet_rmm::exception::trap -- Data Abort (higher), far:88502000
[DEBUG]islet_rmm::exception::trap -- translation, level:3, esr:96000047
[TRACE]islet_rmm::event -- RMI: REALM_CREATE         [88521000, 88524000] > [0, 0]
```

```
[after]
[TRACE]islet_rmm::rmi::realm::params -- Params { flags: lpa2: 0 sve: 0 pmu: 0, s2sz: 32, sve_v1: 0, num_bps: 0, num_wps: 0, p
[TRACE]islet_rmm::event -- RMI: REALM_CREATE         [883CD000, 883D0000] > [0, 0]
 ```

PR #371, opened 2024-10-07
Fix a spec violation in realm destroy

zpzigi754:
This PR fixes a spec violation regarding rtt_state in B4.3.10.3, found during the process of verification. As a result, it fixes the previous issue on ACS tests (mm_rtt_level_start and mm_ha_hd_access), described in #366.

The state should be updated by the number of `rtt_num_start`.

```
B4.3.10.3 Success conditions
rtt_state RttsStateEqual(
              realm.rtt_base, realm.rtt_num_start, DELEGATED)
```

### DIFF
--- a/rmm/src/realm/rd.rs
+++ b/rmm/src/realm/rd.rs
@@ -23,6 +23,7 @@ pub struct Rd {
     vmid: u16,
     state: State,
     rtt_base: usize,
+    rtt_num_start: usize,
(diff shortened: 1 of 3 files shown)


bitboom (approved):
Really nice !

bokdeuk-jeong (approved):

hihi-wang (approved):

PR #372, opened 2024-10-07
Cover all RMIs with MIRI testing

bitboom:
This PR aims to introduce MIRI testing across all RMI functions to strengthen memory safety within RMM, particularly by addressing potential issues arising from `unsafe code`.

To cover all 23 RMI types, 11 new test cases (TCs) have been added (_see Appendix: MIRI Test Cases for details_), leading to the **discovery of 2 bugs** (1 memory leak, 1 integer overflow). Additionally, while writing the MIRI test cases, a manual review revealed a misalignment within the RMM specification, which was reported to Arm.

## How to run
```sh
 $ ./scripts/tests/miri.sh
running 22 tests
test r#macro::test::eprintln_with_format ... ok
test r#macro::test::eprintln_without_arg ... ok
test r#macro::test::eprintln_without_format ... ok
test r#macro::test::println_with_format ... ok
test r#macro::test::println_without_arg ... ok
test r#macro::test::println_without_format ... ok
test r#macro::test::set_of_const_assert ... ok
test rmi::features::test::rmi_features ... ok
test rmi::gpt::test::rmi_granule_delegate_negative ... ok
test rmi::gpt::test::rmi_granule_delegate_positive ... ok
test rmi::gpt::test::rmi_granule_undelegate ... ok
test rmi::realm::test::rmi_realm_create_negative ... ok
test rmi::realm::test::rmi_realm_create_positive ... ok
test rmi::rec::handlers::test::rmi_rec_create_positive ... ok
test rmi::rec::handlers::test::rmi_rec_enter_positive ... ok
test rmi::rtt::test::rmi_data_create_positive ... ok
test rmi::rtt::test::rmi_rtt_create_positive ... ok
test rmi::rtt::test::rmi_rtt_fold_positive ... ok
test rmi::rtt::test::rmi_rtt_init_ripas_positive ... ok
test rmi::rtt::test::rmi_rtt_map_unprotected_positive ... ok
test rmi::rtt::test::rmi_rtt_set_ripas_positive ... ok
test rmi::version::test::rmi_version ... ok

test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 144.88s
```

### Bug 1: Memory Leak in the RMM Page Table
- **Bug Type**: Memory Leak
- **Cause**: During the process of mapping/unmapping the RMM page table, level 1-3 sub-tables were not released even when they contained no entries.
- **Temporary Solution**: As a temporary measure, the RMM page table is dropped upon completion of the MIRI test.
- **Future Discussion**: Creating and destroying sub-tables in the RMM page table with each GRANULE_DELEGATE call may result in performance degradation. It is necessary to find a solution that addresses the memory leak while minimizing performance impacts.
- **Note**: The RMM page table is managed as a static variable, and in Rust, the `drop()` function for static variables is not invoked. As a result, sub-tables used by RMM continue to allocate memory without being released.

```sh
$ cargo miri test rmi::gpt::test::rmi_granule_delegate

error: memory leaked: alloc56894755 (Rust heap, size: 4096, align: 4096), allocated here:
   --> /home/sangwan/.rustup/toolchains/nightly-2024-04-21-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:100:9
    |
100 |         __rust_alloc(layout.size(), layout.align())
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: BACKTRACE:
    = note: inside `alloc::alloc::alloc` at /home/sangwan/.rustup/toolchains/nightly-2024-04-21-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:100:9: 100:52
    = note: inside `<vmsa::page_table::DefaultMemAlloc as vmsa::page_table::MemAlloc>::allocate` at /home/sangwan/islet/lib/vmsa/src/page_table.rs:63:9: 63:36
    = note: inside `vmsa::page_table::PageTable::<vmsa::address::VirtAddr, mm::page_table::L3Table, mm::page_table::entry::Entry, 512>::new_in` at /home/sangwan/islet/lib/vmsa/src/page_table.rs:159:13: 159:92
    = note: inside `<vmsa::page_table::PageTable<vmsa::address::VirtAddr, mm::page_table::L2Table, mm::page_table::entry::Entry, 512> as vmsa::page_table::PageTableMethods<vmsa::address::VirtAddr, mm::page_table::L2Table, mm::page_table::entry::Entry, 512>>::set_page::<mm::page::BasePageSize>` at /home/sangwan/islet/lib/vmsa/src/page_table.rs:392:25: 394:26
    = note: inside `<vmsa::page_table::PageTable<vmsa::address::VirtAddr, mm::page_table::L1Table, mm::page_table::entry::Entry, 512> as vmsa::page_table::PageTableMethods<vmsa::address::VirtAddr, mm::page_table::L1Table, mm::page_table::entry::Entry, 512>>::set_page::<mm::page::BasePageSize>` at /home/sangwan/islet/lib/vmsa/src/page_table.rs:400:13: 400:50
    = note: inside `<vmsa::page_table::PageTable<vmsa::address::VirtAddr, mm::page_table::L1Table, mm::page_table::entry::Entry, 512> as vmsa::page_table::PageTableMethods<vmsa::address::VirtAddr, mm::page_table::L1Table, mm::page_table::entry::Entry, 512>>::set_pages::<mm::page::BasePageSize>` at /home/sangwan/islet/lib/vmsa/src/page_table.rs:215:19: 215:
```

### Bug 2: Integer Overflow in RMI_RTT_SET_RIPAS
- **Bug Type**: Integer Overflow
- **Cause**: An overflow occurs when the `top` in `RMI_RTT_SET_RIPAS` is smaller than `GRANULE_SIZE`.
- **Solution**: Implemented an overflow check and return an appropriate error when detected.

```sh
note: inside closure
   --> rmm/src/rmi/rtt.rs:649:36
    |
648 |     #[test]
    |     ------- in this procedural macro expansion
649 |     fn rmi_rtt_set_ripas_positive() {
    |                                    ^
    = note: this warning originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)

thread 'rmi::rtt::test::rmi_rtt_set_ripas_positive' panicked at rmm/src/rmi/rtt.rs:138:32:
attempt to subtract with overflow
```
## Spec Feedback: Ensuring realm_state validation in B4.3.15

It seems necessary to review between B4.3.15 and D1.2.2 for spec consistency.

1. Specification Restrictions
According to "D1.2.2 Realm Translation Table creation flow" subsequent levels of RTT
can be created when the state of the Realm is REALM_NEW or REALM_ACTIVATE.

> D1.2.2 Realm Translation Table creation flow
> Subsequent levels of RTT are added using the RMI_RTT_CREATE command.
> This can be performed when the state of the Realm is REALM_NEW or REALM_ACTIVE.

2. Missing Failure Conditions
However in "B4.3.15 RMI_RTT_CREATE", the condition to check the realm_state is missing.

3. Suggestion
It would be necessary either to add this realm_state validation within "B4.3.15.2 Failure conditions"
or to clarify the details in D1.2.2 for better alignment.

## Process of Applying MIRI Test
1. Extract inputs and outputs of the RMI to be tested from positive test cases in the ACS tests.
2. Write the MIRI test.
3. Execute the MIRI test.
4. Implement mocking for parts requiring simulation.
5. Apply patches for any bugs encountered.

### Handling Provenance Issues with Host-Provided Memory
- **Problem**: When using MIRI to test Rust code, "No provenance" errors may occur 1) during operations like `core::ptr::write` or 2) when converting externally allocated memory into specific Rust objects. This is because the memory addresses, not allocated by RMM, lack the required provenance that MIRI enforces.
- **Limitation**: Currently, MIRI does not allow the disabling of provenance checks. Additionally, Rust does not natively support allocating memory at specific addresses, complicating this scenario further.
- **Proposed Alternative**: Instead of using externally provided memory addresses directly, simulate these addresses by allocating memory within RMM’s managed space and treat it as if it were external memory. This approach avoids provenance issues while allowing MIRI testing.

### Realm Simulation
- **Problem**: MIRI cannot execute assembly code, making it impossible to directly enter the realm context.
- **Solution**: Write mocking for the necessary parts, simulating the execution of the realm and the invocation of RSI as if it were happening in the test.

## Remaining Works
1. **Enable Stacked Borrows Check**: Currently, the MIRI rule for "Stacked Borrows" is disabled as it is experimental. Plan to enable this check.
2. **Add Negative Test Cases**: In addition to the positive test cases, add negative test cases to ensure more comprehensive testing.

## Appendix: MIRI Test Cases
| Index | Command                   | Test Case                                        | Test Result | Note                                           |
| ----- | ------------------------- | ------------------------------------------------ | ----------- | ---------------------------------------------- |
| 1     | RMI_VERSION               | rmi::version::test::rmi_version                  | PASS        |                                                |
| 2     | RMI_GRANULE_DELEGATE      | rmi::gpt::test::rmi_granule_delegate_positive    | PASS        | Bug Found: Memory Leak in RMM Page Table       |
| 3     | RMI_GRANULE_UNDELEGATE    | rmi::gpt::test::rmi_granule_delegate_positive    | PASS        |                                                |
| 4     | RMI_DATA_CREATE           | rmi::rtt::test::rmi_data_create_positive         | PASS        |                                                |
| 5     | RMI_DATA_CREATE_UNKNOWN   | rmi::rtt::test::rmi_data_create_positive         | PASS        |                                                |
| 6     | RMI_DATA_DESTROY          | rmi::rtt::test::rmi_data_create_positive         | PASS        |                                                |
| 7     | RMI_REALM_ACTIVATE        | rmi::realm::test::rmi_realm_create_positive      | PASS        |                                                |
| 8     | RMI_REALM_CREATE          | rmi::realm::test::rmi_realm_create_positive      | PASS        |                                                |
| 9     | RMI_REALM_DESTROY         | rmi::realm::test::rmi_realm_create_positive      | PASS        |                                                |
| 10    | RMI_REC_CREATE            | rmi::realm::test::rmi_rec_create_positive        | PASS        |                                                |
| 11    | RMI_REC_DESTROY           | rmi::realm::test::rmi_rec_create_positive        | PASS        |                                                |
| 12    | RMI_REC_ENTER             | rmi::rec::handlers::test::rmi_rec_enter_positive | PASS        |                                                |
| 13    | RMI_RTT_CREATE            | rmi::realm::test::rmi_rtt_create_positive        | PASS        | Spec Feedback: Misalignment in the realm state |
| 14    | RMI_RTT_DESTROY           | rmi::realm::test::rmi_rtt_create_positive        | PASS        |                                                |
| 15    | RMI_RTT_MAP_UNPROTECTED   | rmi::rtt::test::rmi_rtt_map_unprotected_positive | PASS        |                                                |
| 16    | RMI_RTT_READ_ENTRY        | rmi::realm::test::rmi_rtt_create_positive        | PASS        |                                                |
| 17    | RMI_RTT_UNMAP_UNPROTECTED | rmi::rtt::test::rmi_rtt_map_unprotected_positive | PASS        |                                                |
| 18    | RMI_PSCI_COMPLETE         | rmi::rec::handlers::test::rmi_rec_enter_positive | PASS        |                                                |
| 19    | RMI_FEATURES              | rmi::features::test::rmi_features                | PASS        |                                                |
| 20    | RMI_RTT_FOLD              | rmi::rtt::test::rmi_rtt_fold_positive            | PASS        |                                                |
| 21    | RMI_REC_AUX_COUNT         | rmi::realm::test::rmi_rec_create_positive        | PASS        |                                                |
| 22    | RMI_RTT_INIT_RIPAS        | rmi::realm::test::rmi_rtt_init_positive          | PASS        |                                                |
| 23    | RMI_RTT_SET_RIPAS         | rmi::realm::test::rmi_rtt_set_positive           | PASS        | Bug Found: Integer Overflow                    |

*Related Issue: https://github.com/islet-project/islet/issues/361*

### DIFF
--- a/rmm/src/rmi/gpt.rs
+++ b/rmm/src/rmi/gpt.rs
@@ -13,6 +13,7 @@ use crate::{get_granule, get_granule_if, set_state_and_get_granule};
 use vmsa::error::Error as MmError;
 
 extern crate alloc;
+extern crate std;
 
 // defined in trusted-firmware-a/include/services/rmmd_svc.h
 pub const MARK_REALM: usize = 0xc400_01b0;
@@ -22,16 +23,21 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     #[cfg(any(not(kani), feature = "mc_rmi_granule_delegate"))]
     listen!(mainloop, rmi::GRANULE_DELEGATE, |arg, _, rmm| {
         let addr = arg[0];
-        #[cfg(feature = "gst_page_table")]
-        let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
-            Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated),
-            other => other,
-        }?;
 
-        #[cfg(not(feature = "gst_page_table"))]
-        let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
+        {
+            // Avoid deadlock in get_granule() in smc
+            #[cfg(feature = "gst_page_table")]
+            let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
+                Err(MmError::MmNoEntry) => {
+                    set_state_and_get_granule!(addr, GranuleState::Undelegated)
+                }
+                other => other,
+            }?;
 
-        #[cfg(not(miri))]
+            #[cfg(not(feature = "gst_page_table"))]
+            let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
+        }
+
         if smc(MARK_REALM, &[addr])[0] != SMC_SUCCESS {
             return Err(Error::RmiErrorInput);
         }
@@ -40,6 +46,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // `page_table` is currently not reachable in model checking harnesses
         rmm.page_table.map(addr, true);
 
+        let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
         set_granule(&mut granule, GranuleState::Delegated).map_err(|e| {
             #[cfg(not(kani))]
             // `page_table` is currently not reachable in model checking harnesses
@@ -82,10 +89,21 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
 
 #[cfg(test)]
 mod test {
+    use crate::granule::array::granule_addr;
+    use crate::rmi::gpt::GranuleState;
     use crate::rmi::{ERROR_INPUT, GRANULE_DELEGATE, GRANULE_UNDELEGATE, SUCCESS};
     use crate::test_utils::*;
+    use crate::{get_granule, get_granule_if};
 
     use alloc::vec;
+
+    #[test]
+    fn rmi_granule_delegate_positive() {
+        let mocking_addr = granule_addr(0);
+        let ret = rmi::<GRANULE_DELEGATE>(&[mocking_addr]);
+        assert_eq!(ret[0], SUCCESS);
+        assert!(get_granule_if!(mocking_addr, GranuleState::Delegated).is_ok());
+    }
 
     // Source: https://github.com/ARM-software/cca-rmm-acs
     // Test Case: cmd_granule_delegate
@@ -102,7 +120,7 @@ mod test {
        Check 10 : gran_gpt; intent id : 0x9 addr : 0x6000000
     */
     #[test]
-    fn rmi_granule_delegate() {
+    fn rmi_granule_delegate_negative() {
         let test_data = vec![
             (0x88303000, SUCCESS),
             (0x88300001, ERROR_INPUT),
--- a/rmm/src/rmi/gpt.rs
+++ b/rmm/src/rmi/gpt.rs
@@ -23,19 +23,17 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     listen!(mainloop, rmi::GRANULE_DELEGATE, |arg, _, rmm| {
         let addr = arg[0];
 
-        {
-            // Avoid deadlock in get_granule() in smc
-            #[cfg(feature = "gst_page_table")]
-            let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
-                Err(MmError::MmNoEntry) => {
-                    set_state_and_get_granule!(addr, GranuleState::Undelegated)
-                }
-                other => other,
-            }?;
+        #[cfg(feature = "gst_page_table")]
+        let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
+            Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated),
+            other => other,
+        }?;
+        #[cfg(not(feature = "gst_page_table"))]
+        let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
 
-            #[cfg(not(feature = "gst_page_table"))]
-            let _granule = get_granule_if!(addr, GranuleState::Undelegated)?;
-        }
+        // Avoid deadlock in get_granule() in smc() on {miri, test} mode
+        #[cfg(any(miri, test))]
+        core::mem::drop(granule);
 
         if smc(MARK_REALM, &[addr])[0] != SMC_SUCCESS {
             return Err(Error::RmiErrorInput);
@@ -45,6 +43,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // `page_table` is currently not reachable in model checking harnesses
         rmm.page_table.map(addr, true);
 
+        #[cfg(any(miri, test))]
         let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
         set_granule(&mut granule, GranuleState::Delegated).map_err(|e| {
             #[cfg(not(kani))]
@@ -61,10 +60,11 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     #[cfg(any(not(kani), feature = "mc_rmi_granule_undelegate"))]
     listen!(mainloop, rmi::GRANULE_UNDELEGATE, |arg, _, rmm| {
         let addr = arg[0];
-        {
-            // Avoid deadlock in get_granule() in smc
-            let _granule = get_granule_if!(addr, GranuleState::Delegated)?;
-        }
+        let mut granule = get_granule_if!(addr, GranuleState::Delegated)?;
+
+        // Avoid deadlock in get_granule() in smc() on {miri, test} mode
+        #[cfg(any(miri, test))]
+        core::mem::drop(granule);
 
         if smc(MARK_NONSECURE, &[addr])[0] != SMC_SUCCESS {
             panic!(
@@ -73,6 +73,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             );
         }
 
+        #[cfg(any(miri, test))]
         let mut granule = get_granule_if!(addr, GranuleState::Delegated)?;
 
         #[cfg(not(kani))]
--- a/rmm/src/rmi/gpt.rs
+++ b/rmm/src/rmi/gpt.rs
@@ -34,7 +34,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             }?;
 
             #[cfg(not(feature = "gst_page_table"))]
-            let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
+            let _granule = get_granule_if!(addr, GranuleState::Undelegated)?;
         }
 
         if smc(MARK_REALM, &[addr])[0] != SMC_SUCCESS {
@@ -63,7 +63,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let addr = arg[0];
         {
             // Avoid deadlock in get_granule() in smc
-            let granule = get_granule_if!(addr, GranuleState::Delegated)?;
+            let _granule = get_granule_if!(addr, GranuleState::Delegated)?;
         }
 
         if smc(MARK_NONSECURE, &[addr])[0] != SMC_SUCCESS {
@@ -93,7 +93,6 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
 
 #[cfg(test)]
 mod test {
-    use crate::granule::array::granule_addr;
     use crate::rmi::gpt::GranuleState;
     use crate::rmi::{ERROR_INPUT, GRANULE_DELEGATE, GRANULE_UNDELEGATE, SUCCESS};
     use crate::test_utils::*;
@@ -103,7 +102,7 @@ mod test {
 
     #[test]
     fn rmi_granule_delegate_positive() {
-        let mocking_addr = granule_addr(0);
+        let mocking_addr = alloc_granule(0);
         let ret = rmi::<GRANULE_DELEGATE>(&[mocking_addr]);
         assert_eq!(ret[0], SUCCESS);
         assert!(get_granule_if!(mocking_addr, GranuleState::Delegated).is_ok());
--- a/rmm/src/rmi/gpt.rs
+++ b/rmm/src/rmi/gpt.rs
@@ -13,7 +13,6 @@ use crate::{get_granule, get_granule_if, set_state_and_get_granule};
 use vmsa::error::Error as MmError;
 
 extern crate alloc;
-extern crate std;
 
 // defined in trusted-firmware-a/include/services/rmmd_svc.h
 pub const MARK_REALM: usize = 0xc400_01b0;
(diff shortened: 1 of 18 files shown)


jinbpark:
Regarding "Memory Leak in the RMM Page Table",  I think the simplest way is to drop unnecessary tables when a realm gets activated. A lot of RMI calls are not allowed for an activated realm (e.g., DATA_CREATE) so it's reasonable to say that it's unlikely further RMM page table mappings are needed for the realm.

1. Realm-A in the launch phase (e.g., granule_delegate & data_create) --> create unnecessary tables for the PA range 0x10000-0x20000 for example.
2. Realm-A gets activated --> clear out unnecessary tables  --> reasonable, it's unlikely that the host requests the same PA range 0x10000--0x20000 (other realms are likely to call RMIs with different physical memory ranges)

Benefit? we can avoid performance overheads of creating and deleting sub tables while launching Realm-A (before being activated).

bitboom:
> Regarding "Memory Leak in the RMM Page Table", I think the simplest way is to drop unnecessary tables when a realm gets activated. A lot of RMI calls are not allowed for an activated realm (e.g., DATA_CREATE) so it's reasonable to say that it's unlikely further RMM page table mappings are needed for the realm.
> 
> 1. Realm-A in the launch phase (e.g., granule_delegate & data_create) --> create sub tables for the PA range 0x10000-0x20000 for example.
> 2. Realm-A gets activated --> clear out unnecessary sub tables  --> reasonable, it's unlikely that the host requests the same PA range 0x10000--0x20000 (other realms are likely to call RMIs with different physical memory ranges)
> 
> Benefit? we can avoid performance overheads of creating and deleting sub tables while launching Realm-A (before being activated).

The proposed approach seems practical and reasonable. As suggested, we can break down the process (1. delegate & data-create , 2. realm-activate) and remove unnecessary sub-tables at specific stages (when the realm is activated). To achieve this, we will need to store information related to the previously used Realm, but this shouldn't pose a significant problem.

However, I have one concern; if someone intentionally does not activate the Realm and repeatedly triggers only the previous stage (delegate & data-create), it still could lead to the exhaustion of RMM's memory. It would be beneficial to include a solution for this scenario as well.

jinbpark:
> if someone intentionally does not activate the Realm 

This is crucial. Agree, we should figure out the way to deal with this. 
Just setting up a condition that triggers "dropping sub tables" suffices to do this. (then, we don't need the function to drop subtables when it gets activated)

(1) keeps track of the number of sub tables;
(2) when the number exceeds a threshold value T (statically configurable), it calls the function to drop sub tables;

jinbpark (approved):
Good to see the miri integration-!

rmm/src/rmi/gpt.rs
```diff
@@ -22,20 +22,30 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
-        #[cfg(feature = "gst_page_table")]
-        let mut granule = match get_granule_if!(addr, GranuleState::Undelegated) {
-            Err(MmError::MmNoEntry) => set_state_and_get_granule!(addr, GranuleState::Undelegated),
-            other => other,
-        }?;
-        #[cfg(not(feature = "gst_page_table"))]
-        let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
+
+        {
+            // Avoid deadlock in get_granule() in smc
```
nook1208, line 27: Could you explain more about this ?? 
Does it mean SMC failed condition ?

rmm/src/rmi/rtt.rs
```diff
@@ -134,6 +134,7 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         if !is_granule_aligned(base)
             || !is_granule_aligned(top)
             || !rd.addr_in_par(base)
+            || top.checked_sub(GRANULE_SIZE).is_none()
```
nook1208, line 137: Good to add this overflow check !

rmm/src/rmi/gpt.rs
bitboom: Deadlock occurred only in test mode (kani, test, miri). I fixed it more accurately in https://github.com/islet-project/islet/pull/372/commits/9af3532ae749681dc106cb0d513db253f683e509.

nook1208 (approved):
Good to add MIRI tests !

rmm/src/rmi/gpt.rs
```diff
@@ -53,13 +64,20 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+
         if smc(MARK_NONSECURE, &[addr])[0] != SMC_SUCCESS {
             panic!(
                 "A delegated granule should only be undelegated on request from RMM. {:X}",
                 addr
             );
         }
 
+        #[cfg(any(kani, miri, test))]
+        let mut granule = get_granule_if!(addr, GranuleState::Delegated)?;
```
zpzigi754, line 79: kani should be excluded if drop is excluded for kani.

rmm/src/rmi/gpt.rs
```diff
@@ -22,20 +22,31 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+                set_state_and_get_granule!(addr, GranuleState::Undelegated)
+            }
             other => other,
         }?;
         #[cfg(not(feature = "gst_page_table"))]
         let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
 
+        // Avoid deadlock in get_granule() in smc() on {kani, miri, test} mode
+        #[cfg(any(kani, miri, test))]
+        core::mem::drop(granule);
```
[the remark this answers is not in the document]
bitboom: Thank you for checking! I fixed it :)

rmm/src/rmi/gpt.rs
```diff
@@ -22,20 +22,31 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
         if smc(MARK_REALM, &[addr])[0] != SMC_SUCCESS {
             return Err(Error::RmiErrorInput);
         }
+
         #[cfg(not(kani))]
         // `page_table` is currently not reachable in model checking harnesses
         rmm.page_table.map(addr, true);
+
+        #[cfg(any(kani, miri, test))]
+        let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
```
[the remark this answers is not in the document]
bitboom: I excluded `kani` here as well.

rmm/src/rmi/gpt.rs
```diff
@@ -53,13 +64,20 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         let addr = arg[0];
         let mut granule = get_granule_if!(addr, GranuleState::Delegated)?;
 
+        // Avoid deadlock in get_granule() in smc() on {kani, miri, test} mode
+        #[cfg(any(kani, miri, test))]
+        core::mem::drop(granule);
```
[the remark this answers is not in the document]
bitboom: I fixed it too :)

rmm/src/rmi/gpt.rs
```diff
@@ -22,20 +22,31 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
...
+                set_state_and_get_granule!(addr, GranuleState::Undelegated)
+            }
             other => other,
         }?;
         #[cfg(not(feature = "gst_page_table"))]
         let mut granule = get_granule_if!(addr, GranuleState::Undelegated)?;
 
+        // Avoid deadlock in get_granule() in smc() on {miri, test} mode
+        #[cfg(any(miri, test))]
+        core::mem::drop(granule);
```
zpzigi754, line 38: Could you explain in detail how the deadlock occurred (e.g., in which scenario)?

rmm/src/rmi/gpt.rs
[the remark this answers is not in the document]
zpzigi754: This one doesn't seem to have excluded `kani` (only this one seems to be missing). Could you please check this again?

rmm/src/rmi/gpt.rs
bitboom: Sure!

1. Acquire a lock with addr in GRANULE_DELEGATE (`let mut granule = get_granule_if!(addr, GranuleState::Undelegated)`, *in rmi/gpt.rs*)
2. call smc() 
3. Acquire the same lock in smc ( `let gpt = get_granule!(addr).map(|guard| guard.gpt).unwrap();` , *in asm.rs*)

rmm/src/rmi/gpt.rs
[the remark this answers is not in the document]
bitboom: My mistake. I'm sorry :( I fixed it!

rmm/src/rmi/gpt.rs
zpzigi754: Oh, I see. Thank you for the answer!

zpzigi754: It might be better to access the object without the locking, as the gpt's access is a ghost operation.

zpzigi754 (approved):

PR #373, opened 2024-10-07
Add model checking harnesses for rmi version and rmi realm activate

zpzigi754:
This PR adds model checking harnesses for rmi version and rmi realm activate: they verify `RMI_VERSION` and `RMI_REALM_ACTIVATE`'s behaviors respectively. It also fixes a spec violation in rmi version regarding B4.3.23.1.2, which was detected by the harness. Output values (lower and higher) should be always available regardless of this RMI's failure.

```
Usage)

(in islet/model-checking)
$RMI_TARGET=(rmi_version|rmi_realm_activate) make verify
```

```
B4.3.23.1.2 Output values
Name    Register   Bits       Type                      Description
result  X0         63:0       RmiCommandReturnCode      Command return status
lower   X1         63:0       RmiInterfaceVersionLower  implemented interface revision
higher  X2         63:0       RmiInterfaceVersionHigher implemented interface revision
```


rmm/src/rmi/version.rs
```diff
@@ -19,6 +19,12 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
     listen!(mainloop, rmi::VERSION, |arg, ret, _| {
         let req = arg[0];
 
+        let lower = encode_version();
```
hihi-wang, line 22: Just bit wondering here, does this lines of code affects in verification, such like in case of returning `Err(Error::RmiErrorInput)`, `ret` should be filled with version info...?

hihi-wang (approved):

rmm/src/rmi/version.rs
zpzigi754: Exactly! Setting the `ret` values ahead (ret[1] and ret[2]) affects the verification result (from failure to success). The reason is because, ret[1] and ret[2] will be set not only for the success condition but also for the failure condition (the below Error::RmiErrorInput).

jinbpark (approved):

bitboom (approved):

PR #374, opened 2024-10-08
Add model checking harnesses for rec_aux_count and rec_destroy

zpzigi754:
This PR adds model checking harnesses for rec_aux_count and rec_destroy: they verify RMI_REC_AUX_COUNT and RMI_REC_DESTROY's behaviors respectively. It also fixes several spec violations two of which were detected by the included harness while others were found by the process of verification.

```
B4.3.12.2 Failure conditions
ID      Condition
num_aux pre: params.num_aux != RecAuxCount(rd)
        post: ResultEqual(result, RMI_ERROR_INPUT)
```

```
B4.3.12.3 Success conditions
ID            Condition
rec_aux       AuxEqual(
                  Rec(rec).aux, params.aux,
                  RecAuxCount(rd))
rec_aux_state AuxStateEqual(
                  Rec(rec).aux, RecAuxCount(rd), REC_AUX)
```

```
B4.3.13.2 Failure conditions
ID        Condition
rec_state pre: Rec(rec).state == REC_RUNNING
          post: ResultEqual(result, RMI_ERROR_REC)
```

```
B4.3.13.3 Success conditions
ID            Condition
rec_aux_state AuxStateEqual(
                  Rec(rec).aux, RecAuxCount(rd), DELEGATED)
```


rmm/src/rmi/rec/handlers.rs
```diff
@@ -94,15 +95,29 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // `rsi` is currently not reachable in model checking harnesses
         HashContext::new(&mut rd)?.measure_rec_params(&params)?;
 
+        for i in 0..rmi::MAX_REC_AUX_GRANULES {
+            let rec_aux = rec.aux(i) as usize;
+            let mut rec_aux_granule = get_granule_if!(rec_aux, GranuleState::Delegated)?;
+            set_granule(&mut rec_aux_granule, GranuleState::RecAux)?;
```
p-sawicki2, line 101: Does Islet actually use Auxiliary Granules? During the setup of a Realm, KVM asks RMM about the number of auxilary granules (rmi::REC_AUX_COUNT) that RMM needs for data storage. RMM returns MAX_REC_AUX_GRANULES. Basing on that information KVM allocates and passes them while calling REC_CREATE. But they seem not to be used at all in Islet RMM. Why not just return 0 granules while calling rmi::REC_AUX_COUNT?

rmm/src/rmi/rec/handlers.rs
```diff
@@ -94,15 +95,29 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
         // `rsi` is currently not reachable in model checking harnesses
         HashContext::new(&mut rd)?.measure_rec_params(&params)?;
 
+        for i in 0..rmi::MAX_REC_AUX_GRANULES {
```
jinbpark, line 98: Not directly related to this PR, 
it seems that `rd.inc_rec_index();` at line-93 in this function should be placed at the end of this function (after all operations are done correctly, as `set_granule` is).

If something goes wrong in this rec_aux loop, it would return without decrementing rec_index by one, leaving rec_index incorrect.

zpzigi754: Good finding! I totally agree with you. Similarly, I think that most of the setting in `rec.init()` above also might need to be done at the end of the function, because they should be only set in the success condition. One issue is that setting the successful conditions also may involve an error handling which is a bit of mess.

jinbpark: It's not just `rec.init()` but there might be similar issues in other functions as well. (e.g., rmm page table mapping. we should unmap things from the rmm page table if something goes wrong in the middle)
We'd better come up with a kind of design pattern to gracefully deal with such issues.

I think it's ok to put this off until we settle up that pattern, and we can apply that pattern to all relevant codes at once. (this PR might be not that small in size?)

rmm/src/rmi/rec/handlers.rs
zpzigi754: That's a really good question. I think that `returning 0` in rmi::REC_AUX_COUNT would be also a reasonable alternative as a shortcut. I am actually not sure how auxiliary granules would be used eventually with a real world realm image: this is actually in one of my query lists to Arm. However, as long as RecAux Granule is not deprecated in the official arm cca's spec, I think that supporting this functionality would be better than not supporting it. Also, note that RecAux have existed since the earliest beta0 version to the latest alpha8 version in the spec. Thus, I expect that it is less likely to be deprecated in the future as well.

jinbpark (approved):

rmm/src/rmi/rec/handlers.rs
zpzigi754: Thank you for opening the issue there!

hihi-wang (approved):

bitboom (approved):

nook1208 (approved):

PR #376, opened 2024-10-10
Add rec.emulatable_abort to keep track of the last emulatable abort

jinbpark:
Any data abort, happened at an IPA within Protected Address Space, should be considered non-emulatable faults. (otherwise, a malicious host can inject some value into the realm's protected space via MMIO)

To follow this rule in accordance with the spec, this PR adds `rec.emulatable_abort` to keep track of whether the last rec exit was due to an emulatable abort. (spec 1.1-alp8, A4.3.4.3 REC exit due to Data Abort)
Note: this PR doesn't modify esr_el2 that is delivered to the host (via run.exit).

It marks `EmulatableAbort` in `handle_data_abort()` only if an abort happens within unprotected IPA, and this flag gets checked in `emulate_mmio()`.

Regarding ACS test,
tests added in skipped-tests.txt are `mm_ripas_destroyed_da_ia`, `mm_hipas_unassigned_ripas_ram_da_ia`, and `exception_non_emulatable_da_2`.  They are all incompatible with this fix and should be modified accordingly. 
I'll modify acs-test-code and take these 3 tests out of skipped-tests.txt again, after this PR gets meged.


jinbpark:
ACS test failed.  I'll get it back to ready-for-review after looking at the issue.

bitboom:
> a malicious host can inject some value into the realm's protected space via MMIO

Nice finding !!

> Regarding ACS test, tests added in skipped-tests.txt are `mm_ripas_destroyed_da_ia`, `mm_hipas_unassigned_ripas_ram_da_ia`, and `exception_non_emulatable_da_2`. They are all incompatible with this fix and should be modified accordingly. I'll modify acs-test-code and take these 3 tests out of skipped-tests.txt again, after this PR gets meged.

Is it an issue with the current ACS test version we are using, or does the ACS test code need to be modified? _(It seems that the latest ACS test supports 1.1-alp8. https://github.com/ARM-software/cca-rmm-acs/commit/114807429a4663ae09d777e7ea9fac7c06a496db)_

bokdeuk-jeong (approved):

rmm/src/rmi/rec/exit.rs
```diff
@@ -112,19 +114,18 @@ fn handle_data_abort(
     let fault_ipa = hpfar_el2 & (HPFAR_EL2::FIPA.mask << HPFAR_EL2::FIPA.shift);
     let fault_ipa = (fault_ipa << 8) as usize;
 
+    // spec: 1.1-alp8, A4.3.4.3 REC exit due to Data Abort
```
zpzigi754, line 117: I think that for the time being, keeping following eac5 instead of 1.1-alp8 might be better until we move to 1.1-alp8 someday.

rmm/src/rmi/rec/exit.rs
```diff
@@ -113,7 +113,13 @@ fn handle_data_abort(
...
 
     let (exit_esr, exit_far) = match is_non_emulatable_data_abort(&rd, fault_ipa, esr_el2)? {
-        true => (esr_el2 & NON_EMULATABLE_ABORT_MASK, 0),
+        true => {
+            // we need to clear out the ISV bit by hand in rec (the esr of the last execution) and run.exit
+            // to inform both RMM (via rec) and Host (via run.exit) that this abort is not emulatable.
+            let mut rec_esr_el2 = EsrEl2::new(rec.context.sys_regs.esr_el2);
+            rec.context.sys_regs.esr_el2 = rec_esr_el2.clear_bits(EsrEl2::ISV).get();
+            (esr_el2 & NON_EMULATABLE_ABORT_MASK, 0)
+        }
```
zpzigi754: Good to find and fix it!

jinbpark:
> Is it an issue with the current ACS test version we are using, or does the ACS test code need to be modified? 

I checked the latest ACS codes and tf-rmm codes.  tf-rmm has fixed this issue but acs test codes has not yet. So, I guess the latest ACS code is not compatible with both islet and tf-rmm. (I expect the acs code will be updated accordingly, though)

bitboom (approved):
We may add this (the vulnerability & related patch) to our developer site :)
This seems like useful information for VM-based models like CCA.

rmm/src/rmi/rec/exit.rs
jinbpark: Thanks for pointing this out. I have just confirmed that this is the same as in eac5. I'll discard the comment.

nook1208 (approved):
Good finding !

zpzigi754 (approved):

PR #385, opened 2024-10-28
Add a reference count in array-based granule status table

zpzigi754:
This adds a reference count and `num_children` support in array-based granule status table missing of which was the main cause for the previous ACS failures when gst_page_table feature was not enabled. Most of ACS tests will be passed with this fix.

```
[before]
REGRESSION REPORT:
==================
   TOTAL TESTS     : 76
   TOTAL PASSED    : 24
   TOTAL FAILED    : 47
   TOTAL SKIPPED   : 3
   TOTAL SIM ERROR : 2
```

```
[after]
REGRESSION REPORT:
==================
   TOTAL TESTS     : 76
   TOTAL PASSED    : 71
   TOTAL FAILED    : 1
   TOTAL SKIPPED   : 3
   TOTAL SIM ERROR : 1
```

### DIFF
--- a/rmm/src/granule/array/entry.rs
+++ b/rmm/src/granule/array/entry.rs
@@ -2,6 +2,7 @@ use crate::granule::array::GRANULE_STATUS_TABLE;
 use crate::rmi::error::Error;
 
 use super::{GranuleState, GRANULE_SIZE};
+use core::sync::atomic::{AtomicU8, Ordering};
 use safe_abstraction::raw_ptr;
 use spinning_top::{Spinlock, SpinlockGuard};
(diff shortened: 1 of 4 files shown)


bitboom (approved):
Looks Good!

Could we remove the `gst_page_table` feature and proceed solely with an array-based approach? If the memory usage is acceptable, i think it would be better to continue using the method we are currently verifying by `kani`.

Personally, I prefer reducing code size and features, especially since we can restore them any time if necessary.

jinbpark (approved):

nook1208 (approved):

Issue #392, opened 2024-11-01
A booting issue in the veraison example with HES


zpzigi754:
I've tried to follow the steps in [veraison guide](https://github.com/islet-project/islet/blob/main/examples/veraison/RUN.md).

For some reasons, if I run `./scripts/fvp-cca --normal-world=linux --realm=linux --rmm=islet --hes --rmm-log-level info`, the booting does not succeed, as it is stuck with the following logs.

```
NOTICE:  Booting Trusted Firmware
NOTICE:  BL1: v2.10.0   (release):0b72924
NOTICE:  BL1: Built : 16:09:33, Nov  1 2024
NOTICE:  Measured boot extend measurement:
NOTICE:   - slot        : 6
NOTICE:   - signer_id   : 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
NOTICE:   - version     :
NOTICE:   - version_size: 0
NOTICE:   - sw_type     : FW_CONFIG
NOTICE:   - sw_type_size: 10
NOTICE:   - algorithm   : 2000009
NOTICE:   - measurement : 0f 00 bd d3 8f c2 01 e9 94 cb d7 f5 3f 58 18 dd
 33 bd fa 45 15 62 b6 52 70 14 7e ec b7 f4 61 e4
NOTICE:   - locking     : true
NOTICE:  [RSS_SERIAL] protocol: EMBED
NOTICE:  [RSS_SERIAL] Serial initialized
NOTICE:  [RSS_SERIAL] sent 116 bytes
```

If I do not use `--hes` option (e.g., executing `$scripts/tests/realm-boot.sh`), the booting would succeed.

zpzigi754:
It seems that there is an issue between `hes` and `telnet`. If I pass `--no-telnet` option like (`./scripts/fvp-cca --normal-world=linux --realm=linux --rmm=islet --hes --rmm-log-level info --no-telnet`), the stuck symptom disappears (the booting succeeds).

bokdeuk-jeong:
@p-sawicki2 
could you take a look into this?

p-sawicki2:
@Havner Could you please check that?

L0czek:
@zpzigi754 I merged https://github.com/islet-project/islet/pull/367 and can't reporduce your issue. Please update your local repository and try again.

Havner:
[...]

PR #393, opened 2024-11-05
Don't start terminal others than 5000 and 5003 with FVP

Havner:
Possible fix for #392.


zpzigi754:
We actually control the same variables (bp.terminal_1.start_telnet and bp.terminal_2.start_telnet) in fvp-cca's `--no-telnet` option ([link](https://github.com/islet-project/islet/blob/main/scripts/fvp-cca#L429-L430)). Thus, I think that controlling with `--no-telnet` might be enough without enforcing it for all cases.

Havner:
> We actually control the same variables (bp.terminal_1.start_telnet and bp.terminal_2.start_telnet) in fvp-cca's `--no-telnet` option ([link](https://github.com/islet-project/islet/blob/main/scripts/fvp-cca#L429-L430)). Thus, I think that controlling with `--no-telnet` might be enough without enforcing it for all cases.

Ports 5001 and 5002 are not used for any human readable communication. IMO there is no need for FVP to ever start telnets automatically.

Even though the `--no-telnet` option turns them off as well it turns all of them off. So the bug still remains if someone without that knowledge simply runs FVP with HES. The same bug you encountered.

jinbpark (approved):

zpzigi754:
I think that this PR can be closed without merge as the symptom has disappeared. Also, note that port 5002 (uart2) is actually used for human readable communication in ACS-tests.

PR #394, opened 2024-11-06
Refactor event handling logic & Handle the data abort case

bitboom:
This PR
 - refactors event handling logic
 - removes temporary mapping for handling data abort (https://github.com/islet-project/islet/pull/394/commits/011e2fae1f740601579eb8c40d40d1a53bd2821d)

_NOTE: I will submit the PR for managing the Page Table as a pinned object separately, as it is experimental._

# Refactor event handling logic

## Problem
- RMM handles two types of events: RSI and RMI,
  but RMI handler was located in the mainloop while RSI handler was part of RMM.
- The mainloop, created and managed by RMM, had a cyclic dependency
  as the RMI handler in the mainloop referenced the monitor.
- Handling RMI in the mainloop also required locking the event queue,
  leading to performance issues.

## Patch
- Centralized both RMI and RSI event handlers within RMM for consistent event management.
- Refactored mainloop to only poll for RMI events, following SRP principle.
- Removed mutual references between mainloop and monitor, breaking the cyclic dependency.
- Eliminated the need for queue locking by shifting event handling to RMM, improving performance.

### DIFF
--- a/rmm/src/event/mainloop.rs
+++ b/rmm/src/event/mainloop.rs
@@ -1,165 +1,27 @@
-extern crate alloc;
-
 use super::Context;
 use crate::asm::smc;
 use crate::rmi;
-use crate::rmi::error::Error;
-use crate::Monitor;
 
-use alloc::boxed::Box;
-use alloc::collections::btree_map::BTreeMap;
-use alloc::collections::vec_deque::VecDeque;
-use spin::mutex::Mutex;
-
-pub type Handler = Box<dyn Fn(&[usize], &mut [usize], &Monitor) -> Result<(), Error>>;
-
-pub struct Mainloop {
-    pub queue: Mutex<VecDeque<Context>>, // TODO: we need a more realistic queue considering multi-core environments if needed
-    pub on_event: BTreeMap<usize, Handler>,
-}
+pub struct Mainloop;
 
 impl Mainloop {
     pub fn new() -> Self {
-        Self {
-            queue: Mutex::new(VecDeque::new()),
-            on_event: BTreeMap::new(),
-        }
+        Self
     }
 
     #[cfg(not(kani))]
-    pub fn add_event_handlers(&mut self) {
-        rmi::features::set_event_handler(self);
-        rmi::gpt::set_event_handler(self);
-        rmi::realm::set_event_handler(self);
-        rmi::rec::set_event_handler(self);
-        rmi::rtt::set_event_handler(self);
-        rmi::version::set_event_handler(self);
-    }
-    #[cfg(kani)]
-    fn add_event_handlers(&mut self) {
-        #[cfg(feature = "mc_rmi_features")]
-        rmi::features::set_event_handler(self);
-        #[cfg(any(
-            feature = "mc_rmi_granule_delegate",
-            feature = "mc_rmi_granule_undelegate"
-        ))]
-        rmi::gpt::set_event_handler(self);
-        #[cfg(any(
-            feature = "mc_rmi_realm_activate",
-            feature = "mc_rmi_realm_destroy",
-            feature = "mc_rmi_rec_aux_count"
-        ))]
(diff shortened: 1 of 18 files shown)


rmm/src/rmi/rec/handlers.rs
```diff
@@ -52,7 +52,9 @@ pub fn set_event_handler(mainloop: &mut Mainloop) {
             return Err(Error::RmiErrorInput);
         }
 
+        rmm.page_table.map(params_ptr, false);
```
jinbpark, line 55: Do we need to call `rmm.page_table.map()` here?

host::copy_from() seems already doing this job.
```
PageTable::get_ref().map(addr, false);
let ret = assume_safe::<T>(addr).map(|safety_assumed| *safety_assumed);
PageTable::get_ref().unmap(addr);
```

bitboom: The mapping in `host::copy_from()` is removed at https://github.com/islet-project/islet/pull/394/commits/64d265e8af0905c198db69a2f52c8e372f2b2e15.
The purpose is to control the page table exclusively through the `Monitor` instance, except for global access.

jinbpark: Oh, I miss that.

Then, should we eliminate some corner cases (see below)?
```
rmm.page_table.map(params_ptr, false);
let params = host::copy_from::<Params>(params_ptr).ok_or(Error::RmiErrorInput)?;  
// If this check of copy_from() fails, unmap() will not be called.
rmm.page_table.unmap(params_ptr);
```

Or, you should just ignore this case this time. I've already got an issue (https://github.com/islet-project/islet/issues/375) dedicated to this kind of issue. (unmap() being not invokved on specific code paths)
We may better handle all such cases in one PR.

bitboom: I believe the error handling you mentioned is a critical aspect, as it is being used in many places. One of the approaches I am currently considering is as follows:

1. Acquire the Granule Lock.
2. Map the given address in `get_granule_*`.
3. When the Granule Lock is released, unmap it based on RAII.

This approach would allow us to eliminate the seperate locks used for the RMM PageTable and GST, as well as simplify the error-handling code.

rmm/src/event/mainloop.rs
```diff
@@ -1,165 +1,28 @@
...
-                ctx.init_arg(&ret[1..arg_num]);
-                ctx.resize_ret(ret_num);
-                self.queue.lock().push_back(ctx);
-            },
-            || {
-                let ctx = Context::new(rmi::NOT_SUPPORTED_YET);
-                self.queue.lock().push_back(ctx);
-            },
-        );
+        // need to check index 1?
```
hihi-wang, line 16: [minor]Do we need this comment?

hihi-wang (approved):

zpzigi754 (approved):
Good refactoring!

rmm/src/test_utils.rs
```diff
@@ -11,16 +11,14 @@ pub use mock::host::{alloc_granule, granule_addr};
 use alloc::vec::Vec;
 
 pub fn rmi<const COMMAND: usize>(arg: &[usize]) -> Vec<usize> {
-    let mut mainloop = Mainloop::new();
-    let monitor = Monitor::new();
-    (&mut mainloop).add_event_handlers();
+    let rmm = Monitor::new();
```
zpzigi754, line 14: [minor]
How about using the previous variable name: `monitor`?

nook1208 (approved):
nice refactoring !

rmm/src/event/mainloop.rs
bitboom: I removed it.

rmm/src/test_utils.rs
bitboom: I renamed it.

PR #396, opened 2024-11-06
Fix veraison example

L0czek:
This PR will fix issues with the latest veraison version.

Requires:
* https://github.com/islet-project/remote-attestation/pull/14

PR #397, opened 2024-11-14
Errata for the veraison PR #396

Havner:
Don't shallow check out the veraison repo. We need a specific frozen revision and it's impossible using depth=1.


zos (approved):

zos:
Add signoff

PR #400, opened 2024-11-15
Add Veraison to app-provisioning example

Havner:
Add full support for Veraison service and its verification for the app-provisioning example.

Various fixes and cleanups in the readme.

Also small 2 commits that cleanup hes and examples/veraison.


jinbpark (approved):

Havner:
This instruction should be followed with `realm-manager` from this branch (until it's merged):

https://github.com/islet-project/realm-manager/tree/dev/m.szaknis/fix_veraison

zpzigi754 (approved):
How about having a soft link in `islet/doc` for `examples/app-provisioning/README.md`? Then, it can be seen in https://islet-project.github.io/islet/.

p-sawicki2:
I believe that keeping the detailed instruction in the examples folder is the best option. `islet/doc` is rather a good place for general description and design documents. If it is needed, we could create some distilled version of our design documentation of the application provisioning mechanism and place it here.

examples/app-provisioning/README.md
```diff
@@ -7,69 +7,113 @@ The process of provisioning is handled by a realm daemon which is responsible fo
...
+    cp out/Image $CCA/islet/out/shared/nasz-realm
+    cp out/initramfs.cpio.gz $CCA/islet/out/shared/initramfs.cpio.gz
 
 ### Build warden daemon
 
 Warden daemon is responsible for providing resources such as disks and networking to realms. It manages realm lifetime by starting and stopping kvmtool. More details are available at [realm-manager/warden/warden_daemon](https://github.com/islet-project/realm-manager/tree/main/warden/warden_daemon).
 
-    cd $ROOT/realm-manager/warden
+    cd $CCA/realm-manager/warden
     make
```
p-sawicki2, line 56: Mention that a nightly version of the Rust toolchain has to be installed to build Warden.

examples/app-provisioning/README.md
```diff
@@ -7,69 +7,113 @@ The process of provisioning is handled by a realm daemon which is responsible fo
...
+    git clone https://github.com/islet-project/realm-manager.git
+    git clone https://github.com/islet-project/realm-metadata-tool.git
+    git clone https://github.com/islet-project/image-registry.git
 
-### Initialize islet
+## Prepare/initialize islet
 
-    cd $ROOT/islet
+    cd $CCA/islet
     ./scripts/init.sh
```
p-sawicki2, line 36: Add a comment, that if there is already rustup installed on a system, it is advisable to run `rustup update` before running the build process.

Havner: Islet has a forced rust version in `rust-toolchain`. Rustup update is not needed here. Besides this document is not an Islet manual.

p-sawicki2: I encountered issues related to outdated toolchain while performing this instruction on my laptop. Rust build tools were not so verbose about the reason of the failure. So, there should be somewhere a note that if one has already installed rustup he has to ensure that everything is up-to-date. At first glance the initialization of Islet is the best place, as the init.sh installs the rustup tool.

Havner: If the init.sh installs the rustup tool than more the reason not to put anything here. You will have a newest version after this command. If you did init.sh some time ago it's outside the scope of this manual. And if anything the better place would be to do a rustup update in that script if you have rustup already installed. This tutorial cannot cover every single use case related to maintaining your linux distribution. It's your job, not the scope of this manual.

PR #403, opened 2024-11-25
Deprecate cli related components

bitboom:
This PR deprecates `cli` and use [rsictl](https://github.com/islet-project/rsictl) as is. _(Fixed https://github.com/islet-project/islet/issues/382.)_

- Add `tools/rsictl` as a submodule
- Remove diverged libs (cose, cca-token, rsi-el0)
- Use `rust-rsi` at `sdk`


bitboom:
@p-sawicki2 

I am currently resolving https://github.com/islet-project/islet/issues/382 in this PR. As the final step, I plan to remove `lib/cca_token`, which is the `token module` in [rust-rsi](https://github.com/islet-project/rust-rsi).

One of the requirements from the [certifier](https://github.com/ccc-certifier-framework/certifier-framework-for-confidential-computing/blob/main/src/islet/islet_shim.cc#L107) is to provide a C API that allows access to the RIM value and the Realm’s Challenge from the attestation report.
```cpp
// certifier
rv = islet_parse(CLAIM_TITLE_RIM,
                 claims,
                 claims_len,
                 measurement_out,
                 measurement_out_size);
return rv == ISLET_SUCCESS;
```

 I have been using `ClaimData` for parsing this information in [SDK](https://github.com/islet-project/islet/blob/main/sdk/src/c_api.rs#L98). However, since the `token module` in rust-rsi is private, it cannot be directly accessed from the SDK.

```rust
// islet-sdk
#[no_mangle]
pub unsafe extern "C" fn islet_parse(
    title: *const c_char,
    claims: *const c_uchar,
    claims_len: c_int,
    value_out: *mut c_uchar,
    value_out_len: *mut c_int,
) -> islet_status_t {
    let do_parse = || -> Result<(), Error> {
...
        match parse(&claims, title) {
            Some(ClaimData::Bstr(value)) => {
...
            }
```

To address this, I believe there are two possible approaches:

1.	Provide an API in `rust-rsi` that allows parsing by passing the name of a specific claim.
2.	Make the token module a public module so that the SDK can access it directly and perform the parsing.

Could you share which approach do you prefer?

p-sawicki2:
@bitboom I would prefer the 2nd option, bacause `rust-rsi` already contains code for parsing ([verifier.rs](https://github.com/islet-project/rust-rsi/blob/main/src/token/verifier.rs)) the attestation token and getting the claims (example usage of `get_claim()` function: [parser.rs](https://github.com/islet-project/rust-rsi/blob/main/src/token/parser.rs)). So, it seems that the only thing is just making them public.
@Havner What is your opinion on that?

Havner:
@bitboom I'm not sure I understand. You need to extract the RIM and the challenge from the token? Because if that is the case you don't need to make any changes to the `rust-rsi`. You do `verify_token` which is re-exported in the `lib.rs` and you have a struct of claims `AttestationClaims` (platform and realm tokens) from which you can extract any information you need.

bokdeuk-jeong (approved):

.gitmodules
```diff
@@ -73,3 +73,6 @@
 [submodule "third-party/dtc"]
 	path = third-party/dtc
 	url = https://git.kernel.org/pub/scm/utils/dtc/dtc.git
+[submodule "tools/rsictl"]
```
Havner, line 76: You don't need to create a submodule, you can do cargo install directly from git. It limits number of entities we need to handle and care about.

You can look here, I did something similar:

https://github.com/islet-project/islet/blob/main/examples/veraison/Makefile#L20

The only thing to care about is that you need to pass linker flags manually as `.cargo/config.toml` is not used from git.

You can also freeze the revision in the `cargo install` command (`--rev`) so it doesn't break by itself and needs to be updated manually as it would in case of a submodule.

Cargo.toml
```diff
@@ -4,15 +4,16 @@ members = [
...
     "lib/uart",
-    "lib/io",
+    "lib/vmsa",
+    "model-checking",
     "plat/fvp",
     "realm/rsi-test",
     "rmm/",
     "sdk",
-    "model-checking",
+    "tools/rsictl",
```
Havner, line 14: This dir from what I see was created specifically for `rsictl`. Installing from git would prevent that. I'd like to cleanup such things, not create more of them.

jinbpark (approved):

.gitmodules
bitboom: It's better to install directly as you said. I reflected it.

Cargo.toml
bitboom: I removed it.

PR #405, opened 2024-12-02
Add a guard page to mitigate stack overflow

bitboom:
This PR makes the first page of stack as the guard page (unmapped) to mitigate stack overflow.

## AS-IS
The RMM memory layout assigns the CPU stacks in a continuous manner, which can lead to stack overflow issues spilling over between CPUs. 

```
 | core 7 stack | core 6 stack | ... | core 0 stack |
 ^                 
stack_start 
(low addr)      >>>      (high addr)
```

## TO-BE
To mitigate this, the first page of each stack is reserved as a guard page.
```
 | guard_page | core 0 stack | guard_page | core 1 stack | ... | core 7 stack |
              ^                    
             stack_start 
(low addr)      >>>      (high addr)
```

### DIFF
--- a/plat/fvp/src/entry.rs
+++ b/plat/fvp/src/entry.rs
@@ -4,25 +4,35 @@ use crate::log::LevelFilter;
 use aarch64_cpu::registers::*;
 use core::ptr::{addr_of, addr_of_mut};
 use io::stdout;
-use islet_rmm::config::{NUM_OF_CPU, RMM_STACK_SIZE};
+use islet_rmm::config::{NUM_OF_CPU, RMM_STACK_GUARD_SIZE, RMM_STACK_SIZE};
 use islet_rmm::logger;
 
+/// Configure the first page of the stack section as a stack guard page
 #[no_mangle]
 #[link_section = ".stack"]
-static mut RMM_STACK: [u8; RMM_STACK_SIZE * NUM_OF_CPU] = [0; RMM_STACK_SIZE * NUM_OF_CPU];
+static mut RMM_STACK: [[u8; RMM_STACK_SIZE + RMM_STACK_GUARD_SIZE]; NUM_OF_CPU] =
+    [[0; RMM_STACK_SIZE + RMM_STACK_GUARD_SIZE]; NUM_OF_CPU];
+
+/// # Safety
+///
+/// This function only reads the address of the stack.
+#[no_mangle]
+pub unsafe extern "C" fn current_cpu_stack() -> usize {
+    let cpu_id = get_cpu_id();
+    if cpu_id >= NUM_OF_CPU {
+        panic!("Invalid CPU ID!");
+    }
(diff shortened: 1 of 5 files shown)


jinbpark:
Can we have a test case that attempts stack overflow to test if it works correctly?
It seems not trivial as it involves real MMU mappings.

Generally speaking, we might need system tests that need to be run on FVP but are not suitable to be in ACS or TF-A tests.
(since this attack tests require adding a sample vulnerability in islet rmm)

It seems that, as of now, this stack overflow case is the only case that demands such a system test environment.
If we face a more number of cases, it might be worth introducing such an environment, but not sure at this point if it's worthwhile.

bitboom:
> Can we have a test case that attempts stack overflow to test if it works correctly? It seems not trivial as it involves real MMU mappings.
> 
> Generally speaking, we might need system tests that need to be run on FVP but are not suitable to be in ACS or TF-A tests. (since this attack tests require adding a sample vulnerability in islet rmm)
> 
> It seems that, as of now, this stack overflow case is the only case that demands such a system test environment. If we face a more number of cases, it might be worth introducing such an environment, but not sure at this point if it's worthwhile.

The way I tested is that RMM allocates a local variable with STACK_SIZE. Without this PR the below case is succeed at ACS test.

```rust
let overflow: [u8; RMM_STACK_SIZE] = [0; RMM_STACK_SIZE];
error!("Trigger overflow {:?}", overflow);
```

I agree with your suggest. I will consider more testing methods to check if stack overflow occurred by adding test cases.

plat/fvp/src/entry.rs
```diff
@@ -9,33 +9,41 @@ use islet_rmm::logger;
 
 #[no_mangle]
 #[link_section = ".stack"]
-static mut RMM_STACK: [u8; RMM_STACK_SIZE * NUM_OF_CPU] = [0; RMM_STACK_SIZE * NUM_OF_CPU];
+static mut RMM_STACK: [[u8; RMM_STACK_SIZE]; NUM_OF_CPU] = [[0; RMM_STACK_SIZE]; NUM_OF_CPU];
```
hihi-wang, line 12: It seems like `RMM_STACK_SIZE` is equal to `actual rmm stack size + stack page`. 

To prevent misconception, how about modfying this and following codes to `RMM_STACK_SIZE + GRANULE_SIZE`, so that `RMM_STACK_SIZE` is equal to be actual stack size for each RMM.

nook1208, line 12: If we use guard pages for the stack, I think that it's better to leave a few comments about it here or somewhere else

bitboom: I applied it! Thanks :)

bitboom: I noted like `/// Configure the first page of the stack section as a stack guard page`. Thanks!

jinbpark (approved):

nook1208 (approved):

hihi-wang (approved):

PR #407, opened 2024-12-10
Update gitmodules for the islet-assets repo split

Havner:
I've created 4 repos for this:

`3rd-android-kernel`
`3rd-gki-build`
`3rd-linux` (for `nw-linux` and `realm-linux`)
`3rd-kvmtool` (for `kvmtool` and `kvmtool-rim-measurer`)

I didn't know the original repositories for the first two so they are just a copy&paste of the assets branch into main. More branches can be added in the future if required. For the latter two I've pushed some branches from the CCA upstream that we based our changes on.

The changed revision for `kvmtool-rim-measurer` is just a rebase to make the branches more consistent. It doesn't change any code.

I have not done `optee` as a dependency on this repository should be removed altogether. This will come later.

I've also marked most of the branches in the `assets` repository not to use them with an appropriate message where the changes should be made from now on.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,16 +3,16 @@
 	url = https://github.com/islet-project/assets
 [submodule "third-party/nw-linux"]
 	path = third-party/nw-linux
-	url = https://github.com/islet-project/assets
-	branch = eac5/3rd-nw-linux-realm-metadata
+	url = https://github.com/islet-project/3rd-linux
+	branch = nw-linux/eac5
 [submodule "third-party/optee-build"]
 	path = third-party/optee-build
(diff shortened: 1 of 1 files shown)


bitboom (approved):

nook1208 (approved):

p-sawicki2 (changes requested):
Shouldn't the changes in the .gitmodules be reflected also in the third-party folder?

Havner:
> Shouldn't the changes in the .gitmodules be reflected also in the third-party folder?

If the revision didn't change as it's the case here then no. I even commented on the one revision that did change.

p-sawicki2 (approved):

PR #408, opened 2024-12-11
Realm FPU context

bokdeuk-jeong:
1. exception: move non-volatile to volatile
    - Move restore_nonvolatile_registers_from_rec_and_run  to restore_volatile_registers_from_rec_and_run.
    - Make the func name short: from xxx_from_rec_and_run to xxx_from_rec.

2. rmi_features: use defined_bits() for feature register
    - rmm: fpu contenxt management for realms
    
3. rmm: fpu contenxt management for realms
   - save the fpu context on rec's exit if fpu is used during its runtime.
   - resotre the fpu context on rec's first access to fpu since its recent REC_ENTER.

### DIFF
--- a/rmm/src/exception/vectors.s
+++ b/rmm/src/exception/vectors.s
@@ -52,8 +52,8 @@
 	stp x1, x2, [x18, #REC_GP_REGS + 8 * 31]
 .endm
 
-.global restore_all_from_rec_and_run
-restore_all_from_rec_and_run:
+.global restore_all_from_rec
+restore_all_from_rec:
 	mrs x0, tpidr_el2
 
 	/* Restore system registers */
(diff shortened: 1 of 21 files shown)


rmm/src/exception/vectors.s
```diff
@@ -200,7 +195,7 @@ restore_volatile_from_stack_and_return:
 	cbnz x0, rmm_enter
 
 	mrs x0, tpidr_el2
-	b restore_nonvolatile_from_rec_and_run
+	b restore_volatile_from_rec
```
nook1208, line 198: Isn't it enough to restore only the callee-saved registers (X19 ... X28) ?

bokdeuk-jeong: No, it's not enough. Callees are not voluntarily taken to EL2 and we had previously experienced problem with that.

nook1208 (approved):
Good to support SME :)

jinbpark (approved):

hihi-wang (approved):

zpzigi754:
Really nice to find a subtle bug and fix it!

PR #413, opened 2024-12-27
script: Add scripts to run AOSP on QEMU

nook1208:
Use the following command to build & launch android on qemu: 
./script/init_android_on_qemu.sh
./scripts/qemu-cca.py -aosp

A number of patches for CCA support are applied to the host kernel. At the moment, it supports rmm-v1.0-eac5. But it'll be updated to the latest. 
It's the below branch in [islet-project/3rd-android-kernel:](https://github.com/islet-project/3rd-android-kernel/tree/common-android15-6.6/cca-host/rmm-v1.0-eac5)
common-android15-6.6/cca-host/rmm-v1.0-eac5

If the qemu is launched, you can see information about it like this: 
"The following lines contain useful debugging information:
  Serial console is disabled; use -console=true to enable it.
  Logcat output: /home/eom/cuttlefish/instances/cvd-1/logs/logcat
  Kernel log: /home/eom/cuttlefish/instances/cvd-1/kernel.log
  Launcher log: /home/eom/cuttlefish/instances/cvd-1/logs/launcher.log
  Instance configuration: /home/eom/cuttlefish/instances/cvd-1/cuttlefish_config.json
  Launcher Build ID: eng.eom"

After the virtual device booted, you can see the following: 
VIRTUAL_DEVICE_BOOT_COMPLETED
Virtual device booted successfully

Then, check your device using adb:
$ adb devices
List of devices attached
0.0.0.0:6520	device

$ adb shell
vsoc_arm64_only:/ $ uname -a
Linux localhost 6.6.57-android15-8-maybe-dirty-4k #1 SMP PREEMPT Thu Jan  1 00:00:00 UTC 1970 aarch64 Toybox

### DIFF
--- a/scripts/config.py
+++ b/scripts/config.py
@@ -73,3 +73,8 @@ FVP_BIN = os.path.join(FVP_DIR, "./FVP_Base_RevC-2xAEMvA")
 PLUGIN_PATH = os.path.join(THIRD_PARTY, "fvp/Base_RevC_AEMvA_pkg/plugins/Linux64_GCC-9.3/")
 TRACE_LIB = os.path.join(PLUGIN_PATH, "TarmacTrace.so")
 TOGGLE_LIB = os.path.join(PLUGIN_PATH, "ToggleMTIPlugin.so")
+
+AOSP_DIR = os.path.join(ROOT, "android_on_qemu/aosp-15.0.0_r8")
+AOSP_KERNEL_OUT = os.path.join(ROOT, "android_on_qemu/android15-6.6/out/virtual_device_aarch64/dist/")
+AOSP_KERNEL_INITRAMFS_PATH = os.path.join(AOSP_KERNEL_OUT, "initramfs.img")
+AOSP_KERNEL_IMAGE_PATH = os.path.join(AOSP_KERNEL_OUT, "Image")
--- a/scripts/config.py
+++ b/scripts/config.py
@@ -74,7 +74,7 @@ PLUGIN_PATH = os.path.join(THIRD_PARTY, "fvp/Base_RevC_AEMvA_pkg/plugins/Linux64
 TRACE_LIB = os.path.join(PLUGIN_PATH, "TarmacTrace.so")
 TOGGLE_LIB = os.path.join(PLUGIN_PATH, "ToggleMTIPlugin.so")
 
-AOSP_DIR = os.path.join(ROOT, "android_on_qemu/aosp-15.0.0_r8")
-AOSP_KERNEL_OUT = os.path.join(ROOT, "android_on_qemu/android15-6.6/out/virtual_device_aarch64/dist/")
+AOSP_DIR = os.path.join(THIRD_PARTY, "android_on_qemu/aosp-15.0.0_r8")
+AOSP_KERNEL_OUT = os.path.join(THIRD_PARTY, "android_on_qemu/android15-6.6/out/virtual_device_aarch64/dist/")
 AOSP_KERNEL_INITRAMFS_PATH = os.path.join(AOSP_KERNEL_OUT, "initramfs.img")
 AOSP_KERNEL_IMAGE_PATH = os.path.join(AOSP_KERNEL_OUT, "Image")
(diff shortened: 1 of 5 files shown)


scripts/init_android_on_qemu.sh
```diff
@@ -0,0 +1,123 @@
...
+		echo "Creating directory $ANDROID_KERNEL_DIR..."
+		mkdir -p "$ANDROID_KERNEL_DIR"
+	fi
+
+	echo "Changing directory to $ANDROID_KERNEL_DIR..."
+	cd $ANDROID_KERNEL_DIR || exit 1
+
+	if [ ! -d "common" ]; then
+		echo "Downloading Android Kernel sources..."
+		repo init --partial-clone -u https://android.googlesource.com/kernel/manifest -b common-$ANDROID_KERNEL_VER
```
bokdeuk-jeong, line 89: Do we need to download the kernel source code from here? we will use our own at line 96.

scripts/init_android_on_qemu.sh
```diff
@@ -0,0 +1,123 @@
...
+		git clone $ANDROID_KERNEL_URL -b $ANDROID_KERNEL_BRANCH --depth 1 --single-branch common
+	fi
+
+	if [ -f "$INITRAMFS_PATH" ] && [ -f "$KERNEL_PATH" ]; then
+		echo "Build is alread done. Skip building Android Kernel"
+		return
+	fi
+
+	# Build
+	echo "Building Android Kernel..."
```
bokdeuk-jeong, line 105: We can build Android kernel apart from AOSP build and provide its built image path as you do in qemu-cca.py. Do we still need this too?

scripts/init_android_on_qemu.sh
nook1208: Oh, you're right. I'll remove it. Thank you :)

scripts/init_android_on_qemu.sh
nook1208: This build only runs if there are no built images (at line 99). 
So we need this code.

scripts/config.py
```diff
@@ -73,3 +73,8 @@
 PLUGIN_PATH = os.path.join(THIRD_PARTY, "fvp/Base_RevC_AEMvA_pkg/plugins/Linux64_GCC-9.3/")
 TRACE_LIB = os.path.join(PLUGIN_PATH, "TarmacTrace.so")
 TOGGLE_LIB = os.path.join(PLUGIN_PATH, "ToggleMTIPlugin.so")
+
+AOSP_DIR = os.path.join(ROOT, "android_on_qemu/aosp-15.0.0_r8")
```
bokdeuk-jeong, line 77: As AOSP is third party code, I think  placing it under THIRD_PARTY looks more consistent with existing modules.

scripts/init_android_on_qemu.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
```
bokdeuk-jeong, line 1: Existing script/init.sh downloads packages and source code. To be consistent with existing ones, I suggest we only download aosp and kernel code here and move the build activities to someother script.
In addition, I think package installation for cuttlefish tools and installaing after downloading the project for the cuttlefish tools is missing here in this init script.

nook1208: > I think package installation for cuttlefish tools and installaing after downloading the project for the cuttlefish tools is missing here in this init script.

We don't need other cuttlefish tool. Because if we build aosp, then cuttlefish tool is also built. we can just use it:
$ which launch_cvd
/home/eom/git/islet/android_on_qemu/aosp-15.0.0_r8/out/host/linux-x86/bin/launch_cvd

bokdeuk-jeong: I was also wondering why we need to install cuttlefish tools while we use them built by our own under aosp. Thanks for the clarification.

scripts/config.py
nook1208: Applied. Thanks !

bokdeuk-jeong (approved):

PR #415, opened 2024-12-31
fix: resolve pip install toml error with virtual python env

drvcodenta:
Fixed the issue where running `./scripts/init.sh` fails on Ubuntu 24 due to the `externally-managed-environment` error, which arises from Ubuntu 24.04 systems preventing conflicts between system and user-managed Python packages.
Implemented a solution by creating a temporary Python virtual environment for the toml installation.
Updated .gitignore to exclude the venv folder to avoid tracking the virtual environment files.
- Raised the issue for review, as using a virtual environment for a single dependency may need further discussion.
- Open to suggestions for alternative solutions
Here is screenshot of error log
Screenshot 2024-12-31 163556

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ examples/cross-platform-e2ee/third-party-arm/
 
 unsafe-call-trace.log
 unsafe-items-list.log
+/venv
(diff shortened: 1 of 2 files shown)


bokdeuk-jeong (approved):
Thanks for the proposal.

nook1208 (approved):
It works on my Ubuntu 22.04 too. Thanks !

PR #422, opened 2025-02-23
rmm: Add more RMI fuzzing harnesses

Skryptonyte:
(LFX Task 2)

This PR adds the following harnesses:

1. rmi_granule_delegate_fuzz: Fuzzes GRANULE_DELEGATE and GRANULE_UNDELEGATE
2. rmi_granule_undelegate_fuzz: Fuzzes invalid GRANULE_UNDELEGATE calls.
3. rmi_realm_create_fuzz: Fuzzes REALM_CREATE, REALM_ACTIVATE, and REALM_DESTROY (RealmParams)
4. rmi_rec_create_fuzz: Fuzzes REC_CREATE and REC_DESTROY (RecParams)
5. rmi_rtt_create_fuzz: Fuzzes RTT_CREATE and RTT_DESTROY
6. rmi_rtt_read_entry_fuzz: Fuzzes RTT_READ_ENTRY
7. rmi_data_create_fuzz: Fuzzes DATA_CREATE and DATA_DESTROY
8. rmi_data_create_unknown_fuzz: Fuzzes DATA_CREATE_UNKNOWN and DATA_DESTROY
9. rmi_rtt_map_unprotected_fuzz: Fuzzes RTT_MAP_UNPROTECTED and RTT_UNMAP_UNPROTECTED
10. rmi_rtt_init_ripas_fuzz: Fuzzes RTT_INIT_RIPAS
11. rmi_rtt_set_ripas_fuzz: Fuzzes RTT_SET_RIPAS 
12. rmi_rec_enter_fuzz: Fuzzes REC_ENTER (Run entry paramaters)
13. rmi_psci_complete_fuzz: Fuzzes PSCI_COMPLETE
14. rmi_rtt_fold_fuzz: Fuzzes RTT_FOLD

I am still working on improving coverage wherever possible and code cleanups, so I am opting to keep it a draft for some time.

### DIFF
--- a/rmm/fuzz/Cargo.toml
+++ b/rmm/fuzz/Cargo.toml
@@ -24,3 +24,101 @@ path = "fuzz_targets/rmi_features_fuzz.rs"
 test = false
 doc = false
 bench = false
+
+[[bin]]
+name = "rmi_granule_delegate_fuzz"
+path = "fuzz_targets/rmi_granule_delegate_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_granule_undelegate_fuzz"
+path = "fuzz_targets/rmi_granule_undelegate_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_realm_create_fuzz"
+path = "fuzz_targets/rmi_realm_create_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_rec_create_fuzz"
+path = "fuzz_targets/rmi_rec_create_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_rtt_create_fuzz"
+path = "fuzz_targets/rmi_rtt_create_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_rtt_read_entry_fuzz"
+path = "fuzz_targets/rmi_rtt_read_entry_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_data_create_fuzz"
+path = "fuzz_targets/rmi_data_create_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_data_create_unknown_fuzz"
+path = "fuzz_targets/rmi_data_create_unknown_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
+name = "rmi_rtt_map_unprotected_fuzz"
+path = "fuzz_targets/rmi_rtt_map_unprotected_fuzz.rs"
+test = false
+doc = false
+bench = false
+
+[[bin]]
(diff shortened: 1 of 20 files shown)


Skryptonyte:
I updated rmi_rtt_fold_fuzz to try to fuzz the following cases where all L3 entries in an L2 entry are:

1. UNASSIGNED_RAM
2. ASSIGNED_RAM
3. UNASSIGNED_NS
4. ASSIGNED_NS

Cases 1 and 3 are getting reached correctly, but cases 2 and 4. seem to fail at this particular snippet.

https://github.com/islet-project/islet/blob/main/rmm/src/realm/mm/stage2_tte.rs#L250-L254

Another thing is all entries seem to start of as UNASSIGNED regardless of if it's in the protected or unprotected space. Is that intended or should the entries mapping unprotected IPAs start off as UNASSIGNED_NS instead? As a workaround, I have to use RTT_MAP_UNPROTECTED and then RTT_UNMAP_UNPROTECTED to create UNASSIGNED_NS entries.

rmi_rtt_fold_fuzz is also quite slow compared to the other fuzzers because of the sheer number of RMI commands it has to execute. This is currently the most complex fuzzer in this PR.

bokdeuk-jeong:
https://github.com/islet-project/islet/blob/main/rmm/src/realm/mm/stage2_tte.rs#L250-L254

It checks if blocks or pages pointed by the last level page table are consecutive and aligned at its parent's page table level.
For example, let's say we have consecutive 512 4KB pages, each of which is pointed by an entry in an L3 page table, say pgt_l3. 
However, using folding, it can also be pointed by an L2 page table, which is the parent of pgt_l3, as a single 2MB block if the 2MB block consisting of the 512 4KB pages is alinged with 2MB. So, the line checks if the address of the first 4KB page pointed by the first entry in pgt_l3 is alinged with its upper level (2MB). Otherwise, it can't be folded.

@zpzigi754, could you answer with the state of memory in a realm? I am bit rusty with the spec.

Skryptonyte:
> https://github.com/islet-project/islet/blob/main/rmm/src/realm/mm/stage2_tte.rs#L250-L254
> 
> It checks if blocks or pages pointed by the last level page table are consecutive and aligned at its parent's page table level. For example, let's say we have consecutive 512 4KB pages, each of which is pointed by an entry in an L3 page table, say pgt_l3. However, using folding, it can also be pointed by an L2 page table, which is the parent of pgt_l3, as a single 2MB block if the 2MB block consisting of the 512 4KB pages is alinged with 2MB. So, the line checks if the address of the first 4KB page pointed by the first entry in pgt_l3 is alinged with its upper level (2MB). Otherwise, it can't be folded.
> 
> @zpzigi754, could you answer with the state of memory in a realm? I am bit rusty with the spec.

Thanks for the explanation! I was able to get the RTT_FOLD fuzzing working on all cases now. rmi_rtt_fold_fuzz coverage has improved significantly now.

bokdeuk-jeong (approved):
It's greate to see the fuzz harness in such a short time. I appreciate the time and effort you put into writing the test code with carefully crafted inputs based on a good understand of the spec.

bitboom (approved):

hihi-wang (approved):

PR #423, opened 2025-03-02
rmi/realm: Add check to bound rtt_level_start below RTT_PAGE_LEVEL

Skryptonyte:
Passing RealmParams to REALM_CREATE with rtt_level_start higher than RTT_PAGE_LEVEL causes a subtract overflow. This PR adds a check to ensure it does not exceed RTT_PAGE_LEVEL.

Resolves https://github.com/islet-project/islet/issues/420


rmm/src/rmi/realm/params.rs
```diff
@@ -129,6 +129,10 @@ impl Params {
         let ipa_bits = self.ipa_bits();
         let rtt_slvl = self.rtt_level_start as usize;
 
+        if rtt_slvl > RTT_PAGE_LEVEL {
+            return Err(Error::RmiErrorInput);
+        }
+
         let level = RTT_PAGE_LEVEL - rtt_slvl;
```
bitboom, line 136: Nice work! Just a minor suggestion: How about using `checked_sub` to make it more concise?

```suggestion
let level = RTT_PAGE_LEVEL.checked_sub(rtt_slvl).ok_or(Error::RmiErrorInput)?;
```

bokdeuk-jeong (approved):

hihi-wang (approved):

bitboom (approved):

nook1208 (approved):
Good finding!

PR #425, opened 2025-03-26
realm/mm: Correct setting of RIPAS=EMPTY on ASSIGNED_EMPTY RTTEs

Skryptonyte:
Currently, if RTT_SET_RIPAS were to be run on ASSIGNED_EMPTY RTTEs with the requested RIPAS being EMPTY, the HIPAS would unexpectedly change to UNASSIGNED. The expected behaviour is that the entry states should remain unchanged.

To fix this, the PR does the following:
1. Change the first case under the EMPTY request to handle UNASSIGNED_RAM instead of UNASSIGNED_EMPTY.
2. Add a fallback case which will handle both UNASSIGNED_EMPTY and ASSIGNED_EMPTY i.e the entries will be left unchanged.

This will hence mirror the code where the requested RIPAS is RAM instead.

### DIFF
--- a/rmm/src/realm/mm/rtt.rs
+++ b/rmm/src/realm/mm/rtt.rs
@@ -384,7 +384,7 @@ pub fn set_ripas(rd: &Rd, base: usize, top: usize, ripas: u8, flags: u64) -> Res
         if ripas as u64 == invalid_ripas::EMPTY {
             new_s2tte |= bits_in_reg(S2TTE::INVALID_RIPAS, invalid_ripas::EMPTY);
 
-            if s2tte.is_unassigned_empty() {
+            if s2tte.is_unassigned_ram() {
                 new_s2tte |= bits_in_reg(S2TTE::INVALID_HIPAS, invalid_hipas::UNASSIGNED);
             } else if s2tte.is_unassigned_destroyed() {
                 if flags & CHANGE_DESTROYED != 0 {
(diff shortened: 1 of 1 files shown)


nook1208 (approved):

rmm/src/realm/mm/rtt.rs
```diff
@@ -384,7 +384,7 @@ pub fn set_ripas(rd: &Rd, base: usize, top: usize, ripas: u8, flags: u64) -> Res
         if ripas as u64 == invalid_ripas::EMPTY {
             new_s2tte |= bits_in_reg(S2TTE::INVALID_RIPAS, invalid_ripas::EMPTY);
 
-            if s2tte.is_unassigned_empty() {
```
nook1208, line 387: > if RTT_SET_RIPAS were to be run on ASSIGNED_EMPTY RTTEs with the requested RIPAS being EMPTY

I think this is not the case that you mention. The current code (AS-IS code) is from UNASSIGNED_EMPTY to UNASSIGNED_EMPTY which it means do nothing. Although that's not a problem, It looks good to use 'continue' in the 
'else' condition like you did 👍 

And it seems that the case which from UNASSIGNED_RAM to UNASSIGNED_EMPTY was missing. It's nice to be added. Thank you !

bokdeuk-jeong (approved):

PR #428, opened 2025-05-02
Update TF-RMM and TF-A (v2.11)

zpzigi754:
This commit prepares for using 1.0-REL0 by updating TF-RMM (rmm-spec-v1.0-eac5 (2023-10-31) to [fdcf08](https://github.com/TF-RMM/tf-rmm/commit/fdcf08ed30527650f13151ea45e9601b1b125dca) (2024-09-17)) and TF-A (v2.10 to v2.11).

TF-RMM's rmm-spec-v1.0-rel0 branch (2024-10-03) has a compatibility issue, so this commit updates TF-RMM and TF-A which do not cause that issue.

[fdcf08](https://github.com/TF-RMM/tf-rmm/commit/fdcf08ed30527650f13151ea45e9601b1b125dca) is the last commit before the commit breaking a backward compatibility ([715ff8](https://github.com/TF-RMM/tf-rmm/commit/715ff8bd4a227890dd8e7849a4502c0ed6a26c44)).

According to [715ff8](https://github.com/TF-RMM/tf-rmm/commit/715ff8bd4a227890dd8e7849a4502c0ed6a26c44), it says that RMM would expect RMM-EL3 interface v0.3, but there seems no build option in TF-A for enforcing that version of the interface, which causes almost all ACS tests to fail (I've tried updating ACS version, but it also didn't work).

Until this issue is resolved, it would be better to use an older version ([fdcf08](https://github.com/TF-RMM/tf-rmm/commit/fdcf08ed30527650f13151ea45e9601b1b125dca) (2024-09-17)).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,7 +20,6 @@
 [submodule "third-party/tf-rmm"]
 	path = third-party/tf-rmm
 	url = https://git.trustedfirmware.org/TF-RMM/tf-rmm.git
-	branch = rmm-spec-v1.0-eac5
 [submodule "third-party/android-kernel"]
 	path = third-party/android-kernel
 	url = https://github.com/islet-project/3rd-android-kernel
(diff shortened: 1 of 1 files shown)


bokdeuk-jeong (approved):

hihi-wang (approved):

nook1208:
Good to prepare the 1.0-REL0 spec !

PR #432, opened 2025-07-08
Update RmmFeatures

bokdeuk-jeong:
1.  Update RmmFeatures
    - add num_recs_order
    - relocate the position of existing fields in output for RMI_FEATURES

    Affected RMIs
    - RMI_FEATURES
    - RMI_REALM_CREATE
    - RMI_REC_CREATE/DESTROY

2. apply checks from clippy and cargo fmt
3. Update lkvm argument for the linux private memory usage

### DIFF
--- a/rmm/src/realm/rd.rs
+++ b/rmm/src/realm/rd.rs
@@ -32,6 +32,7 @@ pub struct Rd {
     s2_starting_level: isize,
     hash_algo: u8,
     rpv: [u8; RPV_SIZE],
+    num_recs: usize,
     pub measurements: [Measurement; MEASUREMENTS_SLOT_NR],
     pub vcpu_index: usize,
     metadata: Option<usize>,
@@ -66,10 +67,15 @@ impl Rd {
         if sve_en {
             self.simd_cfg.sve_vq = sve_vl;
         }
+        self.num_recs = 0;
     }
 
     pub fn id(&self) -> usize {
         self.vmid as usize
+    }
+
+    pub fn num_recs(&self) -> usize {
+        self.num_recs
     }
 
     pub fn s2_table(&self) -> Arc<Mutex<Box<dyn IPATranslation>>> {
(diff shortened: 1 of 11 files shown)


zpzigi754:
It might be better to use `cargo fmt` in main branch to separate 1.0-rel0-related commits from others, as most of formatting issues were caused in main branch.

scripts/fvp/launch-realm-debian.sh
```diff
@@ -7,6 +7,7 @@ echo "copy done"
 
 ./configure-net.sh &
 ./lkvm run \
+    --restricted_mem \
```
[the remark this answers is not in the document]
bokdeuk-jeong: Thanks for the check.

bokdeuk-jeong:
> It might be better to use `cargo fmt` in main branch to separate 1.0-rel0-related commits from others, as most of formatting issues were caused in main branch.

I agree with you in part. Code updates for fuzzing from these checks shouldn't be here. 
However, we do need to check clippy and cargo fmt even for PRs in this branch. 
Are you suggesting to revert the fuzz test code update in this PR?

rmm/src/realm/rd.rs
```diff
@@ -108,10 +114,15 @@ impl Rd {
...
 
-    pub fn inc_rec_index(&mut self) {
+    pub fn inc_recs(&mut self) {
+        self.num_recs += 1;
         self.rec_index += 1;
     }
 
+    pub fn dec_recs(&mut self) {
+        self.rec_index -= 1;
+    }
```
nook1208, line 124: Is this `self.num_recs -= 1;` not necessary here?

bokdeuk-jeong: I made a mistake. It should be "num_rec -= 1" according to the new spec update.

bokdeuk-jeong:
Sorry for the force-push updates. I chose force-push updates over leaving unnecessary commits.

zpzigi754 (approved):

PR #433, opened 2025-07-08
scripts: Add nw-linux in fvp-cca's clean option

zpzigi754:
This is a minor fix in fvp-cca script. nw-linux was missing in fvp-cca's clean option, so added.

Usage 1) Clean nw-linux
```
./scripts/fvp-cca --clean nw-linux
```
Usage 2) Clean all modules including nw-linux.
```
./scripts/fvp-cca --clean all
```


nook1208 (approved):
Good find!

PR #436, opened 2025-07-14
Update the rec exit handling for non-emulatable aborts

bokdeuk-jeong:
1. add a commit, "Bug fixes related to non emulatable aborts"  to cca-rmm-acs.
2. Update handle_realm_exit accordingly.
3. Update lower exception handler for the serror, irq, and wfx exceptions in trap.rs

### DIFF
--- a/rmm/src/rmi/rec/exit.rs
+++ b/rmm/src/rmi/rec/exit.rs
@@ -16,7 +16,7 @@ use crate::rmi::rec::run::Run;
 use crate::Monitor;
 use crate::{rmi, rsi};
 use armv9a::{
-    EsrEl2, DFSC_PERM_FAULTS, DFSC_PERM_FAULT_MASK, EMULATABLE_ABORT_MASK,
+    EsrEl2, DFSC_PERM_FAULTS, DFSC_PERM_FAULT_MASK, EMULATABLE_ABORT_MASK, INST_ABORT_MASK,
     NON_EMULATABLE_ABORT_MASK,
 };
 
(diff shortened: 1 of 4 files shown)


rmm/src/rmi/rec/exit.rs
```diff
@@ -103,32 +187,47 @@ fn handle_data_abort(
...
-                (far_el2 & !(GRANULE_MASK as u64)),
-            )
+            run.set_esr(esr_el2 & EMULATABLE_ABORT_MASK);
+            run.set_far(far_el2 & !(GRANULE_MASK as u64));
+            rmi::SUCCESS
+        }
+        AbortHandleType::DataAbortExit => {
+            rec.set_emulatable_abort(NotEmulatableAbort);
+            run.set_esr(esr_el2 & NON_EMULATABLE_ABORT_MASK);
+            run.set_far(0);
```
nook1208, line 264: I think the previous code always sets the far with far_el2 except the non emulatable case.
But if it's not necessary, then it would be good to set the far only for this DataAbortExit case.

nook1208: Oh I see. Then the 'DataAbortExit' type is not emulatble ? There are three different types (NonEmulatableExit,  EmulatableExit, DataAbortExit,). I'm little bit confused whether the DataAbortExit type is emulatble or not.

zpzigi754 (approved):
I've confirmed that the number of passed tests has increased from 87 to 89 with this PR.
The below tests have additionally passed.
```
mm_hipas_assigned_ripas_empty_da_ia
mm_hipas_unassigned_ripas_empty_da_ia
```

rmm/src/rmi/rec/exit.rs
bokdeuk-jeong: At a first glance, it looked like there should be only two types of data abort, Emulatable and Non-Emulatable.
However, after looking into the conditions for Emulatable and Non-Emulatable, I think there could be some cases that don't fall into these too.  I am not sure what the design intends.

So, I added DataAbortExit just as an internal classification, not seen externally.
As spec describes "On REC exit not due to Emulatable Data Abort, rec.emulatable_abort is NOT_EMULATABLE_ABORT."
For the DataAbortExit, I set it as NotEmulatableAbort instead.

nook1208 (approved):
LGTM 👍

PR #438, opened 2025-07-15
REC entry following REC exit due to Data Abort

bokdeuk-jeong:
SEA injection from the host is now correctly handled according to the spec, A4.2.3 REC entry following REC exit due to Data Abort.
Previously, acs-cca-test for mm_hipas_unassigned_ns_da_ia has been passed but wasn't actually handled correctly. No SEA injection took place on behalf of the host.

1. Update emulate_mmio() and add a new function, host_sea_inject(). 
    1) Rec context update:
       - add elr_el1, spsr_el1, far_el2 regsiters
       - don't restore exception registers at EL2 when returning to a realm. It's not necessary.
    2) move inject_sea to the rec directory. (rec/sea.rs)
       Since it updates realm context and inject_sea is called on both rec_exit and rec_enter, it is better to be under the rec.
       - don't directly manupulate sys registers, update context.sys_regs instead.
         (sea injection is always handled after realm exit.)
2.  move mmio.rs under rec
    Since emulate_mmio() modifies the rec context based on the result of handling an emulatable data abort,  it is better to move to the rec directory.

### DIFF
--- a/rmm/src/rmi/rec/run.rs
+++ b/rmm/src/rmi/rec/run.rs
@@ -232,30 +232,6 @@ define_bits!(
     EMUL_MMIO[0 - 0]
 );
 
-/// Whether the host has completed emulation for an Emulatable Data Abort.
-///  val 0: Host has not completed emulation for an Emulatable Abort.
-///  val 1: Host has completed emulation for an Emulatable Abort.
-pub const REC_ENTRY_FLAG_EMUL_MMIO: u64 = 1 << 0;
-/// Whether to inject a Synchronous External Abort (SEA) into the Realm.
-///  val 0: Do not inject an SEA into the Realm.
-///  val 1: Inject an SEA into the Realm.
-#[allow(dead_code)]
-pub const REC_ENTRY_FLAG_INJECT_SEA: u64 = 1 << 1;
-/// Whether to trap WFI execution by the Realm.
-///  val 0: Trap is disabled.
-///  val 1: Trap is enabled.
-#[allow(dead_code)]
-pub const REC_ENTRY_FLAG_TRAP_WFI: u64 = 1 << 2;
-/// Whether to trap WFE execution by the Realm.
-///  val 0: Trap is disabled.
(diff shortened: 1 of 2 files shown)


rmm/src/rmi/rec/run.rs
```diff
@@ -204,6 +206,32 @@ impl core::fmt::Debug for Run {
     }
 }
 
+// EntryFlag corresponds to the RmiRecEnterFlags fieldset in the spec document,
+// containing flags provided by the Host during REC entry.
+define_bits!(
+    EntryFlag,
```
zpzigi754, line 212: [minor]
Any reason to use a different name (`EntryFlag` instead of `EnterFlags`)?
I think that it would be better to use the same name if possible (e.g, use `Enter` type instead of `Entry` type for `RmiRecEnter` type).

rmm/src/lib.rs
```diff
@@ -17,7 +17,6 @@ pub mod granule;
 pub(crate) mod host;
 pub mod logger;
 pub mod mm;
-pub mod mmio;
```
zpzigi754, line 20: I think that keeping the previous structure (locating mmio under top instead `./rec/`) might be better, because mmio is not usually thought of a sub category of execution context.

bokdeuk-jeong: My suggestion is focused on current mmio.rs's role changing the rec's context and this way looks inline with the existing gic.rs and timer.rs under the rec directory for the same reason. 
Howerver, taking mmio out of rec as a module providing rec with services handling mmio is also worth considering. In that case, we need to move all mmio related to mmio.rs such as the code in rmi/rec/exit.rs determining emulatable and non-emulatable as well.

rmm/src/rmi/rec/run.rs
bokdeuk-jeong: I wanted to avoid using plural as struct name and I thought EntryFlag delivers its intended meaning enough. But, either is okay to me. :)

PR #447, opened 2025-07-30
RMM Spec Version Update to v1.0-rel0

bokdeuk-jeong:
- ACS-Test (88 success/7failure from measurement and attestation/5 skipped out of 101 tests) and launching a linux realm 
- Remaing work including attestation update is to be done after mering into the main.


zpzigi754:
Realm-linux booting in ci requires preparing for the pre-built image, so I've disabled the test temporarily.

PR #449, opened 2025-08-01
Update optee-build

bokdeuk-jeong:
1. Use recent optee-build 
   - url: https://git.codelinaro.org/linaro/dcap/op-tee-4.2.0/build.git
   - branch: remotes/m/cca/v4 -> clo/cca/v4

2. Add qemu build support.
   Only for tf-rmm.
   ./script/qemu-cca -nw linux -rm linux -rmm tf-rmm -bo --no-sdk

3. use prebuilt edk2 and don't clone qemu submodules
    To save storage space and time for compilation,
    use prebuilt edk stored in assets/prebuilt
    and don't clone qemu's submodules, which optee-build doesn't use.

4. Remote obsolete android


nook1208:
There is a `qemu-cca.py` script which run Host Android with prebuilt images using qemu in the same scripts directory. 
But it is different from the `qemu-cca` script which you pushed. So I think that they don't need to be merged into one script, but the naming of the `qemu-cca.py` have to be change.

bokdeuk-jeong:
> There is a `qemu-cca.py` script which run Host Android with prebuilt images using qemu in the same scripts directory. But it is different from the `qemu-cca` script which you pushed. So I think that they don't need to be merged into one script, but the naming of the `qemu-cca.py` has to be changed to avoid confusion.

Current qemu-cca.py need to be modified along with the qemu-cca. I regard this is out of scope in this PR.

nook1208 (approved):

PR #457, opened 2025-08-28
Move rec independent code from rec

bokdeuk-jeong:
1.  gic, smid code split
There was an opinion that gic and simd are better kept out of the 'rec' directory in a previous PR.
To meet in the middle, I moved gic and simd code  which are independent from the realm context to the upper directory.

2. use boot arguments provided by EL2 on rmm entry
- x0: cpuid
- x1: version (unused)
- x2: total core count (unused)
- x3: shared buf containing manifest


nook1208 (approved):
Looks good 👍

PR #459, opened 2025-09-01
Fix errors discoverd during the miri/fuzz test.

bokdeuk-jeong:
1. Misimplementation against Spec is fixed
    - rtt: correct RMI_RTT_CREATE handler
        Remove the condition, !parent_s2tte.is_assigned_ns(level - 1), which is not in the success or failure conditions for RMI_RTT_CREATE in the RMMv1.0-rel0 spec. 
        (Fuzzer reached to this condition and triggered the assertion inside.)
   -  correct RMI_FEATURES miri test
       In case that the index input is zero, it should test if the output bits[42:63] is MBZ according to the spec.  (Current miri tests if bits[30:63] is MBZ)

2.  Use hard-coded feat_reg0 value for fuzzing. Otherwise, fuzzer crashes while it tries to initialize feat_reg0 with architecture specific values such as simd and gic.
     Add 'fuzzing' to the lines in  #[cfg(not(any(miri, test)))])

3. Add #cfg to conditionally build functions only used for the fuzz testing

### DIFF
--- a/rmm/src/rmi/rec/run.rs
+++ b/rmm/src/rmi/rec/run.rs
@@ -42,12 +42,10 @@ struct Exit {
     0x410 cntv_ctl: u64,
     0x418 cntv_cval: u64,
     0x500 ripas_base: u64,
-    0x508 ripas_size: u64,
+    0x508 ripas_top: u64,
     0x510 ripas_value: u8,
     0x600 imm: u16,
-    0x700 pmu_ovf: u64,
-    0x708 pmu_intr_en: u64,
-    0x710 pmu_cntr_en: u64,
+    0x700 pmu_ovf: u8,
     0x800 => @END,
 }
 );
@@ -131,9 +129,9 @@ impl Run {
         Ok(())
     }
 
-    pub fn set_ripas(&mut self, base: u64, size: u64, state: u8) {
+    pub fn set_ripas(&mut self, base: u64, top: u64, state: u8) {
         self.exit.ripas_base = base;
-        self.exit.ripas_size = size;
+        self.exit.ripas_top = top;
         self.exit.ripas_value = state;
     }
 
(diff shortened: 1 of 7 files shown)


nook1208 (approved):

rmm/src/rmi/rec/run.rs
```diff
@@ -73,10 +73,12 @@ impl Run {
         self.entry.gicv3_hcr
     }
 
+    #[cfg(fuzzing)]
```
zpzigi754, line 74: I guess that multiple cfg can be reduced to one like

```
#[cfg(fuzzing)]
impl Run {
    pub fn set_entry_flags(&mut self, flags: u64) {
    ...
    pub fn set_entry_gpr(&mut self, idx: usize, val: u64) -> Result<(), Error> {
    ...
}
```

rmm/src/rsi/mod.rs
```diff
@@ -102,7 +102,9 @@ pub fn do_host_call(
         )
         .ok_or(Error::RmiErrorInput)?;
 
-    let mut host_call = assume_safe::<HostCall>(pa.into())?;
+    // page mapping result 'pa' is PAGE_SIZE aligned. Get hsot_call_ptr.
+    let host_call_ptr: usize = pa.as_usize() + ipa % crate::config::PAGE_SIZE;
```
zpzigi754, line 106: The commit message says that alignment requirement RSI_HOST_CALL was relaxed from 4KB to 256B. 
In this context, why is `PAGE_SIZE` (4KB) used instead of 256B?

I also wonder why ipa is added to pa for getting `host_call_ptr`.

bokdeuk-jeong: B5.3.4.2 Failure conditions
addr_align
   pre: !AddrIsAligned(addr, 256)
   post: result == RSI_ERROR_INPUT

The address of the HostCall is alinged 256 bytes. Thus, it can be anywhere in a page as long as it is 256 bytes aligned.
As pa is the result of mapping a page containing the HostCall object, we need the offset of HostCall object from the begining of the page.

PR #462, opened 2025-09-05
Support tf-a-tests

bokdeuk-jeong:
1. Enable tf-a-tests for rmmv1.0-rel
    There's a topics/rmm-1.0-rel0-rc1 branch in the tf-a-tests upstream. However, its RSI_IPA_STATE_GET command usage is still following eac5. So, I used the tags/lts-v2.12.1 branch which is close to the rmm-1.0-rel0-rc1.
   https://github.com/islet-project/3rd-tf-a-tests

2. Make islet-rmm pass tf-a-tests (realm payload test only)
   - [x] Realm create
   - [x] Realm memory management
   - [ ]  ~Multi-core~
      No failures have occured during the test.
   - [x] SVE
          + In case of SIMD including neon/sve/sme, 
          existing rmm saved/restored realm simd contexts if only used while saving/restoring ns contexts on every realm switch.
          Now, rmm does not store/restore NS simd contexts either if a realm doesn't use SIMD during its runtime (between REC_ENTER and REC EXIT)           
   - [x] SME
          SME is not supported for realms, but rmm preserves SME contexts for NS if used.
   - [ ] ~Interrupt~
        No failures have occured during the test.
   - [x] Etc(may disable): PMU/Pauth/lpa2 : 
     It already skips tests for Pauth and LPA2 if RMM doesn't support them, 
     but proceeds tests for PMU regardlessly. So, I commented PMU test out from the test list.
   

To test SVE and SME, update the FVP configuration as following.
```
diff --git a/scripts/.config b/scripts/.config
index 3049f852..cfcf58c8 100644
--- a/scripts/.config
+++ b/scripts/.config
@@ -64,3 +64,7 @@ pci.pci_smmuv3.mmu.SMMU_ROOT_IDR0=3
 pci.pci_smmuv3.mmu.SMMU_ROOT_IIDR=0x43B
 pci.pci_smmuv3.mmu.root_register_page_offset=0x20000
 pctl.startup=0.0.0.0
+SVE.ScalableVectorExtension.veclen=8                // for SVE
+SVE.ScalableVectorExtension.enable_at_reset=1  // for SME
+cluster0.has_arm_v9-2=1                                      // for SME
+SVE.ScalableVectorExtension.has_sme=1             // for SME
```
https://developer.arm.com/documentation/109246/0100/Toolchains-and-model-support/How-to-run-an-SME-application


scripts/tests/tf-a-tests.sh
```diff
@@ -31,33 +47,33 @@ check_result()
 }
 
 # tf-rmm tests
-$ROOT/scripts/fvp-cca -bo -nw=tf-a-tests -rmm=tf-rmm
-$ROOT/scripts/fvp-cca -ro -nw=tf-a-tests -rmm=tf-rmm &
+#$ROOT/scripts/fvp-cca -bo -nw=tf-a-tests -rmm=tf-rmm
+#$ROOT/scripts/fvp-cca -ro -nw=tf-a-tests -rmm=tf-rmm &
```
zpzigi754, line 51: [minor] I guess that the previous code can be removed if there is no intention to use it later.

rmm/src/exception/trap.rs
```diff
@@ -207,7 +208,11 @@ pub extern "C" fn handle_lower_exception(
                 debug!("Synchronous: SIMD");
                 let abort: bool = match Syndrome::from(esr) {
                     Syndrome::SVE => !rec.context.simd.cfg.sve_en,
-                    Syndrome::SME => !rec.context.simd.cfg.sme_en,
+                    // Note: Since only FEAT_SVE is set for Realms and
+                    // we are doing lazy restore, being reported as Syndrome::SME
+                    // is actually comming from the NW's SME setting.
+                    Syndrome::SME => !rec.context.simd.cfg.sve_en,
+                    //Syndrome::SME => !rec.context.simd.cfg.sme_en,
```
zpzigi754, line 215: It looks a bit strange that `cfg.sve_en` instead of `cfg.sme_en` is used for the case of `Syndrome::SME`. 
Is this intended? If so, maybe the commented code below can be removed.

rmm/src/simd.rs
```diff
@@ -90,3 +90,7 @@ pub fn sve_en() -> bool {
 pub fn max_sve_vl() -> u64 {
     SIMD_CONFIG.sve_vq
 }
+
+pub fn sme_en() -> bool {
+    SIMD_CONFIG.sme_en
```
zpzigi754, line 95: Accessing to a global variable may need to be wrapped with a lock.

nook1208 (commented):
> rec/exit: memeory excess outside the Realm IPA space

There are typos in your commit title. Maybe this is the correct sentence:
rec/exit: memory access outside the Realm IPA space

zpzigi754 (approved):

PR #474, opened 2025-11-07
Update confidential-ml example to be a standalone example

jinbpark:
This commit gets confidential-ml example to be a standalone example, which you can try from building from sources to running them on the host machine without relying on a special docker image.

This commit only supports simulated-enclave for now, meaning that running "device" binary on realm is not supported. This realm-execution supports will come in a next PR. I'm still working on this.

This PR is to solve https://github.com/islet-project/islet/issues/471, making confidential-ml example better reproducible.


bokdeuk-jeong (approved):
Thank you for the update! Better force merging the PR even if it has two ci failures. The failures are from using the old version of rust which is recently fixed in the main branch.

jinbpark:
https://github.com/islet-project/islet/issues/471#issuecomment-3540740956  --> [Carol1014](https://github.com/Carol1014) has just reproduced this PR successfully. I'm going to merge this PR and prepare for the next PR where "device" binary can be running inside Realm.

PR #485, opened 2026-05-26
Let SDK use the recent version of rust-rsi

p-sawicki2:
Make SDK compatible with RMM specification 1.0-rel0. Fix the expected "profile" string in the SDK's example.


Havner:
Also change the string in `sdk/src/lib.rs`

Havner:
I've tested simulated and qemu sdk examples

Havner (approved):

PR #486, opened 2026-05-26
Add restricted_mem option to lkvm while launching kvm-unit-tests

p-sawicki2:
The Islet/Arm CCA stack currently requires the use of Guest Private Memory. This is achieved by adding the --restricted_mem option to lkvm. Without it, the script for launching kvm-unit-tests fails to launch realms.

Here is the description on how to run the tests: https://github.com/islet-project/islet/blob/main/doc/getting-started/islet-how-to.md#testing-the-realm-features

> [!NOTE]
> Please test this PR using TF-RMM. For some reason Islet RMM fails to launch realms for kvm-unit-tests image.


Havner:
Verfied with TF-RMM on FVP and QEMU. With and without `attest`.

Everything works except one test on FVP that expects more CPUs. We have FVP configured with just one.

Havner (approved):

